PR #4, opened 2024-07-05
ci: add asf license check

QuakeWang:
This PR will add [skywalking-eyes](https://github.com/apache/skywalking-eyes) for checking ASF license header.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,8 +1,16 @@
-#[cfg(test)]
-mod tests {
-    #[test]
-    fn it_works() {
-        let result = 2 + 2;
-        assert_eq!(result, 4);
-    }
-}
+// Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -1,8 +1,16 @@
-#[cfg(test)]
```
Xuanwo, line 1: Remove this block could make CI to fail. We can remove them after we have added real code here.

Xuanwo (commented):
Mostly LGTM, thanks a lot!

src/lib.rs
QuakeWang: > Remove this block could make CI to fail. We can remove them after we have added real code here.

OK, I have updated it, PTAL.

Xuanwo:
Hi, @JingsongLi, I believe this PR is ready for merging.

JingsongLi (approved):
+1

PR #8, opened 2024-07-08
Remove main branch protect

JingsongLi:
Every commit will send a email to private:
Protected Branches for paimon-rust.git has been updated

Maybe we don't need to use protect branch?

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,16 +31,12 @@ github:
   enabled_merge_buttons:
     squash: true
     merge: false
-    rebase: false
+    rebase: true
   features:
     issues: true
     discussions: true
     wiki: false
     projects: false
-  protected_branches:
-    main:
-      required_pull_request_reviews:
-        required_approving_review_count: 1
     gh-pages:
       whatever: Just a placeholder to make it take effects
   collaborators:


.asf.yaml
```diff
@@ -31,16 +31,12 @@ github:
   enabled_merge_buttons:
     squash: true
     merge: false
-    rebase: false
+    rebase: true
```
JingsongLi, line 34: We can use rebase to some PRs collaborations with multiple people.

Xuanwo:
> Every commit will send a email to private:
> Protected Branches for paimon-rust.git has been updated

I've never heard of this. It might be because the project is relatively new and the same PR base doesn't have an asf.yml file yet. I believe this issue will resolve itself in the future.

.asf.yaml
Xuanwo: I'm fine with this.

Xuanwo:
> I've never heard of this. It might be because the project is relatively new and the same PR base doesn't have an asf.yml file yet. I believe this issue will resolve itself in the future.

Confirmed with @JingsongLi. I don't know what happened. Let's disable it for now.

SteNicholas (approved):
LGTM.

PR #18, opened 2024-07-14
feat(spec): Add primitive data types

XuQianJin-Stars:
This PR will add primitive data types.

### DIFF
--- a/crates/paimon/Cargo.toml
+++ b/crates/paimon/Cargo.toml
@@ -27,8 +27,10 @@ license.workspace = true
 version.workspace = true
 
 [dependencies]
+bitflags = "2.6.0"
 chrono = {version = "0.4.38", features = ["serde"]}
 serde = { version = "1", features = ["derive"] }
+serde_json = "1.0"
 serde_with = "3.8.3"
 snafu = "0.8.3"
 typed-builder = "^0.18"
--- a/crates/paimon/Cargo.toml
+++ b/crates/paimon/Cargo.toml
@@ -30,7 +30,6 @@ version.workspace = true
 bitflags = "2.6.0"
 chrono = {version = "0.4.38", features = ["serde"]}
 serde = { version = "1", features = ["derive"] }
-serde_json = "1.0"
 serde_with = "3.8.3"
 snafu = "0.8.3"
 typed-builder = "^0.18"
--- a/crates/paimon/src/spec/types.rs
+++ b/crates/paimon/src/spec/types.rs
@@ -151,8 +151,8 @@ pub struct DataType {
 }
 
 impl Display for DataType {
-    fn fmt(&self, _: &mut Formatter<'_>) -> std::fmt::Result {
-        todo!()
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.as_sql_string())
     }
 }
 
@@ -164,6 +164,7 @@ impl FromStr for DataType {
     }
 }
 
+#[allow(dead_code)]
 impl DataType {
     fn new(is_nullable: bool, type_root: DataTypeRoot) -> Self {
         Self {
@@ -267,16 +268,12 @@ impl DataType {
         serde_json::to_string(self).unwrap()
     }
 
-    fn with_nullability(&self, format: &str, params: &[&str]) -> String {
+    fn with_nullability(&self, _format: &str, params: &[&str]) -> String {
         if !self.is_nullable() {
-            format!("{} NOT NULL", format!("{}{}", format, params.concat()))
+            format!("{}{} NOT NULL",  _format, params.concat())
         } else {
-            format!("{}{}", format, params.concat())
+            format!("{}{}", _format, params.concat())
         }
-    }
-
-    fn to_string(&self) -> String {
-        self.as_sql_string()
     }
 
     fn accept<T>(&self, visitor: &mut T)
@@ -285,8 +282,6 @@ impl DataType {
     {
         visitor.visit(self);
     }
-
-    fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {}
 
     fn not_null(&self) -> Self {
         self.copy(false)
@@ -307,7 +302,6 @@ pub struct ArrayType {
 }
 
 impl ArrayType {
-    pub const FORMAT: &str = "ARRAY<{}>";
 
     pub fn new(is_nullable: bool) -> Self {
         Self {
@@ -317,10 +311,6 @@ impl ArrayType {
 
     pub fn default_value() -> Self {
         ArrayType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -341,10 +331,6 @@ impl BigIntType {
 
     pub fn default_value() -> Self {
         BigIntType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -394,10 +380,6 @@ impl BinaryType {
     pub fn get_length(&self) -> usize {
         self.length
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// BooleanType for paimon.
@@ -418,10 +400,6 @@ impl BooleanType {
     pub fn default_value() -> Self {
         BooleanType::new(true)
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// CharType for paimon.
@@ -441,14 +419,12 @@ impl CharType {
 
     pub const MAX_LENGTH: usize = 255;
 
-    pub const FORMAT: &str = "CHAR(%d)";
-
     pub fn new(is_nullable: bool, length: usize) -> Self {
         CharType::new_with_result(is_nullable, length).unwrap()
     }
 
     pub fn new_with_result(is_nullable: bool, length: usize) -> Result<Self, &'static str> {
-        if length < Self::MIN_LENGTH || length > Self::MAX_LENGTH {
+        if !(Self::MIN_LENGTH..=Self::MAX_LENGTH).contains(&length) {
             Err("Character string length must be between 1 and 255 (both inclusive).")
         } else {
             Ok(CharType {
@@ -472,10 +448,6 @@ impl CharType {
     pub fn get_length(&self) -> usize {
         self.length
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// DateType for paimon.
@@ -495,10 +467,6 @@ impl DateType {
 
     pub fn default_value() -> Self {
         DateType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -528,14 +496,15 @@ impl DecimalType {
     }
 
     pub fn new_with_result(is_nullable: bool, precision: u32, scale: u32) -> Result<Self, String> {
-        if precision < Self::MIN_PRECISION || precision > Self::MAX_PRECISION {
+        if !(Self::MIN_PRECISION..=Self::MAX_PRECISION).contains(&precision) {
             return Err(format!(
                 "Decimal precision must be between {} and {} (both inclusive).",
                 Self::MIN_PRECISION,
                 Self::MAX_PRECISION
             ));
         }
-        if scale < Self::MIN_SCALE || scale > precision {
+       
+        if !(Self::MIN_SCALE..=precision).contains(&scale) {        
             return Err(format!(
                 "Decimal scale must be between {} and the precision {} (both inclusive).",
                 Self::MIN_SCALE,
@@ -571,10 +540,6 @@ impl DecimalType {
     pub fn get_scale(&self) -> u32 {
         self.scale
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// DoubleType for paimon.
@@ -594,10 +559,6 @@ impl DoubleType {
 
     pub fn default_value() -> Self {
         DoubleType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -619,10 +580,6 @@ impl FloatType {
     pub fn default_value() -> Self {
         FloatType::new(true)
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// IntType for paimon.
@@ -642,10 +599,6 @@ impl IntType {
 
     pub fn default_value() -> Self {
         IntType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -670,7 +623,7 @@ impl LocalZonedTimestampType {
     }
 
     pub fn new_with_result(is_nullable: bool, precision: u32) -> Result<Self, String> {
-        if precision < Self::MIN_PRECISION || precision > Self::MAX_PRECISION {
+        if !(Self::MIN_PRECISION..=Self::MAX_PRECISION).contains(&precision) {    
             return Err(format!(
                 "Timestamp precision must be between {} and {} (both inclusive).",
                 Self::MIN_PRECISION,
@@ -698,10 +651,6 @@ impl LocalZonedTimestampType {
     pub fn get_precision(&self) -> u32 {
         self.precision
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// Next TODO: MapType、MultisetType、RowType
@@ -723,10 +672,6 @@ impl SmallIntType {
 
     pub fn default_value() -> Self {
         SmallIntType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -751,7 +696,7 @@ impl TimeType {
     }
 
     pub fn new_with_result(is_nullable: bool, precision: u32) -> Result<Self, String> {
-        if precision < Self::MIN_PRECISION || precision > Self::MAX_PRECISION {
+        if !(Self::MIN_PRECISION..=Self::MAX_PRECISION).contains(&precision) { 
             return Err(format!(
                 "Time precision must be between {} and {} (both inclusive).",
                 Self::MIN_PRECISION,
@@ -779,10 +724,6 @@ impl TimeType {
     pub fn get_precision(&self) -> u32 {
         self.precision
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// TimestampType for paimon.
@@ -806,7 +747,7 @@ impl TimestampType {
     }
 
     pub fn new_with_result(is_nullable: bool, precision: u32) -> Result<Self, String> {
-        if precision < Self::MIN_PRECISION || precision > Self::MAX_PRECISION {
+        if !(Self::MIN_PRECISION..=Self::MAX_PRECISION).contains(&precision) {     
             return Err(format!(
                 "Timestamp precision must be between {} and {} (both inclusive).",
                 Self::MIN_PRECISION,
@@ -834,10 +775,6 @@ impl TimestampType {
     pub fn get_precision(&self) -> u32 {
         self.precision
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// TinyIntType for paimon.
@@ -857,10 +794,6 @@ impl TinyIntType {
 
     pub fn default_value() -> Self {
         TinyIntType::new(true)
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
 
@@ -909,10 +842,6 @@ impl VarBinaryType {
     pub fn get_length(&self) -> u32 {
         self.length
     }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
-    }
 }
 
 /// VarCharType for paimon.
@@ -936,7 +865,7 @@ impl VarCharType {
     }
 
     pub fn new_with_result(is_nullable: bool, length: u32) -> Result<Self, String> {
-        if length < VarCharType::MIN_LENGTH || length > VarCharType::MAX_LENGTH {
+        if !(Self::MIN_LENGTH..=Self::MAX_LENGTH).contains(&length) {         
             return Err(format!(
                 "Character string length must be between {} and {} (both inclusive).",
                 VarCharType::MIN_LENGTH,
@@ -963,9 +892,5 @@ impl VarCharType {
 
     pub fn get_length(&self) -> u32 {
         self.length
-    }
-
-    pub fn collect_field_ids(&self, field_ids: &mut Vec<i32>) {
-        self.element_type.collect_field_ids(field_ids);
     }
 }
--- a/crates/paimon/src/spec/types.rs
+++ b/crates/paimon/src/spec/types.rs
@@ -969,10 +969,3 @@ impl VarCharType {
         self.element_type.collect_field_ids(field_ids);
     }
 }
-
-fn main() {
-    let data_type: DataType = DataType::new(true, DataTypeRoot::Char);
-    println!("{:?}", data_type);
-
-    let array_type: ArrayType = ArrayType::new(true);
-}
--- a/crates/paimon/src/spec/types.rs
+++ b/crates/paimon/src/spec/types.rs
@@ -48,12 +48,12 @@ bitflags! {
 /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataTypeRoot.java#L49>
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
 pub enum DataTypeRoot {
-    Char,
+    Char { length: usize },
     Varchar,
     Boolean,
-    Binary,
+    Binary { length: usize },
     Varbinary,
-    Decimal,
+    Decimal { precision: u32, scale: u32 },
     Tinyint,
     Smallint,
     Integer,
@@ -73,12 +73,19 @@ pub enum DataTypeRoot {
 impl DataTypeRoot {
     pub fn families(&self) -> DataTypeFamily {
         match self {
-            Self::Char => DataTypeFamily::PREDEFINED | DataTypeFamily::CHARACTER_STRING,
+            Self::Char { length: _ } => {
+                DataTypeFamily::PREDEFINED | DataTypeFamily::CHARACTER_STRING
+            }
             Self::Varchar => DataTypeFamily::PREDEFINED | DataTypeFamily::CHARACTER_STRING,
             Self::Boolean => DataTypeFamily::PREDEFINED,
-            Self::Binary => DataTypeFamily::PREDEFINED | DataTypeFamily::BINARY_STRING,
+            Self::Binary { length: _ } => {
+                DataTypeFamily::PREDEFINED | DataTypeFamily::BINARY_STRING
+            }
             Self::Varbinary => DataTypeFamily::PREDEFINED | DataTypeFamily::BINARY_STRING,
-            Self::Decimal => {
+            Self::Decimal {
+                precision: _,
+                scale: _,
+            } => {
                 DataTypeFamily::PREDEFINED | DataTypeFamily::NUMERIC | DataTypeFamily::EXACT_NUMERIC
             }
             Self::Tinyint => {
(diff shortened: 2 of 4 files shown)


crates/paimon/Cargo.toml
```diff
@@ -27,6 +27,7 @@ license.workspace = true
 version.workspace = true
 
 [dependencies]
+chrono = "0.4.38"
```
SteNicholas, line 30: This has been changed in #15.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1463 @@
...
+/// Data type for paimon table.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L45>
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DataType {
+    is_nullable: bool,
+    type_root: DataTypeRoot,
+}
+
+pub trait DataTypeTrait {
```
Xuanwo, line 168: Question: We need this trait because we want users to have their own data type?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1463 @@
...
+                "ARRAY"
+            } else {
+                "ARRAY NOT NULL"
+            },
+        )?;
+        state.serialize_field("element", &self.get_element_type())?;
+        state.end()
+    }
+
+    fn equals(&self, other: &Self) -> bool {
```
Xuanwo, line 399: Rust doesn't implement `equals` or `hash` in this way. We need to use `Hash` and `Eq`/`PartialEq`.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1463 @@
...
+    }
+}
+
+/// DecimalType for paimon.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/DecimalType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct DecimalType {
+    element_type: DataType,
+    precision: u32,
```
Xuanwo, line 723: How about using `rust_decimal`: https://docs.rs/rust_decimal/latest/rust_decimal/

XuQianJin-Stars:
Thanks  @Xuanwo @SteNicholas  I will modify it according to the comments later.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+    /// Returns true if the data type is with the family.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L224>
+    fn copy(&self, is_nullable: bool) -> Self {
+        Self {
+            is_nullable,
+            type_root: self.type_root.clone(),
+        }
+    }
+
+    /// Returns true if the data type is with the family.
```
Aitozi, line 228: Incorrect comment

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L214>
+    fn is_any_of_family(&self, families: &[DataTypeFamily]) -> bool {
+        families
+            .iter()
+            .any(|f: &DataTypeFamily| self.is_with_family(f.clone()))
+    }
+
+    /// Returns true if the data type is with the family.
```
Aitozi, line 219: Incorrect comment

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+    ///
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L195>
+    fn is_with_family(&self, family: DataTypeFamily) -> bool {
+        self.type_root.families().contains(family)
+    }
+
+    /// Returns true if the data type is with the family.
```
Aitozi, line 204: Incorrect comment

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+        }
+    }
+
+    /// Returns true if the data type is with the family.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L231>
+    fn copy_with_nullable(&self) -> Self {
+        self.copy(self.is_nullable)
+    }
+
+    /// Returns true if the data type is with the family.
```
Aitozi, line 234: ditto

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L240>
+    fn copy_ignore_nullable(&self) -> Self {
+        self.copy(false)
+    }
+
+    fn as_sql_string(&self) -> String {
```
Aitozi, line 240: This method should be moved to the subclass, Right?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+            element_type: DataType::new(is_nullable, DataTypeRoot::Array),
+        }
+    }
+
+    pub fn default_value() -> Self {
+        Self::new(true)
+    }
+
+    pub fn as_sql_string(&self) -> String {
+        format!("ARRAY<{}>", self.element_type.as_sql_string())
```
Aitozi, line 316: The `nullability` is not displayed in the sql string.

Aitozi (commented):
@XuQianJin-Stars Thanks for your contribution, Left some minor comments.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,956 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+use std::str::FromStr;
+
+bitflags! {
+/// An enumeration of Data type families for clustering {@link DataTypeRoot}s into categories.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/DataTypeFamily.java>
+#[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct DataTypeFamily: u32 {
```
Aitozi, line 29: I'm curious why we use bitset here instead of the original `HashSet` ?

Xuanwo: It's better to use a bitset here for easier readability and to avoid unnecessary allocation. `DataTypeFamily` just a `u32` which can store in stack but `HashSet` needs heap allocations.

Aitozi: Get it, thanks.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,985 @@
...
+/// Data type for paimon table.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L45>
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+pub struct DataType {
+    is_nullable: bool,
+    type_root: DataTypeRoot,
+}
+
+impl Display for DataType {
```
Aitozi, line 153: I think we should impl Display for the concrete types such as `BigIntType` not for the `DataType`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        const TIMESTAMP = 1 << 10;
+        const COLLECTION = 1 << 11;
+        const EXTENSION = 1 << 12;
+    }
+}
+
+/// The root of data type.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataTypeRoot.java#L49>
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
```
Xuanwo, line 49: Maybe we can make `DataTypeRoot` `Copy`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+    fn copy(&self, is_nullable: bool) -> Self {
+        Self {
+            is_nullable,
+            type_root: self.type_root.clone(),
+        }
+    }
+
+    /// Returns a deep copy of this type. It requires an implementation of {@link #copy(boolean)}.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L120>
+    fn copy_with_nullable(&self) -> Self {
```
Xuanwo, line 230: This is `Copy` in rust.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L94>
+    fn is_any_of_family(&self, families: &[DataTypeFamily]) -> bool {
+        families
+            .iter()
+            .any(|f: &DataTypeFamily| self.is_with_family(f.clone()))
+    }
+
+    /// Returns a deep copy of this type with possibly different nullability.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L113>
+    fn copy(&self, is_nullable: bool) -> Self {
```
Xuanwo, line 221: We can make this API called `fn with_nullable(self, is_nullable: bool) -> Self`

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    /// Returns a deep copy of this type. It requires an implementation of {@link #copy(boolean)}.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L120>
+    fn copy_with_nullable(&self) -> Self {
+        self.copy(self.is_nullable)
+    }
+
+    /// Compare two data types without nullable.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L129>
+    fn copy_ignore_nullable(&self) -> Self {
```
Xuanwo, line 236: This is duplicated with `with_nullable(false)`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        self.copy(false)
+    }
+
+    fn serialize_json(&self) -> String {
```
Xuanwo, line 240: Ideally, we don't need to provide API like this directly. If we have special requirement for `Serialize`, we need to implement by hand instead of using `#[derive(Serialize)]`

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        serde_json::to_string(&self.to_string()).unwrap()
+    }
+
+    fn with_nullability(&self, args: Arguments) -> String {
```
Xuanwo, line 244: The `with_xxx` API in Rust is typically used as a constructor. Should this API be provided by `DataType`, or would it be more appropriate for `Arguments`, such as in `Arguments::display_with_nullability()`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        visitor.visit(self);
+    }
+
+    fn not_null(&self) -> Self {
+        self.copy(false)
+    }
+
+    fn nullable(&self) -> Self {
+        self.copy(true)
+    }
```
Xuanwo, line 265: duplicated with `with_nullable`.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+/// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/ArrayType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct ArrayType {
+    pub element_type: DataType,
+}
+
+impl Display for ArrayType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 279: nit: Maybe use `if self.element_type.is_nullable() { abc } else { def }` for better reading?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+impl ArrayType {
+    pub fn new(is_nullable: bool) -> Self {
+        Self {
+            element_type: DataType::new(is_nullable, DataTypeRoot::Array),
+        }
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 294: Please implement `Default` instead.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/BigIntType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+pub struct BigIntType {
+    pub element_type: DataType,
+}
+
+impl Display for BigIntType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 309: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+impl BigIntType {
+    pub fn new(is_nullable: bool) -> Self {
+        Self {
+            element_type: DataType::new(is_nullable, DataTypeRoot::Bigint),
+        }
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 324: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+    }
+}
+
+/// BinaryType for paimon.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/BinaryType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct BinaryType {
+    pub element_type: DataType,
```
Xuanwo, line 335: I don't understand why we need to make `element_type` public. But it's not a big issue.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+    length: usize,
+}
+
+impl Display for BinaryType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 341: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    pub const MAX_LENGTH: usize = isize::MAX as usize;
+
+    pub const DEFAULT_LENGTH: usize = 1;
+
+    pub fn new(is_nullable: bool, length: usize) -> Self {
+        Self::new_with_result(is_nullable, length).unwrap()
+    }
+
+    pub fn new_with_result(is_nullable: bool, length: usize) -> Result<Self, &'static str> {
```
Xuanwo, line 360: We can call this API `fn try_new() -> Result<Self, Error>`. 

The Error type is not defined yet, I will polish this part later.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        } else {
+            write!(f, "BINARY({})", self.length)
+        }
+    }
+}
+
+impl BinaryType {
+    pub const MIN_LENGTH: usize = 1;
+
+    pub const MAX_LENGTH: usize = isize::MAX as usize;
```
Xuanwo, line 352: How about using `usize::MAX`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    pub const DEFAULT_LENGTH: usize = 1;
+
+    pub fn new(is_nullable: bool, length: usize) -> Self {
+        Self::new_with_result(is_nullable, length).unwrap()
+    }
+
+    pub fn new_with_result(is_nullable: bool, length: usize) -> Result<Self, &'static str> {
+        if length < Self::MIN_LENGTH {
+            Err("Binary string length must be at least 1.")
```
Xuanwo, line 362: nit: How about using early return like

```rust
if length < Self::MIN_LENGTH {
   return xxx
}

Ok(abc)
```

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+                element_type: DataType {
+                    is_nullable,
+                    type_root: DataTypeRoot::Binary,
+                },
+                length,
+            })
+        }
+    }
+
+    pub fn with_length(length: usize) -> Self {
```
Xuanwo, line 374: Nice API! You are mastering rust API pattern now!

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        Self::new(true, length)
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 378: Implement `Default` instead.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        Self::with_length(Self::DEFAULT_LENGTH)
+    }
+
+    pub fn get_length(&self) -> usize {
```
Xuanwo, line 382: How about using `fn length(&self) -> usize`

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/BooleanType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+pub struct BooleanType {
+    pub element_type: DataType,
+}
+
+impl Display for BooleanType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 397: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+impl BooleanType {
+    pub fn new(is_nullable: bool) -> Self {
+        Self {
+            element_type: DataType::new(is_nullable, DataTypeRoot::Boolean),
+        }
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 412: Implement `Default` instead.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CharType {
+    element_type: DataType,
+    length: usize,
+}
+
+impl Display for CharType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 429: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    pub const MIN_LENGTH: usize = 1;
+
+    pub const MAX_LENGTH: usize = 255;
+
+    pub fn new(is_nullable: bool, length: usize) -> Self {
+        Self::new_with_result(is_nullable, length).unwrap()
+    }
+
+    pub fn new_with_result(is_nullable: bool, length: usize) -> Result<Self, &'static str> {
```
Xuanwo, line 448: We can use `try_new`.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        if !(Self::MIN_LENGTH..=Self::MAX_LENGTH).contains(&length) {
```
Xuanwo, line 449: Let's perform early return.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+                length,
+            })
+        }
+    }
+
+    pub fn with_length(length: usize) -> Self {
+        Self::new(true, length)
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 466: Implement `Default` instead.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+        Self::with_length(Self::DEFAULT_LENGTH)
+    }
+
+    pub fn get_length(&self) -> usize {
```
Xuanwo, line 470: Use `length` instead.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/DateType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct DateType {
+    element_type: DataType,
+}
+
+impl Display for DateType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 485: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+impl DateType {
+    pub fn new(is_nullable: bool) -> Self {
+        Self {
+            element_type: DataType::new(is_nullable, DataTypeRoot::Date),
+        }
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 500: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    pub const MIN_SCALE: u32 = 0;
+
+    pub const DEFAULT_SCALE: u32 = 0;
+
+    pub fn new(is_nullable: bool, precision: u32, scale: u32) -> Self {
+        Self::new_with_result(is_nullable, precision, scale).unwrap()
+    }
+
+    pub fn new_with_result(is_nullable: bool, precision: u32, scale: u32) -> Result<Self, String> {
```
Xuanwo, line 540: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+            precision,
+            scale,
+        })
+    }
+
+    pub fn with_precision_and_scale(precision: u32, scale: u32) -> Self {
+        Self::new(true, precision, scale)
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 571: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+}
+
+impl DoubleType {
+    pub fn new(is_nullable: bool) -> Self {
+        Self {
+            element_type: DataType::new(is_nullable, DataTypeRoot::Double),
+        }
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 609: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/FloatType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+pub struct FloatType {
+    element_type: DataType,
+}
+
+impl Display for FloatType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
```
Xuanwo, line 624: The same.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+    pub fn default_value() -> Self {
+        Self::new(true)
+    }
+}
+
+/// LocalZonedTimestampType for paimon.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/types/TimestampType.java>.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize, Hash)]
+pub struct LocalZonedTimestampType {
```
Xuanwo, line 678: Interesting, paimon has a special type for local zone timestamp.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+impl Display for LocalZonedTimestampType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        if !self.element_type.is_nullable() {
+            write!(
+                f,
+                "TIMESTAMP WITH LOCAL TIME ZONE({}) NOT NULL",
+                self.precision
+            )
+        } else {
+            write!(f, "TIMESTAMP WITH LOCAL TIME ZONE({})", self.precision)
```
Xuanwo, line 692: We can implement like:

```rust
write!(f, "TIMESTAMP WITH LOCAL TIME ZONE({})", self.precision)？;
if !self.element_type.is_nullable() {
    write!(f, " NOT NULL")?;
}
Ok(())
```

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1039 @@
...
+            },
+            precision,
+        })
+    }
+
+    pub fn with_precision(precision: u32) -> Self {
+        Self::new(true, precision)
+    }
+
+    pub fn default_value() -> Self {
```
Xuanwo, line 730: The same.

Xuanwo (commented):
Thanks a lot for your work. We are on the right track. There are some comments for detailed API design, please take a look.

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1035 @@
...
+    ///
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L75>
+    fn is(&self, type_root: &DataTypeRoot) -> bool {
+        &self.type_root == type_root
+    }
+
+    /// Returns whether the family type of the type equals to the family or not.
+    ///
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L103>
+    fn with_family(&self, family: DataTypeFamily) -> bool {
```
Xuanwo, line 200: How about naming it `is_family`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1035 @@
...
+    /// Returns whether the root of the type equals to at least on of the type_roots or not.
+    ///
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L84>
+    fn is_any_of(&self, type_roots: &[DataTypeRoot]) -> bool {
+        type_roots.iter().any(|tr: &DataTypeRoot| self.is(tr))
+    }
+
+    /// Returns whether the root of the type is part of at least one family of the families or not.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L94>
+    fn is_any_with_family(&self, families: &[DataTypeFamily]) -> bool {
```
Xuanwo, line 213: How about naming it `is_any_of_family` to align with `is_any_of`?

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1035 @@
...
+        families
+            .iter()
+            .any(|f: &DataTypeFamily| self.with_family(f.clone()))
```
Xuanwo, line 216: `DataTypeFamily` is Copy, so I'm guessing we don't need to clone it.

Have you tried:

```rust
families
    .iter()
    .any(self.is_family)
```

crates/paimon/src/spec/types.rs
```diff
@@ -0,0 +1,1035 @@
...
+    }
+
+    /// Returns a deep copy of this type with possibly different nullability.
+    /// Impl Reference: <https://github.com/apache/paimon/blob/db8bcd7fdd9c2705435d2ab1d2341c52d1f67ee5/paimon-common/src/main/java/org/apache/paimon/types/DataType.java#L113>
+    fn with_nullable(&self, is_nullable: bool) -> Self {
```
Xuanwo, line 221: We can place construct API in the front to make it easier to find.

Xuanwo (approved):
Mostly LGTM to me now. There all only some small nits.

Xuanwo (approved):
Great, thanks a lot for your effort!

Xuanwo:
cc @SteNicholas, @Aitozi and @JingsongLi for another look.

Aitozi:
I have one more question, Why the DataType is not defined as trait ? If not, how the DataType used in other class. eg: `org.apache.paimon.types.DataField#asSQLString` the type string can not be obtained ?

Xuanwo:
> I have one more question, Why the DataType is not defined as trait ? If not, how the DataType used in other class. eg: `org.apache.paimon.types.DataField#asSQLString` the type string can not be obtained ? CC @Xuanwo

Hi, thanks for bringing this up. I don't quite understand the issue you mentioned. 

Are you trying to say that it's possible for users to implement they own data types and we don't know how to display those types?

Xuanwo:
After some offline discussion with @Aitozi, I got his point now. I think we need to build an enum for `DataType` instead of just `DataTypeRoot` in which we carries the entire data type information. We can implement this as follow-up of this PR.

Aitozi:
An enum `DataType` make sense to me. +1 for the current shape. Let's move for next.

Aitozi (approved):
+1

PR #21, opened 2024-07-25
feat(spec): implement data field

### LINKED ISSUE #22: spec: Implement Data Field
Try to implement data field with https://paimon.apache.org/docs/master/concepts/specification/

devillove084:
This PR will implement DataField.
close #22

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -18,7 +18,7 @@
 use crate::spec::types::DataType;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Display};
 
 /// The table schema for paimon table.
 ///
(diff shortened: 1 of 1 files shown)


Xuanwo:
Thanks a lit! We're planning to refactor the data type first, so this PR may need to wait a bit until we figure out the correct approach. Sorry for any inconvenience.

devillove084:
I noticed that the new data type has been merged, so I rebased onto the latest main branch and ensured that all local tests and clippy checks passed. When you have time, could you please approve the changes and review the code? If there are any omissions or inaccuracies, please point them out, and I'll address them promptly. @Xuanwo

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
     typ: DataType,
     description: Option<String>,
 }
+
+impl DataField {
+    pub fn new(id: i32, name: String, typ: DataType, description: Option<String>) -> Self {
```
Xuanwo, line 56: The `description` is optional; perhaps we could introduce an API named `with_description` to make `new()` short.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+
+    pub fn data_type(&self) -> &DataType {
+        &self.typ
+    }
+
+    pub fn description(&self) -> Option<&String> {
+        self.description.as_ref()
+    }
+
+    pub fn new_id(&self, new_id: i32) -> Self {
```
Xuanwo, line 81: We can have an API called `with_id`

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+        Self::new(
+            new_id,
+            self.name.clone(),
+            self.typ.clone(),
+            self.description.clone(),
+        )
+    }
+
+    pub fn new_name(&self, new_name: String) -> Self {
```
Xuanwo, line 90: The same, `with_name`.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+        Self::new(
+            self.id,
+            new_name,
+            self.typ.clone(),
+            self.description.clone(),
+        )
+    }
+
+    pub fn new_description(&self, new_description: String) -> Self {
```
Xuanwo, line 99: The same.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+        Self::new(
+            self.id,
+            self.name.clone(),
+            self.typ.clone(),
+            Some(new_description),
+        )
+    }
+
+    pub fn as_sql_string(&self) -> String {
```
Xuanwo, line 108: We don't need `format_string` and `as_sql_string`, we can implement `Display` instead.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+                "{} {} '{}'",
+                self.escape_identifier(&self.name),
+                type_string,
+                self.escape_single_quotes(desc)
+            ),
+            None => format!("{} {}", self.escape_identifier(&self.name), type_string),
+        }
+    }
+
+    fn escape_identifier(&self, identifier: &str) -> String {
```
Xuanwo, line 124: It's a bit odd to have those functions as methods of `DataField`. Perhaps we should move them out.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,176 @@ pub struct DataField {
...
+        text.replace('\'', "''")
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::spec::IntType;
+
+    use super::*;
+    // use crate::spec::DataTypeRoot;
```
Xuanwo, line 138: Let's remove not used code.

crates/paimon/src/spec/schema.rs
devillove084: Ok, I'll edit this.

crates/paimon/src/spec/schema.rs
devillove084: Ok

crates/paimon/src/spec/schema.rs
devillove084: Ok

crates/paimon/src/spec/schema.rs
devillove084: Ok

crates/paimon/src/spec/schema.rs
devillove084: Ok

crates/paimon/src/spec/schema.rs
devillove084: You are right, I put these functions outside as utility functions

crates/paimon/src/spec/schema.rs
devillove084: Ok

Xuanwo:
Hi, this PR includes some unrelated changes that make it difficult to review. Could you please address this?

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,138 @@ pub struct DataField {
...
+
+    pub fn data_type(&self) -> &DataType {
+        &self.typ
+    }
+
+    pub fn description(&self) -> Option<&String> {
+        self.description.as_ref()
+    }
+
+    pub fn with_id(&self, new_id: i32) -> Self {
```
Xuanwo, line 81: Hi, `with_id` better to take `Self`. So users can use them like:

```rust
let df = DataField::new(id, name, typ)
    .with_id(new_id)
    .with_description(Some(description));
```

In this way, we can avoid unneeded allocation.

devillove084: Are you suggesting that we use mut self directly in the with method to update the contents of self, rather than calling the new method?

devillove084: done :)

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,133 @@ pub struct DataField {
...
+        self.id = new_id;
+        self
+    }
+
+    pub fn with_name(mut self, new_name: String) -> Self {
+        self.name = new_name;
+        self
+    }
+
+    pub fn with_description(mut self, new_description: String) -> Self {
```
Xuanwo, line 91: I personal feel like it's better to have an `Option<String>` here so users can remove `description` if needed. 

It's also possible to have `impl Into<Option<String>>`, but I feel it's a bit over-designed for now.

crates/paimon/src/spec/schema.rs
```diff
@@ -51,3 +51,133 @@ pub struct DataField {
...
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn data_type(&self) -> &DataType {
+        &self.typ
+    }
+
+    pub fn description(&self) -> Option<&String> {
```
Xuanwo, line 77: Better to return `Option<&str>` instead. `&String` is rarely used in rust.

crates/paimon/src/spec/schema.rs
devillove084: I also think that impl Into<Option<String>> might be more appropriate if needed in the future, but I agree with using Option<String> for now.

crates/paimon/src/spec/schema.rs
devillove084: This is my oversight, I will fix it and remember this usage convention

crates/paimon/src/spec/schema.rs
devillove084: done

crates/paimon/src/spec/schema.rs
devillove084: done

Xuanwo (approved):
Thanks a lot!

JingsongLi:
Test failed, @devillove084 can you fix the test?

devillove084:
> Test failed, @devillove084 can you fix the test?

Okay, I will repair it later

devillove084:
I've already checked and fixed the issues with the tests. Whenever you have time, could you please help approve it? Thank you very much! @JingsongLi

JingsongLi (approved):
+1

PR #24, opened 2024-07-30
chore: Fix all reference links to tag instead

Xuanwo:
This PR fixed all reference links to tag instead to better reflect which version we are referring to.


SteNicholas (approved):
LGTM.

Issue #32, opened 2024-08-02
Refactor error in `DataType`


Xuanwo:
Would you like to perform this refactor? The error type has been defined at https://github.com/apache/paimon-rust/blob/a5cfac79ae4e9eaf87b752b3720fb0575cef3cbb/crates/paimon/src/error.rs#L25-L40

_Originally posted by @Xuanwo in https://github.com/apache/paimon-rust/pull/29#discussion_r1701900349_

XuQianJin-Stars:
I will try to implement this.

Xuanwo:
> I will try to implement this.

That's will be great!

PR #36, opened 2024-08-03
refactor: Refactor error in DataType

XuQianJin-Stars:
link: https://github.com/apache/paimon-rust/issues/32

### DIFF
--- a/crates/paimon/src/error.rs
+++ b/crates/paimon/src/error.rs
@@ -31,19 +31,15 @@ pub enum Error {
     },
     #[snafu(
         visibility(pub(crate)),
+        display("Paimon data type invalid for {}", message)
+    )]
+    DataTypeInvalid { message: String },
+    #[snafu(
+        visibility(pub(crate)),
         display("Paimon hitting unexpected error {}: {:?}", message, source)
     )]
     IoUnexpected {
         message: String,
         source: opendal::Error,
     },
-}
-
-#[derive(Debug, Snafu)]
-pub enum DataTypeError {
-    #[snafu(
-        visibility(pub(crate)),
-        display("Paimon data type invalid for {}", message)
-    )]
-    DataTypeInvalid { message: String },
 }
--- a/crates/paimon/src/error.rs
+++ b/crates/paimon/src/error.rs
@@ -40,95 +40,10 @@ pub enum Error {
 }
 
 #[derive(Debug, Snafu)]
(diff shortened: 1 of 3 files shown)


crates/paimon/src/error.rs
```diff
@@ -38,3 +38,12 @@ pub enum Error {
         source: opendal::Error,
     },
 }
+
+#[derive(Debug, Snafu)]
+pub enum DataTypeError {
```
Xuanwo, line 43: Hi, we usually expose only one error in each crate to keep them user-friendly. I will suggest to add `DataTypeInvalid` as a new enum in `Error` directly. In this way, we can use `paimon::Error` everywhere.

crates/paimon/src/spec/types.rs
```diff
@@ -320,13 +323,16 @@ impl CharType {
...
-    pub fn new(length: usize) -> Result<Self, &'static str> {
+    pub fn new(length: usize) -> Result<Self, Error> {
         Self::with_nullable(true, length)
     }
 
-    pub fn with_nullable(nullable: bool, length: usize) -> Result<Self, &'static str> {
+    pub fn with_nullable(nullable: bool, length: usize) -> Result<Self, Error> {
         if !(Self::MIN_LENGTH..=Self::MAX_LENGTH).contains(&length) {
-            return Err("Character string length must be between 1 and 255 (both inclusive).");
+            return DataTypeInvalidSnafu {
```
Aitozi, line 332: Can we simplify this by `ensure!(length < Self::MIN_LENGTH, DataTypeInvalidSnafu {message: "Binary string length must be at least 1."});` ?

The same as below

Aitozi (commented):
Left a minor comment.

crates/paimon/src/spec/types.rs
XuQianJin-Stars: > Can we simplify this by `ensure!(length < Self::MIN_LENGTH, DataTypeInvalidSnafu {message: "Binary string length must be at least 1."});` ?
> 
> The same as below

This simplifies things a bit, but the readability of the code is reduced.

Aitozi: Ok, make sense

Aitozi (approved):
LGTM

Xuanwo (approved):
THanks

PR #39, opened 2024-08-04
ut: add test for `Snapshot`

QuakeWang:
part of #26

PR #41, opened 2024-08-05
spec: Add missing property in Snapshot

Aitozi:
Some fields in Snapshot are lost, add them back

### DIFF
--- a/crates/paimon/src/spec/snapshot.rs
+++ b/crates/paimon/src/spec/snapshot.rs
@@ -39,6 +39,16 @@ pub struct Snapshot {
     #[builder(default = None)]
     index_manifest: Option<String>,
     commit_user: String,
+    // Mainly for snapshot deduplication.
+    //
+    // If multiple snapshots have the same commitIdentifier, reading from any of these snapshots
+    // must produce the same table.
+    //
+    // If snapshot A has a smaller commitIdentifier than snapshot B, then snapshot A must be
+    // committed before snapshot B, and thus snapshot A must contain older records than snapshot B.
+    commit_identifier: i64,
+    commit_kind: CommitKind,
+    time_millis: u64,
     /// record count of all changes occurred in this snapshot
     #[builder(default = None)]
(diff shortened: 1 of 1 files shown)


crates/paimon/src/spec/snapshot.rs
```diff
@@ -134,4 +144,38 @@ impl Snapshot {
...
+        self.commit_kind
+    }
+}
+
+/// Type of changes in this snapshot.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/Snapshot.java#L506>.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub enum CommitKind {
```
QuakeWang, line 172: @Aitozi We do not need this enum, you can refer #11.

crates/paimon/src/spec/snapshot.rs
```diff
@@ -39,6 +39,16 @@ pub struct Snapshot {
     #[builder(default = None)]
     index_manifest: Option<String>,
     commit_user: String,
+    // Mainly for snapshot deduplication.
+    //
+    // If multiple snapshots have the same commitIdentifier, reading from any of these snapshots
```
Aitozi, line 44: I have a question out of this PR, Currently, the overwrite (expire partition) share the same commit identifier  with the  current snapshots's commit identifier. So I think this comments may not right ? CC @JingsongLi

crates/paimon/src/spec/snapshot.rs
Aitozi: Hi @QuakeWang I think that just means `log_offsets` ?

QuakeWang:
It's my bad, when add the snapshot lost these field. Maybe I can add these in #39 with some tests. What do you think of this?

Aitozi:
@QuakeWang Sure, thanks.

Issue #33, opened 2024-08-02
Implement Ser/De for DataType Correct


Xuanwo:

XuQianJin-Stars:
I will try to implement this.

XuQianJin-Stars:
I will try to implement this.

Xuanwo:
> I will try to implement this.

Thanks in advance. We can use [`serde_with::DisplayFromStr`](https://docs.rs/serde_with/3.9.0/serde_with/struct.DisplayFromStr.html) to make it easier.

PR #42, opened 2024-08-06
feat: Implement Ser/De for DataType Correct

XuQianJin-Stars:
link: https://github.com/apache/paimon-rust/issues/33

### DIFF
--- a/crates/paimon/Cargo.toml
+++ b/crates/paimon/Cargo.toml
@@ -32,7 +32,7 @@ chrono = { version = "0.4.38", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde_bytes = "0.11.15"
 serde_json = "1.0.120"
-serde_with = "3.8.3"
+serde_with = "3.9.0"
 snafu = "0.8.3"
 typed-builder = "^0.19"
 opendal = "0.48"
--- a/crates/paimon/src/spec/tokens.rs
+++ b/crates/paimon/src/spec/tokens.rs
@@ -1,0 +1,1248 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::error::DataTypeParsingSnafu;
+use crate::spec::types::DataType;
+use crate::Error;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::{collections::HashSet, fmt, str::FromStr};
+
+use super::{
+    ArrayType, BigIntType, BinaryType, BooleanType, CharType, DataField, DateType, DecimalType,
+    DoubleType, FloatType, IntType, LocalZonedTimestampType, MapType, MultisetType, RowType,
+    SmallIntType, TimeType, TimestampType, TinyIntType, VarBinaryType, VarCharType,
+};
+
+/// --------------------------------------------------------------------------------------------
+/// Tokenizer
+/// --------------------------------------------------------------------------------------------
+
+const CHAR_BEGIN_SUBTYPE: char = '<';
+const CHAR_END_SUBTYPE: char = '>';
+const CHAR_BEGIN_PARAMETER: char = '(';
+const CHAR_END_PARAMETER: char = ')';
+const CHAR_LIST_SEPARATOR: char = ',';
+const CHAR_STRING: char = '\'';
+const CHAR_IDENTIFIER: char = '`';
+const CHAR_DOT: char = '.';
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub enum Keyword {
+    CHAR,
+    VARCHAR,
+    STRING,
+    BOOLEAN,
+    BINARY,
+    VARBINARY,
+    BYTES,
+    DECIMAL,
+    NUMERIC,
+    DEC,
+    TINYINT,
+    SMALLINT,
+    INT,
+    INTEGER,
+    BIGINT,
+    FLOAT,
+    DOUBLE,
+    PRECISION,
+    DATE,
+    TIME,
+    WITH,
+    WITHOUT,
+    LOCAL,
+    ZONE,
+    TIMESTAMP,
+    TimestampLtz,
+    INTERVAL,
+    YEAR,
+    MONTH,
+    DAY,
+    HOUR,
+    MINUTE,
+    SECOND,
+    TO,
+    ARRAY,
+    MULTISET,
+    MAP,
+    ROW,
+    NULL,
+    RAW,
+    LEGACY,
+    NOT,
+}
+
+impl fmt::Display for Keyword {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Keyword::CHAR => write!(f, "CHAR"),
+            Keyword::VARCHAR => write!(f, "VARCHAR"),
+            Keyword::STRING => write!(f, "STRING"),
+            Keyword::BOOLEAN => write!(f, "BOOLEAN"),
+            Keyword::BINARY => write!(f, "BINARY"),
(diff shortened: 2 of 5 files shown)


crates/paimon/src/spec/types.rs
```diff
@@ -346,10 +429,13 @@ impl BooleanType {
 /// Data type of a fixed-length character string.
 ///
 /// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/types/CharType.java>.
+#[serde_as]
 #[derive(Debug, Clone, PartialEq, Hash, Eq, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CharType {
+    #[serde_as(as = "DisplayFromStr")]
```
Xuanwo, line 436: Hi, I believe this is wrong since `CharType` should be serialized into `CHAR(n) NOT NULL`.

The `DisplayFromStr` should be added upon `CharType` instead of field instead. Also, `CharType` should implement `FromStr` which can parse a `CharType` from given string `CHAR(n) NOT NULL`.

Xuanwo (changes requested):

XuQianJin-Stars:
This pull request still has `Deserialize` and some optimizations to be completed.

crates/paimon/src/spec/types.rs
```diff
@@ -124,6 +126,10 @@ impl DataType {
             DataType::Row(v) => v.nullable,
         }
     }
+
+    fn serde_json(&self) -> serde_json::Value {
```
Xuanwo, line 130: Hi, I belive we don't need API like `serde_json`. Users will use `serde_json::to_string(&data_type)` directly

crates/paimon/src/spec/types.rs
```diff
@@ -157,7 +163,40 @@ impl FromStr for DataType {
     type Err = Error;
 
     fn from_str(_: &str) -> Result<Self, Self::Err> {
-        todo!()
+        todo!();
+    }
+}
+
+impl Serialize for DataType {
```
Xuanwo, line 170: Hi, this has native support in serde. Please add `#[serde(untagged)]` for `DataType` instead.

Reference: https://serde.rs/enum-representations.html#untagged

crates/paimon/src/spec/types.rs
```diff
@@ -200,12 +239,24 @@ impl ArrayType {
...
 
+impl Serialize for ArrayType {
+    fn serialize<S>(
+        &self,
+        serializer: S,
+    ) -> Result<<S as serde::Serializer>::Ok, <S as serde::Serializer>::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
```
Xuanwo, line 250: Can we use `#[serde_as(as = "DisplayFromStr")]` instead of implement `Serialize` by hand?

crates/paimon/src/spec/types.rs
```diff
@@ -243,12 +294,54 @@ impl BigIntType {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for BigIntType {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_str(BigIntTypeVisitor)
```
Xuanwo, line 314: I prefer to implement `from_str` instead and forward `Deserialize` to `from_str`.

crates/paimon/src/spec/types.rs
```diff
@@ -1338,4 +1496,204 @@ mod tests {
         ));
         assert_eq!(row_type.to_string(), "ROW<INTEGER, ARRAY<INTEGER>>");
     }
+
+    #[test]
+    fn test_datatype_serialize_deserialize() {
+        // BooleanType
+        let bool_type: BooleanType = BooleanType::with_nullable(true);
```
Xuanwo, line 1503: We don't need to specify the `BooleanType` and `String` here.

crates/paimon/src/spec/types.rs
```diff
@@ -1338,4 +1496,204 @@ mod tests {
...
+        let serialized: String = serde_json::to_string(&bool_type).unwrap();
+        assert_eq!(serialized, json!("BOOLEAN").to_string());
```
Xuanwo, line 1505: Maybe we can compare to `BOOLEAN` directly?

Xuanwo (commented):
Thanks a lot! This PR looks much better now, only some small nit.

crates/paimon/src/spec/types.rs
```diff
@@ -1338,4 +1496,204 @@ mod tests {
...
+        assert_eq!(serialized, json!("TIMESTAMP(6) NOT NULL").to_string());
+
+        // ArrayType
+        let int_type: DataType = DataType::Int(IntType::with_nullable(true));
+        let array_type: ArrayType = ArrayType::with_nullable(true, int_type.clone());
+        let serialized: String = serde_json::to_string(&array_type).unwrap();
+        assert_eq!(serialized, json!("ARRAY<INTEGER>").to_string());
+        let array_type: ArrayType = ArrayType::with_nullable(false, int_type.clone());
+        let serialized: String = serde_json::to_string(&array_type).unwrap();
+        assert_eq!(serialized, json!("ARRAY<INTEGER> NOT NULL").to_string());
```
Aitozi, line 1648: This serialized json is not right. These four types have different representation from the `SerializeDisplay`

image

Aitozi: image

The serialized version for `ARRAY<STRING>`

Xuanwo: I'm a bit confused. @JingsongLi mentioned that Paimon will use `as_sql_string` everywhere. I'll double-check this part.

crates/paimon/src/spec/tokens.rs
```diff
@@ -0,0 +1,1246 @@
...
+        &self.tokens[self.current_token as usize]
+    }
+
+    pub fn token_as_int(&self) -> isize {
+        self.token().value.parse::<isize>().unwrap_or_else(|_| {
+            self.parsing_error(String::from("Invalid integer value."));
+        })
+    }
+
+    pub fn token_as_keyword(&self) -> Keyword {
```
devillove084, line 344: Here you can directly use the Keyword::from_str method instead of repeatedly defining the matching logic. Something like: 
``` rust
Keyword::from_str(self.token().value.as_str())
        .unwrap_or_else(|_| self.parsing_error(format!("Expected keyword, found: {}", self.token().value)))
```

crates/paimon/src/spec/tokens.rs
```diff
@@ -0,0 +1,1246 @@
...
+        }
+        cursor - 1
+    }
+
+    pub fn tokenize(&mut self, chars: &str) -> Vec<Token> {
+        let mut tokens = Vec::new();
+        let mut builder = String::new();
+        let mut cursor = 0;
+        while cursor < chars.len() {
+            let cur_char = chars.chars().nth(cursor).unwrap();
```
devillove084, line 719: Repeatedly calling chars.chars().nth(cursor) to get the current character can lead to performance issues in Rust, especially when dealing with large strings. The reason is that each call to chars.chars().nth(cursor) iterates from the beginning of the string until it reaches the cursor position. This means that as the cursor increases, this operation becomes increasingly expensive.

crates/paimon/src/spec/tokens.rs
```diff
@@ -0,0 +1,1246 @@
...
+    pub fn last_cursor(&self) -> usize {
+        if self.last_valid_token < 0 {
+            0
+        } else {
+            self.tokens[self.last_valid_token as usize].cursor_position as usize + 1
+        }
+    }
+
+    pub fn parsing_error(&self, cause: String) -> ! {
+        panic!(
```
devillove084, line 322: Using panic directly here might not be ideal, but we can leave a TODO to explore a better approach in the future.

Xuanwo:
Thank you very much for your effort. I believe we can implement this in a more rusty way. I presented my idea here: https://github.com/apache/paimon-rust/pull/48. Would you like to take a look?

crates/paimon/Cargo.toml
```diff
@@ -32,7 +32,7 @@ chrono = { version = "0.4.38", features = ["serde"] }
 serde = { version = "1", features = ["derive"] }
 serde_bytes = "0.11.15"
 serde_json = "1.0.120"
-serde_with = "3.8.3"
+serde_with = "3.9.0"
```
SteNicholas, line 35: Why update this version？

crates/paimon/src/spec/tokens.rs
SteNicholas: Any TODO comment？

crates/paimon/src/spec/tokens.rs
```diff
@@ -0,0 +1,1246 @@
...
+            "Could not parse type at position {}: {}\nInput type string: {}",
```
SteNicholas, line 323:
```suggestion
            "Could not parse type at position {}: {}\nInput type: {}",
```

Xuanwo:
Hi, @SteNicholas, this PR can be closed now.

PR #45, opened 2024-08-08
feat(io): Align FileStatus with the Java API.

### LINKED ISSUE #44: Align FileStatus with the Java API.

Aitozi:
close #44

### DIFF
--- a/crates/paimon/src/io/file_io.rs
+++ b/crates/paimon/src/io/file_io.rs
@@ -18,7 +18,9 @@
 use crate::error::*;
 use std::collections::HashMap;
 
-use opendal::services::MemoryConfig;
+use chrono::offset::Utc;
+use chrono::DateTime;
+use opendal::services::Fs;
 use opendal::{Metakey, Operator};
 use snafu::ResultExt;
 
@@ -34,7 +36,7 @@ impl FileIO {
     ///
     /// TODO: Support building Operator from HashMap via options.
     pub fn new(_: HashMap<String, String>) -> Result<Self> {
-        let op = Operator::from_config(MemoryConfig::default())
+        let op = Operator::new(Fs::default().root("/"))
             .context(IoUnexpectedSnafu {
                 message: "Failed to create operator".to_string(),
             })?
@@ -72,6 +74,9 @@ impl FileIO {
 
         Ok(FileStatus {
             size: meta.content_length(),
+            is_dir: meta.is_dir(),
+            last_modification_time: meta.last_modified(),
+            path: path.to_string(),
         })
     }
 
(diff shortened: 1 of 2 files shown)


crates/paimon/src/io/file_io.rs
```diff
@@ -94,6 +97,13 @@ impl FileIO {
             .into_iter()
             .map(|meta| FileStatus {
                 size: meta.metadata().content_length(),
+                is_dir: meta.metadata().is_dir(),
+                last_modification_time: meta
```
Aitozi, line 101: When testing, I found that the directory's `last_modification_time` = 0 for a dir. But for the `LocalFileIO` in java it's not. 

image

Is this a bug for opendal `Fs` service ? CC @Xuanwo

Xuanwo: I believe it's related to https://github.com/apache/opendal/issues/4615. Would you like to take a look? It might be complex to resolve entirely, but we can start by fixing it for the fs backend first.

crates/paimon/src/io/file_io.rs
```diff
@@ -155,6 +165,9 @@ impl FileIO {
 #[derive(Clone, Debug)]
 pub struct FileStatus {
     pub size: u64,
+    pub is_dir: bool,
+    pub path: String,
+    pub last_modification_time: i64,
```
Xuanwo, line 170: I'm thinking about using [`chrono::DateTime`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html) to make it easier to use. What do you think?

Aitozi: Make sense, updated.

Aitozi:
CC @Xuanwo

crates/paimon/src/io/file_io.rs
```diff
@@ -72,6 +74,9 @@ impl FileIO {
 
         Ok(FileStatus {
             size: meta.content_length(),
+            is_dir: meta.is_dir(),
+            last_modification_time: meta.last_modified(),
```
Xuanwo, line 78: How about using `last_modified` or `modified` for short?

crates/paimon/src/io/file_io.rs
```diff
@@ -34,7 +36,7 @@ impl FileIO {
     ///
     /// TODO: Support building Operator from HashMap via options.
     pub fn new(_: HashMap<String, String>) -> Result<Self> {
-        let op = Operator::from_config(MemoryConfig::default())
+        let op = Operator::new(Fs::default().root("/"))
```
Xuanwo, line 39: It's better not to hard-code fs here, but I think it's a good starting point.

crates/paimon/src/io/file_io.rs
Aitozi: done

crates/paimon/src/io/file_io.rs
Aitozi: yes, I think here should construct the fileIO according to the scheme. I'd like to do it in another pr.

Xuanwo (approved):
Thanks, LGTM!

SteNicholas (approved):
LGTM.

PR #47, opened 2024-08-09
feat(test): add data type fixtures

### LINKED ISSUE #46: Implement Data Type test fixtures
In the original implementation of Paimon, as_sql_string is used for output or debugging, while JSON serialization is used for TableSchema format transmission, it is essential to distinguish between these two implementations. Therefore, we require the correct data type json format output as a reference.

devillove084:
Let's try using the serialization results from Java to check if the Rust implementation is correct！
close #46


XuQianJin-Stars (approved):
+1

Xuanwo (approved):
Thanks!

JingsongLi (approved):
+1

PR #49, opened 2024-08-10
feat: Implement ser/de for mostly all types

Xuanwo:
This PR implements mostly all data type Serialize & Deserialize.

Only data types that carrys `length` not covered.


devillove084 (approved):
Badass.

JingsongLi (approved):
+1

Issue #26, opened 2024-08-01
spec: Introduce unittests and integration tests


Xuanwo:
Our community works hard on implementation specifications. It's time for us to introduce unit tests and integration tests. We need to ensure that we can serialize/deserialize specifications correctly.

QuakeWang:
I will add the `Snapshot` test for soon.

QuakeWang:
The serialize/deserialize in `Snapshot` has completd by #39. I will add ut for the rest files.

Xuanwo:
> The serialize/deserialize in `Snapshot` has completd by #39. I will add ut for the rest files.

That would be greatly appreciated.

QuakeWang:
The following JSON is output by `ManifestFileData` with Java.

```json
{manifest-eddc1df1-3bfd-4918-bff3-a41560a8d280, 10000, 65, 35, org.apache.paimon.stats.BinaryTableStats@c2774299, 0}
```

The following JSON is output by `ManifestFileData` with Rust. In the Rust version has using `#[serde(rename = "_FILE_NAME")]` for rename the definition. 

```json
{"_FILE_NAME":"manifest_file_meta.json","_FILE_SIZE":1024,"_NUM_ADDED_FILES":10,"_NUM_DELETED_FILES":5,"_PARTITION_STATS":{"_MIN_VALUES":[1,2,3],"_MAX_VALUES":[4,5,6],"_NULL_COUNTS":[]},"_SCHEMA_ID":0}
```

Which one should be used when adding the serialize/deserialize test?

cc @Xuanwo

Xuanwo:
Does it mean `ManifestFileData` in java has it's own output format? It doesn't look like a valid JSON to me.

cc @devillove084, would you like to generate some fixture for `ManifestFileData` too?

devillove084:
> Does it mean `ManifestFileData` in java has it's own output format? It doesn't look like a valid JSON to me.
> 
> cc @devillove084, would you like to generate some fixture for `ManifestFileData` too?

@Xuanwo No problem, I will mention another ISSUE and related PR in some time later.

devillove084:
> > Does it mean `ManifestFileData` in java has it's own output format? It doesn't look like a valid JSON to me.
> > cc @devillove084, would you like to generate some fixture for `ManifestFileData` too?
> 
> @Xuanwo No problem, I will mention another ISSUE and related PR in some time later.

link #54

PR #56, opened 2024-08-14
ut: add serialize/deserialize tests for spec

QuakeWang:
part of #26


crates/paimon/src/spec/data_file.rs
```diff
@@ -90,16 +72,173 @@ pub struct DataFileMeta {
     pub delete_row_count: Option<i64>,
     // file index filter bytes, if it is small, store in data file meta
     pub embedded_index: Option<Vec<u8>>,
-    pub file_source: Option<FileSource>,
 }
 
 impl Display for DataFileMeta {
-    fn fmt(&self, _: &mut Formatter<'_>) -> std::fmt::Result {
-        todo!()
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
Xuanwo, line 78: `Debug` will generate this automatically; we don't need to implement it manually.

crates/paimon/src/spec/data_file.rs
```diff
@@ -90,16 +72,173 @@ pub struct DataFileMeta {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_data_file_meta_serialize_deserialize() {
+        let json_data = r#"
```
Xuanwo, line 189: Hi, @devillove084

https://github.com/apache/paimon-rust/blob/main/crates/paimon/tests/fixtures/manifest_file_meta_schema.json

This fixture appears to be just the schema of `ManifestFileMeta`. How is `ManifestFileMeta` data stored on disk?

Xuanwo: Oh, we should also have `DataFileMeta`'s datatype and `DataFileMeta` on disk format.

Aitozi: The default manifest file is of `avro` format. I think we should also generate a corresponding manifest file to verify the deserialize. But may be we should support avro/orc reader first ?

PR #58, opened 2024-08-17
feat: Implement ser/de for the remaining all

XuQianJin-Stars:


crates/paimon/src/spec/types.rs
```diff
@@ -223,6 +227,44 @@ impl Default for BinaryType {
     }
 }
 
+impl FromStr for BinaryType {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts: Vec<&str> = s.split_whitespace().collect();
```
Xuanwo, line 234: Hi, every `collect` here involves an allocation. It would be better if we could avoid it.

We can use `find` instead of `split` and `collect`, for example:

```rust
impl FromStr for BinaryType {
    type Err = Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if !s.starts_with(serde_utils::BINARY::NAME) {
            return DataTypeInvalidSnafu {
                message: xxx
            }
            .fail();
        }

        let Some(open_bracket) = s.find('(') else {
             return DataTypeInvalidSnafu {
                message: xxx
            }
            .fail();
        };
        let Some(close_bracket) = s.find(')') else {
             return DataTypeInvalidSnafu {
                message: xxx
            }
            .fail();
        };

        if open_bracket >= close_bracket {
            return DataTypeInvalidSnafu {
                message: xxx
            }
            .fail();
        }

        let length_str = &s[open_bracket + 1..close_bracket];
        let length = length_str.parse::<usize>().map_err(|_| DataTypeInvalidSnafu {
            message: xxx
        })?;

        let nullable = !s[close_bracket..].contains("NOT NULL");

        Ok(BinaryType {
            nullable,
            length,
        })
    }
}

```

XuQianJin-Stars:
The comments on this pull request have been addressed.

crates/paimon/src/spec/types.rs
```diff
@@ -601,6 +765,57 @@ impl Default for LocalZonedTimestampType {
...
+            .fail();
+        }
+
+        let Some(open_bracket) = s.find('(') else {
+            return DataTypeInvalidSnafu {
+                message: "Invalid LocalZonedTimestamp specification. Missing opening bracket.",
+            }
+            .fail();
+        };
+        let Some(close_bracket) = s.find(')') else {
```
Aitozi, line 786: What about extract a method to extract the `(number)` format ? There are almost 8 same usage

crates/paimon/src/spec/types.rs
```diff
@@ -706,6 +922,55 @@ impl Default for TimeType {
     }
 }
 
+impl FromStr for TimeType {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if !s.starts_with(serde_utils::TIME::NAME) || s.contains("STAMP") {
```
Aitozi, line 929: emm.. I feel here is a bit tricky.
BTW, Can `highly_complex_nested_row_type.json` pass this check ? We should also include this in tests ?

XuQianJin-Stars:
The comments on this pull request have been addressed.

Xuanwo (approved):
Thanks a lot for the contribution from @XuQianJin-Stars and @Aitozi's review! 

Some implementation details, like the `precision_scale_str.split(',').collect::<Vec<&str>>()`, could be improved, but I don't think it's a blocker for this PR.

Aitozi (approved):
LGTM, thanks @XuQianJin-Stars for your updates

Issue #52, opened 2024-08-12
Implement the Catalog API


QuakeWang:
when #27 is ready we can add [Catalog](https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java) API. Maybe we can add struct with definition first, the function and the detail behavior can be implemented for the further PRs.

PR #62, opened 2024-08-22
feat: add catalog API

QuakeWang:
fix: #52

### DIFF
--- a/crates/paimon/src/catalog/mod.rs
+++ b/crates/paimon/src/catalog/mod.rs
@@ -1,0 +1,254 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::collections::HashMap;
+use std::fmt;
+use std::hash::Hash;
+
+use async_trait::async_trait;
+use chrono::Duration;
+
+use crate::error::Result;
+use crate::io::FileIO;
+use crate::spec::{RowType, SchemaChange, TableSchema};
+
+/// This interface is responsible for reading and writing metadata such as database/table from a paimon catalog.
+///
+/// Impl References: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java#L42>
+#[async_trait]
+pub trait Catalog: Send + Sync {
+    const DEFAULT_DATABASE: &'static str = "default";
+    const SYSTEM_TABLE_SPLITTER: &'static str = "$";
+    const SYSTEM_DATABASE_NAME: &'static str = "sys";
+
+    /// Returns the warehouse root path containing all database directories in this catalog.
+    fn warehouse(&self) -> &str;
+
+    /// Returns the catalog options.
+    fn options(&self) -> &HashMap<String, String>;
+
+    /// Returns the FileIO instance.
+    fn file_io(&self) -> &FileIO;
+
+    /// Lists all databases in this catalog.
+    async fn list_databases(&self) -> Result<Vec<String>>;
+
+    /// Checks if a database exists in this catalog.
+    async fn database_exists(&self, database_name: &str) -> Result<bool>;
+
+    /// Creates a new database.
+    async fn create_database(
+        &self,
+        name: &str,
+        ignore_if_exists: bool,
+        properties: Option<HashMap<String, String>>,
+    ) -> Result<()>;
+
+    /// Loads database properties.
+    async fn load_database_properties(&self, name: &str) -> Result<HashMap<String, String>>;
+
+    /// Drops a database.
+    async fn drop_database(
+        &self,
+        name: &str,
+        ignore_if_not_exists: bool,
+        cascade: bool,
+    ) -> Result<()>;
+
+    /// Returns a Table instance for the specified identifier.
+    async fn get_table(&self, identifier: &Identifier) -> Result<impl Table>;
+
+    /// Lists all tables in the specified database.
+    async fn list_tables(&self, database_name: &str) -> Result<Vec<String>>;
+
+    /// Checks if a table exists.
+    async fn table_exists(&self, identifier: &Identifier) -> Result<bool> {
+        match self.get_table(identifier).await {
+            Ok(_) => Ok(true),
+            Err(e) => match e {
+                crate::error::Error::TableNotExist { .. } => Ok(false),
+                _ => Err(e),
+            },
+        }
+    }
+
+    /// Drops a table.
+    async fn drop_table(&self, identifier: &Identifier, ignore_if_not_exists: bool) -> Result<()>;
+
+    /// Creates a new table.
+    async fn create_table(
+        &self,
+        identifier: &Identifier,
+        schema: TableSchema,
+        ignore_if_exists: bool,
+    ) -> Result<()>;
+
+    /// Renames a table.
+    async fn rename_table(
+        &self,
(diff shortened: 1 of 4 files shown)


crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+use crate::error::Result;
+use crate::io::FileIO;
+use crate::spec::{RowType, SchemaChange, TableSchema};
+
+/// This interface is responsible for reading and writing metadata such as database/table from a paimon catalog.
+///
+/// Impl References: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java#L42>
+#[async_trait]
+pub trait Catalog: Send + Sync {
+    const DEFAULT_DATABASE: &'static str = "default";
```
Xuanwo, line 34: Using `const` will make this trait non-object safe, preventing users from invoking `Catalog` through `Arc<dyn Catalog>`. How about introducing a new API, such as `info`, for this purpose?

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+    const SYSTEM_DATABASE_NAME: &'static str = "sys";
+
+    /// Returns the warehouse root path containing all database directories in this catalog.
+    fn warehouse(&self) -> &str;
+
+    /// Returns the catalog options.
+    fn options(&self) -> &HashMap<String, String>;
+
+    /// Returns the FileIO instance.
+    fn file_io(&self) -> &FileIO;
```
Xuanwo, line 45: The catalog is integrated with file I/O, which is somewhat surprising to me.

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// Drops a database.
+    async fn drop_database(
+        &self,
+        name: &str,
+        ignore_if_not_exists: bool,
+        cascade: bool,
+    ) -> Result<()>;
+
+    /// Returns a Table instance for the specified identifier.
+    async fn get_table(&self, identifier: &Identifier) -> Result<impl Table>;
```
Xuanwo, line 73: Implementing `Table` makes this trait non-object safe. Perhaps we could return a `Table` struct instead.

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Lists all tables in the specified database.
+    async fn list_tables(&self, database_name: &str) -> Result<Vec<String>>;
+
+    /// Checks if a table exists.
+    async fn table_exists(&self, identifier: &Identifier) -> Result<bool> {
```
Xuanwo, line 79: It's a bit surprising that we use `Identifier` in some places but not in others. Shouldn't we be consistent?

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Drops a partition from the specified table.
+    async fn drop_partition(
+        &self,
+        identifier: &Identifier,
+        partitions: &HashMap<String, String>,
+    ) -> Result<()>;
+
+    /// Returns whether this catalog is case-sensitive.
+    fn case_sensitive(&self) -> bool {
```
Xuanwo, line 124: I believe we can consolidate those metadata-related items into a single API.

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+impl fmt::Display for Identifier {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.full_name())
+    }
+}
+
+/// A table provides basic abstraction for a table type and table scan, and table read.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/table/Table.java#L41>
+pub trait Table {
```
Xuanwo, line 193: I feel like `Table` should be a struct instead of a trait.

Xuanwo:
Also invite @Aitozi for a reivew.

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+/// Impl References: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java#L42>
+#[async_trait]
+pub trait Catalog: Send + Sync {
+    const DEFAULT_DATABASE: &'static str = "default";
+    const SYSTEM_TABLE_SPLITTER: &'static str = "$";
+    const SYSTEM_DATABASE_NAME: &'static str = "sys";
+
+    /// Returns the warehouse root path containing all database directories in this catalog.
+    fn warehouse(&self) -> &str;
+
```
XuQianJin-Stars, line 40: Do APIs like `lockFactory` need to be added?

crates/paimon/src/catalog/mod.rs
QuakeWang: OK, I will do it later.

crates/paimon/src/catalog/mod.rs
QuakeWang: > It's a bit surprising that we use `Identifier` in some places but not in others. Shouldn't we be consistent?

In the Java version, `Identifier` is usually used for table-related methods.

crates/paimon/src/catalog/mod.rs
Xuanwo: I need some input from @JingsongLi, @SteNicholas, and @Aitozi: Should a catalog be coupled with a specific file I/O?

Aitozi: Currently, the catalog is bind with a warehouse path which determines the fileIO. So these two things are coupled in current shape, and the table's schema are also retrieved from the filesystem. Also like to see some inputs from @JingsongLi @SteNicholas

crates/paimon/src/catalog/mod.rs
zmlcc: In the java version，`Table` interface represents various specialized table types. Is it possible to define a simple `Table` struct in rust ? 
What about returning `Box<dyn Table>` instead?

QuakeWang: > In the java version，`Table` interface represents various specialized table types. Is it possible to define a simple `Table` struct in rust ? What about returning `Box<dyn Table>` instead?

Maybe we will not use `dyn`, when returning `Box<dyn Table>` has some error like `error[E0404]: expected trait, found struct Table`. In Rust, `dyn` is used for dynamic dispatch with trait objects.

crates/paimon/src/catalog/mod.rs
QuakeWang: I will create a simple `Table` struct as the return value of the get_table method. With more design details about `Table` we can create a new issue for discussion. What do you think of it?

crates/paimon/src/catalog/mod.rs
zmlcc: > > In the java version，`Table` interface represents various specialized table types. Is it possible to define a simple `Table` struct in rust ? What about returning `Box<dyn Table>` instead?
> 
> Maybe we will not use `dyn`, when returning `Box<dyn Table>` has some error like `error[E0404]: expected trait, found struct Table`. In Rust, `dyn` is used for dynamic dispatch with trait objects.

Why `Box<dyn Table>` would give  `error[E0404]` ?  It should not when using properly.
I think in many cases, the return value of `get_table` needs to be downcast to a concrete type, so it makes sense to use `dyn Table`.
Alternatively,  maybe using  `enum Table` is feasible

crates/paimon/src/catalog/mod.rs
SteNicholas: @QuakeWang, @Xuanwo, @Aitozi, the warehoused path is required to determine which FileIO to use. This interface makes sense to me.

JiaQiTang98:
Hi teams, any update about the PR? Can we merge it and do something based on it?

Xuanwo:
I believe this PR is good to go.

Xuanwo:
> Hi teams, any update about the PR? Can we merge it and do something based on it?

I believe there some unresolved comments left: https://github.com/apache/paimon-rust/pull/62#discussion_r1738563385

JiaQiTang98:
@QuakeWang Hi, do you mind if I continue your job?

QuakeWang:
> @QuakeWang Hi, do you mind if I continue your job?

Sorry I've been busy recently and didn't check back sooner. Please feel free to do this with this as planned.

Issue #34, opened 2024-08-02
Implement FileIndex


devillove084:
- [x] Implement basic IO Module
- [ ] Implement a FunctionVistor(may be a simple one for poc) and FieldRef for FileIndexReader
- [x] Implement FileIndexFormat and FileIndexWriter/Reader and other interface
- [ ] Implement BitmapFileIndex
- [ ] Implement BloomfilterFileIndex

Regarding the work of implementing FileIndex, I will split it into multiple PRs to facilitate review and comment, so that there will be no problems in the implementation due to deviations in my personal understanding.

devillove084:
I've made some effort to separate the following implementation. If you have any suggestions for further improvement, please let me know. cc @Xuanwo

Xuanwo:
Thanks a lot for starting this!

suxiaogang223:
There is a new IndexFile called bit-slice bitmap index, i would like to try fellow your IndexFile framework to try to implement this functionality

PR #63, opened 2024-08-23
feat(file_index): add file index format read and write

devillove084:
part #34


devillove084:
This update involves the reconstruction of FileIndexFormat's reading and writing functionalities based on the new FileIO. To keep the changes manageable, I’ve split the PR into smaller parts. If you have a moment, I would greatly appreciate reviews. @Xuanwo @SteNicholas

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+use std::collections::HashMap;
+
+use bytes::{Buf, BufMut, Bytes, BytesMut};
+
+use crate::{
+    io::{FileIO, FileRead, InputFile, OutputFile},
+    Error,
+};
+
+pub const MAGIC: u64 = 1493475289347502;
```
Xuanwo, line 27: It's better to include a comment explaining its meaning and origin.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+                total_length += 2 + index_name.len();
+                // start_pos (4 bytes) + length (4 bytes)
+                total_length += 8;
+            }
+        }
+
+        Ok(total_length)
+    }
+
+    pub fn from_file_io(file_io: FileIO, path: &str) -> crate::Result<Self> {
```
Xuanwo, line 196: It's better to provide a single method for accomplishing a task. I believe we can use the same pattern throughout the entire project by consistently accepting a FileIO.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+/// - `redundant length`: 4-byte integer (for compatibility with future versions; content is zero in this version)
+/// - `redundant bytes`: variable-length bytes (for compatibility with future versions; empty in this version)
+/// - `BODY`: sequence of index data (concatenated index data for each column)
+/// ```
+#[derive(Debug)]
+pub struct FileIndexFormatWriter {
+    writer: OutputFile,
+}
+
+impl FileIndexFormatWriter {
```
Xuanwo, line 84: I'm guessing we need a `close` or `finish` method to finish the write?

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+                        },
+                    );
+                }
+            }
+        }
+
+        let head = self.write_head(&body_info).await?;
+        let mut combined = BytesMut::with_capacity(head.len() + body.len());
+        combined.extend_from_slice(&head);
+        combined.extend_from_slice(&body);
```
Xuanwo, line 124: Please write directly into `writer` to avoid extra alloc and memory usage unless you are certain that the data is small enough to be stored in memory.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+
+        self.writer.write(combined.freeze()).await?;
+        Ok(())
+    }
+
+    async fn write_head(
```
Xuanwo, line 130: `write_xxx` might mislead users into thinking we're writing to storage. It's actually more akin to `build_header`, isn't it?

By the way, it seems we don't need `async`?

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub fn from_file_io(file_io: FileIO, path: &str) -> crate::Result<Self> {
+        let output_file = file_io.new_output(path)?;
+        Ok(Self::new(output_file))
+    }
+}
+
+#[derive(Debug)]
+struct IndexInfo {
+    start_pos: i32,
+    length: i32,
```
Xuanwo, line 205: The `i32` type will overflow after 4GiB. Is that sufficient?

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+    }
+}
+
+impl From<Version> for i32 {
+    fn from(version: Version) -> Self {
+        match version {
+            Version::V1 => 1,
+        }
+    }
+}
```
Xuanwo, line 228: Maybe we can just use `#[repr(u8)]` for enum to avoid code like this?

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub fn new(writer: OutputFile) -> Self {
+        Self { writer }
+    }
+
+    pub async fn write_column_indexes(
+        &mut self,
+        indexes: HashMap<String, HashMap<String, Bytes>>,
+    ) -> crate::Result<()> {
+        let mut body_info: HashMap<String, HashMap<String, IndexInfo>> = HashMap::new();
+        let mut body = BytesMut::new();
```
Xuanwo, line 94: We can pre-allocate memory by calculating the entire data size first.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+            reader: Box::new(reader),
+            header: HashMap::new(),
+        };
+        file_reader.read_header().await?;
+        Ok(file_reader)
+    }
+
+    async fn read_header(&mut self) -> crate::Result<()> {
+        // Magic (8 bytes)
+        let mut buffer = self.reader.read(0..8).await?;
```
Xuanwo, line 248: Hi, reading 8B of data from storage services is very slow and costly. It's better to optimize by reading the entire header and then decoding it. We can try reading the first 1MiB to see if it contains the entire header and read more if it doesn't.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+pub struct FileIndexFormatWriter {
+    writer: OutputFile,
+}
+
+impl FileIndexFormatWriter {
+    pub fn new(writer: OutputFile) -> Self {
+        Self { writer }
+    }
+
+    pub async fn write_column_indexes(
```
Xuanwo, line 89: Perhaps we can gradually build the column indexes by accepting `column_name, index_name, data` and managing the index logic internally? This API makes me think why we need `FileIndexFormatWriter`? It looks more like a free function like `write_column_indexes(fileio, path, indexes)`.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+                        redundant_bytes.len()
+                    ),
+                });
+            }
+        }
+
+        Ok(())
+    }
+
+    pub async fn read_column_index(
```
Xuanwo, line 355: I suggest splitting the column index logic out of the reader. For example, have the reader return a `FileIndex` struct that includes all the indexes, allowing users to use it to `get_column_index` or `get_index`.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+    ) -> crate::Result<Bytes> {
+        let data_bytes = self
+            .reader
+            .read(index_info.start_pos as u64..(index_info.start_pos + index_info.length) as u64)
+            .await?;
+
+        Ok(data_bytes)
+    }
+
+    pub async fn from_file_io(file_io: FileIO, path: &str) -> crate::Result<Self> {
```
Xuanwo, line 385: The same.

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+/// - `index number`: 4-byte integer (number of index items below)
+/// - `index name x`: variable-length UTF-8 string
+/// - `start pos`: 4-byte integer
+/// - `length`: 4-byte integer
+/// - `redundant length`: 4-byte integer (for compatibility with future versions; content is zero in this version)
+/// - `redundant bytes`: variable-length bytes (for compatibility with future versions; empty in this version)
+/// - `BODY`: sequence of index data (concatenated index data for each column)
+/// ```
+#[derive(Debug)]
+pub struct FileIndexFormatWriter {
```
Xuanwo, line 80: I'm curious about the name. Why `FileIndexFormatWriter` instead of `FileIndexWriter`? Will we have other `FileIndexAbcWriter`?

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,514 @@
...
+
+#[cfg(test)]
+mod file_index_format_tests {
+    use super::*;
+    use bytes::Bytes;
+    use std::collections::HashMap;
+
+    #[tokio::test]
+    async fn test_single_column_single_index() -> crate::Result<()> {
+        let path = "file:/tmp/test_single_column_single_index";
```
Xuanwo, line 399: We can use in-memory file I/O for testing.

crates/paimon/Cargo.toml
```diff
@@ -34,6 +34,7 @@ storage-memory = ["opendal/services-memory"]
 storage-fs = ["opendal/services-fs"]
 
 [dependencies]
+rand = "0.8.5"
```
Xuanwo, line 37: Should this be `dev-dependencies`?

devillove084: You're right, that was an oversight on my part.

crates/paimon/src/file_index/file_index_format.rs
devillove084: No problem.

crates/paimon/src/file_index/file_index_format.rs
devillove084: In Java semantics, FileIndexWriter functions as a specialized interface class, and Format was originally intended as a shell class to expose static methods for handling the Format structure, I made some adjustments to reflect this design.

crates/paimon/src/file_index/file_index_format.rs
devillove084: OK

crates/paimon/src/file_index/file_index_format.rs
devillove084: Well, I might try to add corresponding methods in OutputFile, maybe adjust the current "write" method to "write" and explicit "finish"

crates/paimon/src/file_index/file_index_format.rs
devillove084: I will avoid this behavior based on the explicit "finish" method above.

crates/paimon/src/file_index/file_index_format.rs
devillove084: OK

crates/paimon/src/file_index/file_index_format.rs
devillove084: OK

crates/paimon/src/file_index/file_index_format.rs
devillove084: Your suggestion is better, it makes more sense to use the "build" semantics here, and in fact, async is not needed at this point.

crates/paimon/src/file_index/file_index_format.rs
devillove084: OK

crates/paimon/src/file_index/file_index_format.rs
devillove084: Indeed, that is better. I will figure it out.

crates/paimon/src/file_index/file_index_format.rs
devillove084: OK

crates/paimon/src/file_index/file_index_format.rs
devillove084: done

devillove084:
Sorry for being late again. I have corrected all your comments. Please review them when you have time to see if there are any problems. Thank you very much. @Xuanwo

crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -0,0 +1,517 @@
...
+/// - `column x`: variable-length UTF-8 string (length + bytes)
+/// - `index number`: 4-byte integer (number of index items below)
+/// - `index name x`: variable-length UTF-8 string
+/// - `start pos`: 4-byte integer
+/// - `length`: 4-byte integer
+/// - `redundant length`: 4-byte integer (for compatibility with future versions; content is zero in this version)
+/// - `redundant bytes`: variable-length bytes (for compatibility with future versions; empty in this version)
+/// - `BODY`: sequence of index data (concatenated index data for each column)
+/// ```
+pub async fn write_column_indexes(
```
XuQianJin-Stars, line 98: Add the link: https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/fileindex/FileIndexFormat.java ?

devillove084: ok

Xuanwo (approved):
Thanks, let's move!

PR #68, opened 2024-09-04
feat(file_index): add file index other interfaces

devillove084:
part #34


Xuanwo:
Hi, @devillove084, thanks a lot for your effort. However, I have some concerns that we are not heading in the right direction. Translating every Java class into a trait and using `Box<dyn Trait>` doesn't seem ideal to me now. I'm wondering if it's possible to implement a real file index struct first and then consider how to expose them.

devillove084:
Close it and move on!

PR #69, opened 2024-09-06
feat(spec): Impl the ManifestFile read functionality

### LINKED ISSUE #61: spec: Impl the ManifestFile read functionality
org.apache.paimon.manifest.ManifestEntrySerializer

Aitozi:
close #61


crates/paimon/src/spec/data_file.rs
```diff
@@ -99,7 +114,44 @@ impl Display for DataFileMeta {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 impl DataFileMeta {
     // TODO: implement me
     pub const SCHEMA: RowType = RowType::new(vec![]);
```
Xuanwo, line 120: Do we still need this?

crates/paimon/src/spec/data_file.rs
```diff
@@ -99,7 +114,44 @@ impl Display for DataFileMeta {
...
+
+    pub fn new(
```
Xuanwo, line 122: I feel like we don't need this since all fields of DataFileMeta is `pub`.

crates/paimon/src/spec/manifest_entry.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::spec::DataFileMeta;
+use serde::Deserialize;
+use serde_repr::{Deserialize_repr, Serialize_repr};
+use serde_with::serde_derive::Serialize;
+
+/// Entry representing a file.
+///
+/// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/manifest/FileEntry.java>
+#[allow(dead_code)]
+pub trait FileEntry {
```
Xuanwo, line 27: I don't think we need a trait for this. Will we have different types of file entries with the same API?

crates/paimon/src/spec/objects_file.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::Error;
+use apache_avro::types::Value;
+use apache_avro::{from_value, Reader};
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
+/// This file includes several [T]
+pub struct ObjectsFile<T> {
```
Xuanwo, line 30: Maybe we don't this struct? Just providing free functions like `read_manifest_file_meta` should be enough.

crates/paimon/src/spec/manifest_entry.rs
Aitozi: Yes, we have a `SimpleFileEntry` to reduce the memory cost in some case.

crates/paimon/src/spec/data_file.rs
Aitozi: No need, can be removed.

crates/paimon/src/spec/objects_file.rs
Aitozi: Removed.

crates/paimon/src/spec/manifest_entry.rs
Xuanwo: I suggest dividing them into two separate structures and converting them as needed. I believe this approach is much better than storing a `Box<dyn FileEntry>` everywhere.

Aitozi: Removed, and keep the `ManifestEntry` now, we can add the `SimpleFileEntry` when used later

Xuanwo (approved):
Thanks!

Aitozi:
CC @JingsongLi

SteNicholas (approved):
LGTM.

Issue #82, opened 2025-04-20
Support ManifestFileMeta new filed: _MIN_BUCKET,_MAX_BUCKET,_MIN_LEVEL,_MAX_BUCKET


camilesing:
As the title suggests, ManifestFileMeta has recently added some fields. We also need support here.

i can do this.

Xuanwo:
Thank you and welcome.

Issue #3, opened 2024-07-05
Tracking issues of paimon rust 0.1.0 release


Xuanwo:
Hello everyone, this is the tracking update for the paimon rust 0.1.0 release. This will be our initial release, and I aim to include basic read support in it.

## Tasks

- Spec: Implement a types that needed by paimon.
- Arrow Integration: Allow reading paimon data into arrow format.
- IO: Integrate with [Apache OpenDAL](https://github.com/apache/opendal) for IO.

QuakeWang:
@Xuanwo I will do the `Snapshot` task.

Xuanwo:
> @Xuanwo I will do the `Snapshot` task.

Welcome, have fun!

dharanad:
hello @Xuanwo , I would like to work on implementing `Manifest`.

Xuanwo:
> hello @Xuanwo , I would like to work on implementing `Manifest`.

Great, have fun!

crrow:
Hey, I'm new around here and this project sounds interesting. Can I jump in and help out?

Xuanwo:
> Manifest

Thanks a lot, welcome to pick up a part. I'm willing to update tracking issues for you.

crrow:
> > Manifest
> 
> Thanks a lot, welcome to pick up a part. I'm willing to update tracking issues for you.

okay, let me try data file

Asura7969:
I want to try the manifest list

thexiay:
Cool, it's good project to be familiar with rust and big data infrastructure😆

QuakeWang:
I will do the `Changelog` task, and the corresponding issue have been created #16.

Xuanwo:
> Cool, it's good project to be familiar with rust and big data infrastructure😆

Welcome!

> I will do the `Changelog` task, and the corresponding issue have been created #16.

Updated.

suxiaogang223:
Hi, I am a big fan of rust, and I usually do data lake-related development. Be happy to join the development of paimon-rust in the future.

Aitozi:
Does the FileSystemCatalog rely on the Integrate with [Apache OpenDAL](https://github.com/apache/opendal) ? There are some IO operation in the FileSystemCatalog. @Xuanwo

Xuanwo:
> Does the FileSystemCatalog rely on the Integrate with [Apache OpenDAL](https://github.com/apache/opendal) ?

Yes, my plan is using opendal for those IO operations.

[...]

PR #83, opened 2025-04-20
add field min, max, bucket, level for ManifestFileMeta

camilesing:
for https://github.com/apache/paimon-rust/issues/82


Aitozi:
Hi, @camilesing our first version is to work with the paimon 0.8.2 spec, these new fields are not included in 0.8.2. I suggest to align the spec with the 0.8.2 spec, otherwise we may mess up in different version.

camilesing:
@Aitozi In order to complete the first version as soon as possible, I agree to do so. However, I suggest stating in this issue that the parts we need to implement should follow Paimon 0.8.2. To avoid others doing unnecessary things.

Aitozi:
Yes, the necessary for 0.1 should be stated in https://github.com/apache/paimon-rust/issues/3 . Let's target 0.8.2 first

Issue #81, opened 2025-04-20
update dependency


camilesing:
This project has had few updates in the past six months, and some dependencies can be updated:
- rust-version: 1.77.1 -> 1.86.0
- version: 0.49 -> 0.53

i can do this

Xuanwo:
Welcome, I suggest we can bump MSRV first.

camilesing:
@Xuanwo you are right. Do you have any suggestions?

Xuanwo:
Bumping the MSRV should be straightforward, we just need to update rust-toolchain.toml and the `rust-version`, and address any potential Clippy errors.

PR #84, opened 2025-04-22
chore(deps):update rust version be 1.86.0

camilesing:
for https://github.com/apache/paimon-rust/issues/81


Xuanwo (approved):
Thank you for this change. Looks good to me.

---

Ideally, we should establish an MSRV policy, but paimin-rust is still a long way from being usable. I’m fine with setting the MSRV to the latest stable version of Rust.

Issue #86, opened 2025-05-22
CI: `cargo clippy --all-targets --workspace -- -D warnings`  error


QuakeWang:
When using `cargo clippy --all-targets --workspace -- -D warnings` to test the main branch code, which has errors like the following information, we need to fix this.

```bash
$ cargo clippy --all-targets --workspace -- -D warnings
    Checking paimon v0.0.0 (/Users/wangerxi/workspaces/paimon-rust/crates/paimon)
error: the `Err`-variant returned from this function is very large
   --> crates/paimon/src/file_index/file_index_format.rs:192:6
    |
192 |   ) -> crate::Result<usize> {
    |        ^^^^^^^^^^^^^^^^^^^^
    |
   ::: crates/paimon/src/error.rs:41:5
    |
41  | /     IoUnexpected {
42  | |         message: String,
43  | |         source: opendal::Error,
44  | |     },
    | |_____- the variant `IoUnexpected` contains at least 152 bytes
...
59  | /     DataUnexpected {
60  | |         message: String,
61  | |         source: apache_avro::Error,
62  | |     },
    | |_____- the largest variant contains at least 280 bytes
    |
    = help: try reducing the size of `error::Error`, for example by boxing large elements or replacing it with `Box<error::Error>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `-D clippy::result-large-err` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::result_large_err)]`

error: the `Err`-variant returned from this function is very large
  --> crates/paimon/src/io/file_io.rs:40:36
   |
40 |       pub fn from_url(path: &str) -> crate::Result<FileIOBuilder> {
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
  ::: crates/paimon/src/error.rs:41:5
   |
41 | /     IoUnexpected {
42 | |         message: String,
43 | |         source: opendal::Error,
44 | |     },
   | |_____- the variant `IoUnexpected` contains at least 152 bytes
...
59 | /     DataUnexpected {
60 | |         message: String,
61 | |         source: apache_avro::Error,
62 | |     },
   | |_____- the largest variant contains at least 280 bytes
   |
   = help: try reducing the size of `error::Error`, for example by boxing large elements or replacing it with `Box<error::Error>`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

[...]

PR #88, opened 2026-02-08
ci: fix clippy warn error to long issue

luoyuxia:
Fix issue #86 

I'm going to contribute back some features to paimon-rust, but the ci won't work now. So, submit a pr to fix the ci as well as running `cargo clippy --all-targets --fix --allow-dirty --allow-staged` do some format.

### DIFF
--- a/crates/paimon/src/error.rs
+++ b/crates/paimon/src/error.rs
@@ -40,7 +40,8 @@ pub enum Error {
     )]
     IoUnexpected {
         message: String,
-        source: opendal::Error,
+        #[snafu(source(from(opendal::Error, Box::new)))]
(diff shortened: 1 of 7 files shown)


crates/paimon/src/error.rs
```diff
@@ -40,7 +40,8 @@ pub enum Error {
     )]
     IoUnexpected {
         message: String,
-        source: opendal::Error,
+        #[snafu(source(from(opendal::Error, Box::new)))]
+        source: Box<opendal::Error>,
```
luoyuxia, line 44: see https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
which suggest us to wrap big size error to Box.

PR #92, opened 2026-02-14
feat: introduce catalog api 

### LINKED ISSUE #52: Implement the Catalog API

luoyuxia:
### Purpose

Linked issue: close #52

### DIFF
--- a/crates/paimon/src/catalog/mod.rs
+++ b/crates/paimon/src/catalog/mod.rs
@@ -1,0 +1,212 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Catalog API for Apache Paimon.
+//!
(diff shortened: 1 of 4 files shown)


crates/paimon/src/spec/schema.rs
```diff
@@ -102,6 +102,344 @@ pub fn escape_single_quotes(text: &str) -> String {
...
+    }
+
+    /// Normalize primary keys: optionally take from table options (`primary-key`), remove from options.
+    /// Corresponds to Java `normalizePrimaryKeys`.
+    fn normalize_primary_keys(
+        primary_keys: &[String],
+        options: &mut HashMap<String, String>,
+    ) -> crate::Result<Vec<String>> {
+        if let Some(pk) = options.remove(PRIMARY_KEY_OPTION) {
+            if !primary_keys.is_empty() {
```
XiaoHongbo-Hope, line 154: Will this affect the the original options? We can work on a clone of options to avoid mutating the original one.

luoyuxia: Yes, It'll. I will make the constructor to be private to prevent user using it. Use `SchemaBuilder` is simpler than the constructor since `SchemaBuilder` accepts &str and String,

crates/paimon/src/catalog/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    ) -> Result<()>;
+
+    // ======================= table methods ===============================
+
+    /// Get table metadata for the given identifier.
+    ///
+    /// # Errors
+    /// * [`crate::Error::DatabaseNotExist`] - database in identifier does not exist.
+    /// * [`crate::Error::TableNotExist`] - table does not exist.
+    async fn get_table(&self, identifier: &Identifier) -> Result<Table>;
```
luoyuxia, line 159: both iceberg-rust and delta-rust return a rust struct instead of rust trait. And struct is simpler to use than trait. Return struct should also be fine to paimon-rust.

PR #96, opened 2026-02-15
Introduce table scan

### LINKED ISSUE #105: introduce table scan
### Search before asking

### Description

as per-title

### Willingness to contribute

- [ ] I'm willing to submit a PR!

luoyuxia:
### Purpose

Linked issue: close #105

### DIFF
--- a/crates/paimon/src/table/source.rs
+++ b/crates/paimon/src/table/source.rs
@@ -34,7 +34,7 @@ pub struct DataSplit {
     partition: BinaryRow,
     bucket: i32,
     bucket_path: String,
-    total_buckets: Option<i32>,
+    total_buckets: i32,
     data_files: Vec<DataFileMeta>,
 }
 
@@ -51,7 +51,7 @@ impl DataSplit {
     pub fn bucket_path(&self) -> &str {
         &self.bucket_path
     }
-    pub fn total_buckets(&self) -> Option<i32> {
+    pub fn total_buckets(&self) -> i32 {
         self.total_buckets
     }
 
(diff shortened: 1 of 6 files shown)


crates/paimon/src/table/source.rs
```diff
@@ -34,7 +34,7 @@ pub struct DataSplit {
     partition: BinaryRow,
     bucket: i32,
     bucket_path: String,
-    total_buckets: Option<i32>,
+    total_buckets: i32,
```
luoyuxia, line 37: java use nuaable Integer for potential null value. But I think it won't be null in new version paimon. So, I remove `Option`.
For reference: https://github.com/apache/paimon/pull/5537 may be a related pr.

crates/paimon/src/table/table_scan.rs
```diff
@@ -0,0 +1,187 @@
...
+        for e in entries {
+            let key = (e.partition().to_vec(), e.bucket());
+            groups.entry(key).or_default().push(e);
+        }
+
+        let snapshot_id = snapshot.id();
+        let base_path = table_path;
+        let mut splits = Vec::new();
+
+        for ((_partition, bucket), group_entries) in groups {
```
QuakeWang, line 154: @luoyuxia Hi, for partitioned tables, `_partition` is dropped right after grouping, but the split is later built with `BinaryRow::new(0)` and `"{table_path}/bucket-{bucket}"`. This is fine for unpartitioned tables, but for partitioned tables, it loses partition identity, and `bucket_path` also misses the partition directory prefix (`k=v/...`). It would be better to reconstruct the partition from the grouped partition bytes, build splits with the real partition, and generate `bucket_path` as `partition_path/bucket-{bucket}`. What do you think of this?

luoyuxia: as todo comment in the following code, I'll create seperate pr to support partitioned table since I want to make each pr as small as possible

QuakeWang: > as todo comment in the following code, I'll create seperate pr to support partitioned table since I want to make each pr as small as possible

I see. Once the tasks are separated, feel free to assign them to me—I'd be delighted to participate.

PR #101, opened 2026-03-01
fix(io): return correct list_status paths and reuse storage operators

QuakeWang:
### Purpose

Fix `FileIO::list_status` path semantics: each returned `FileStatus.path` now points to the actual listed entry instead of the input directory path.
Also, keep reusable operators in `Storage` to preserve memory backend state across calls.

### Brief change log

- Fix `list_status` to return real entry paths (`base_path + entry.path()`), and fetch required metadata via `list_with(...).metakey(...)`.
- Refactor `Storage` to reuse `Operator` instances across calls (including memory backend).
- Add regression tests for `list_status` on both `fs` and `memory` backends.

### Tests

- `cargo clippy --all-targets --workspace -- -D warnings`
- `cargo test -p paimon io::file_io::file_action_test`
- `cargo test --all-targets --workspace`

All tests passed locally.

### API and Format

- No public API signature changes.
- No storage format changes.
- Behavior fix only: `list_status` now returns correct entry paths.

### Documentation

### DIFF
--- a/crates/paimon/src/io/file_io.rs
+++ b/crates/paimon/src/io/file_io.rs
@@ -382,18 +382,24 @@ impl OutputFile {
 
 #[cfg(test)]
 mod file_action_test {
-    use std::collections::{BTreeMap, BTreeSet};
+    use std::collections::BTreeSet;
     use std::fs;
 
     use super::*;
     use bytes::Bytes;
 
     fn setup_memory_file_io() -> FileIO {
-        FileIOBuilder::new("memory").build().unwrap()
+        let storage = Storage::Memory;
+        FileIO {
+            storage: Arc::new(storage),
+        }
     }
 
     fn setup_fs_file_io() -> FileIO {
-        FileIOBuilder::new("file").build().unwrap()
+        let storage = Storage::LocalFs;
+        FileIO {
+            storage: Arc::new(storage),
+        }
     }
 
     async fn common_test_get_status(file_io: &FileIO, path: &str) {
@@ -466,18 +472,14 @@ mod file_action_test {
 
         let file_a = format!("{dir_path}a.txt");
         let file_b = format!("{dir_path}b.txt");
-        file_io
-            .new_output(&file_a)
-            .unwrap()
-            .write(Bytes::from("a"))
-            .await
-            .unwrap();
-        file_io
-            .new_output(&file_b)
-            .unwrap()
-            .write(Bytes::from("b"))
-            .await
-            .unwrap();
+        for file in [&file_a, &file_b] {
+            file_io
+                .new_output(file)
+                .unwrap()
+                .write(Bytes::from("test data"))
+                .await
+                .unwrap();
+        }
 
         let statuses = file_io.list_status(dir_path).await.unwrap();
         assert_eq!(statuses.len(), 2);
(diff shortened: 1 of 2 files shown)


crates/paimon/src/io/file_io.rs
```diff
@@ -137,7 +138,7 @@ impl FileIO {
             statuses.push(FileStatus {
                 size: meta.content_length(),
                 is_dir: meta.is_dir(),
-                path: entry.path().to_string(),
+                path: format!("{base_path}{}", entry.path()),
```
luoyuxia, line 141: IIUC, `format!("{base_path}{}", entry.path())` just add leading `/` to `entry.path()`?
Considering openal will always call `normalize_path`  to remove leading `/`, do we real need to add eading `/` to `entry.path()`? Could you just keep `entry.path()`?

QuakeWang: @luoyuxia Thanks for the comment. `entry.path()` in OpenDAL is relative to operator root, while `FileStatus.path` here is expected to preserve the original FileIO path prefix (e.g., `memory:/` or
  `file:/`) for consistency with `get_status` and follow-up FileIO operations. `normalize_path` applies to Operator input paths, not to `Entry::path()` output.

luoyuxia: Thanks. It sounds reasonable to me to return the full path with scheme prefix.

But it doesn't seem to be necessary now for reusing the storage operator. Maybe you could create a separate PR for reusing the storage operator part in which we can revist it further it.

crates/paimon/src/io/file_io.rs
```diff
@@ -461,6 +457,55 @@ mod file_action_test {
...
+        let expected_paths: BTreeSet<String> =
+            [file_a.clone(), file_b.clone()].into_iter().collect();
+        let actual_paths: BTreeSet<String> =
+            statuses.iter().map(|status| status.path.clone()).collect();
+        assert_eq!(
+            actual_paths, expected_paths,
+            "list_status should return exact entry paths"
+        );
+
+        assert!(
```
luoyuxia: Regarding to this test, Do we really need these following test code:
```
statuses.iter().all(|status| !status.is_dir),
            "listed entries should be files in this test"
        );

        let sizes_by_path: BTreeMap<String, u64> = statuses
            .iter()
            .map(|status| (status.path.clone(), status.size))
            .collect();
        assert_eq!(sizes_by_path.get(&file_a), Some(&1));
        assert_eq!(sizes_by_path.get(&file_b), Some(&1));
```
Can it be removed?

crates/paimon/src/io/file_io.rs
```diff
@@ -461,6 +457,55 @@ mod file_action_test {
...
+    async fn common_test_list_status_paths(file_io: &FileIO, dir_path: &str) {
+        if let Some(local_dir) = dir_path.strip_prefix("file:/") {
+            let _ = fs::remove_dir_all(local_dir);
+        }
+
+        file_io.mkdirs(dir_path).await.unwrap();
+
+        let file_a = format!("{dir_path}a.txt");
+        let file_b = format!("{dir_path}b.txt");
+        file_io
```
luoyuxia: nit:
can be simplify to
```
for file in [&file_a, &file_b] {
            file_io
                .new_output(&file)
                .unwrap()
                .write(Bytes::from("test data"))
                .await
                .unwrap();
        }
```
?

luoyuxia (approved):
@QuakeWang Thanks. +1

PR #103, opened 2026-03-03
feat: introduce snapshot manager

### LINKED ISSUE #30: Impl SnapshotManager
When creating table, will use the SnapshotManager. So we should introduce this util.

luoyuxia:
### Purpose

Linked issue: close #30

### DIFF
--- a/crates/paimon/src/table/snapshot_manager.rs
+++ b/crates/paimon/src/table/snapshot_manager.rs
@@ -1,0 +1,114 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Snapshot manager for reading snapshot metadata using FileIO.
+//!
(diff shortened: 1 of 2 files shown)


crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Path to the snapshot file for the given id (e.g. `snapshot/snapshot-1`).
+    pub fn snapshot_path(&self, snapshot_id: i64) -> String {
+        format!("{}/snapshot-{}", self.snapshot_dir(), snapshot_id)
+    }
+
+    /// Get the latest snapshot, or None if LATEST does not exist.
+    /// Returns an error if LATEST exists but the snapshot file (snapshot-{id}) does not exist.
+    pub async fn get_latest_snapshot(&self) -> crate::Result<Option<Snapshot>> {
```
XiaoHongbo-Hope, line 67: Suggest adding unit tests for get_latest_snapshot

crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub fn new(file_io: FileIO, table_path: String) -> Self {
+        Self {
+            file_io,
+            table_path,
+        }
+    }
+
+    /// Path to the snapshot directory (e.g. `table_path/snapshot`).
+    pub fn snapshot_dir(&self) -> String {
+        format!("{}/{}", self.table_path, SNAPSHOT_DIR)
```
XiaoHongbo-Hope, line 52: Does this work well in Windows?

luoyuxia: I should(but i can't give a try) since rust's `Path` will handle the path across different operator systems. See
https://doc.rust-lang.org/std/path/struct.Path.html
> This type supports a number of operations for inspecting a path, including breaking the path into its components (separated by / on Unix and by either / or \ on Windows), extracting the file name, determining whether the path is absolute, and so on.

crates/paimon/src/table/snapshot_manager.rs
luoyuxia: I was intending to add, but found it's not a simple work since it'll require write to tables. I think we can add when we support  write to tables.

PR #106, opened 2026-03-04
feat(io): refactor FileIO to reuse storage operators and improve path handling

### LINKED ISSUE #104: task(io): reuse storage operator in FileIO
### Search before asking

### Description

In PR #101, we intentionally scoped out the `storage operator reuse` part for easier review.

This issue tracks reusing OpenDAL `Operator` per `FileIO` instance in Rust `Storage`, instead of rebuilding one on every call. The goal is to improve consistency (especially for the memory backend across calls) and reduce unnecessary operator initialization overhead.

Reference from Java Paimon:
  - `ResolvingFileIO` caches FileIO instances.
  - `PluginFileIO` reuses lazy FileIO.
  - `HadoopFileIO` caches FileSystem instances.

### Willingness to contribute

QuakeWang:
### Purpose

Linked issue: close #104

### DIFF
--- a/crates/paimon/src/io/file_io.rs
+++ b/crates/paimon/src/io/file_io.rs
@@ -136,11 +136,7 @@ impl FileIO {
             if entry_path.trim_start_matches('/') == list_path_normalized {
                 continue;
             }
-            // OpenDAL 0.55 removed list metakey selection, so stat each entry to
-            // guarantee FileStatus metadata is populated consistently across backends.
-            let meta = op.stat(entry_path).await.context(IoUnexpectedSnafu {
-                message: format!("Failed to stat listed entry '{}' in '{path}'", entry_path),
-            })?;
+            let meta = entry.metadata();
             statuses.push(FileStatus {
                 size: meta.content_length(),
                 is_dir: meta.is_dir(),


JingsongLi:
I am OK to merge this, cc @luoyuxia to take a look

crates/paimon/src/io/file_io.rs
```diff
@@ -129,20 +132,22 @@ impl FileIO {
...
             let entry_path = entry.path();
-            let entry_path_normalized = entry_path.trim_start_matches('/');
-            if entry_path_normalized == list_path_normalized {
+            if entry_path.trim_start_matches('/') == list_path_normalized {
                 continue;
             }
-            let meta = entry.metadata();
+            // OpenDAL 0.55 removed list metakey selection, so stat each entry to
+            // guarantee FileStatus metadata is populated consistently across backends.
+            let meta = op.stat(entry_path).await.context(IoUnexpectedSnafu {
```
JingsongLi: op.stat is a remote IO? If it is, this is a big performance issue.

QuakeWang: > op.stat is a remote IO? If it is, this is a big performance issue.

@JingsongLi Good point. `op.stat()` can become remote IO on object stores, and doing that for every listed entry would create an N+1 pattern.

 I’ve updated the implementation to avoid per-entry `stat()` in `list_status()`. It now uses `entry.metadata()` directly again, so this PR stays focused on operator euse and does not introduce extra remote calls during listing.

PR #116, opened 2026-03-10
feat: support read to arrow

### LINKED ISSUE #115: support read parquet to arrow
### Search before asking

### Description

as per-title

### Willingness to contribute

- [ ] I'm willing to submit a PR!

luoyuxia:
### Purpose

Linked issue: close #115

### DIFF
--- a/crates/paimon/src/arrow/reader.rs
+++ b/crates/paimon/src/arrow/reader.rs
@@ -24,13 +24,14 @@ use futures::future::BoxFuture;
 use futures::{StreamExt, TryFutureExt};
 use parquet::arrow::arrow_reader::ArrowReaderOptions;
 use parquet::arrow::async_reader::{AsyncFileReader, MetadataFetch};
-use parquet::file::metadata::ParquetMetaDataReader;
 use parquet::arrow::ParquetRecordBatchStreamBuilder;
 use parquet::file::metadata::ParquetMetaData;
+use parquet::file::metadata::ParquetMetaDataReader;
 use std::ops::Range;
 use std::sync::Arc;
 use tokio::try_join;
 
+/// Builder to create ArrowReader
 pub struct ArrowReaderBuilder {
     batch_size: Option<usize>,
     file_io: FileIO,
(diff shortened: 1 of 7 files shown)


crates/paimon/src/arrow/reader.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// Take a stream of DataSplits and read every data file in each split.
+    /// Returns a stream of Arrow RecordBatches from all files.
+    pub fn read(self, data_splits: &[DataSplit]) -> crate::Result<ArrowRecordBatchStream> {
+        let file_io = self.file_io.clone();
+        let batch_size = self.batch_size;
+        let paths_to_read: Vec<String> = data_splits
+            .iter()
+            .flat_map(|ds| ds.data_file_entries().map(|(p, _)| p))
+            .map(|p| {
+                if !p.to_ascii_lowercase().ends_with(".parquet") {
```
XiaoHongbo-Hope, line 75: only parquet?

XiaoHongbo-Hope, line 75: I can get your point, only parquet is supported. But I am afraid that this check is not enough.  See  `DataFilePathFactory.formatIdentifier`. Or can we have a better way to get the format.

luoyuxia: Yes, let's only support parquet now. We can support more in next following pr.

XiaoHongbo-Hope: > 

👌

luoyuxia: I understand the concern. However, for parquet files we normally do not expect extra compression suffixes like .gz, since compression is handled within the parquet format itself.

So for the current parquet-only read path, I think checking for `.parquet` should be sufficient. If we later need to support other file naming conventions, we can revisit this and align with `DataFilePathFactory.formatIdentifier`.

Btw, From the Python side, it seems we only inspect the last extension via `os.path.splitext()`. That means a path like `*.json.gz` would be identified as gz, not json.

So the current behavior there does not appear to handle compressed suffixes in a generalized way either.

For parquet, this is less of a concern since we normally do not expect an extra compression suffix such as .parquet.gz.

XiaoHongbo-Hope: > 

Thanks for your very kind explanation. And thanks again for pointing out the issue in python side, I wiil check it.

Issue #131, opened 2026-03-13
Support partitioned table in TableScan and TableRead


luoyuxia:
## Parent Issue
Part of #124 (support partitioned table)
Depends on #126 (BinaryRow deserialization), #127 (partition path generation)

## Background

Currently, `TableScan` and `TableRead` do not support partitioned tables:

1. **`TableScan::plan_snapshot()`** discards partition info and builds wrong bucket paths (`{base_path}/bucket-{bucket}` instead of `{base_path}/{partition_path}/bucket-{bucket}`)
2. **`TableRead::to_arrow()`** explicitly rejects partitioned tables with `Unsupported` error
3. No integration tests for partitioned table reading

### 1. Fix TableScan to generate correct partition paths

- Pass partition type info (from `TableSchema`) into `plan_snapshot()`, or change it to an instance method that can access `self.table.schema`
- Decode partition bytes from `ManifestEntry` into `BinaryRow` using `BinaryRow::from_bytes(arity, data)`
- Use partition path utils (from #127) to compute the partition path segment
- Construct `bucket_path` as `{table_path}/{partition_path}/bucket-{bucket}`
- Store actual decoded `BinaryRow` in `DataSplit` instead of empty `BinaryRow::new(0)`

### 2. Remove partitioned table read restriction in TableRead

- Remove the partition key check in `TableRead::to_arrow()` (`read_builder.rs:88-95`)
- ArrowReader should work without changes once paths are correct

### 3. Add integration tests

- Prepare test fixtures for partitioned tables (single and multiple partition keys)
- Integration test: read a partitioned table end-to-end, verify DataSplits have correct `bucket_path` with partition segments
- Verify partition column values are correct in returned RecordBatches

## Affected files
- `crates/paimon/src/table/table_scan.rs` — `plan_snapshot()` method
- `crates/paimon/src/table/read_builder.rs` — `TableRead::to_arrow()` method
- `crates/integration_tests/tests/` — new test file(s)

Issue #127, opened 2026-03-13
Implement partition path generation (PartitionPathUtils)


luoyuxia:
## Parent Issue
Part of #124 (support partitioned table)
Depends on #126 (BinaryRow deserialization)

## Background

Paimon stores data files under partition directories like `{table_path}/dt=2024-01-01/hr=12/bucket-0/`. Currently there is no code to convert partition values (from `BinaryRow`) into this filesystem path format.

## What needs to be done

1. **Create a partition path utility** (e.g., `spec/partition_utils.rs` or similar)
   - Given partition keys (names from `TableSchema.partition_keys()`) + partition field types (from `TableSchema.fields()`) + partition values (decoded `BinaryRow`), generate the partition path string
   - Format: `{key1}={value1}/{key2}={value2}/...`

2. **Handle value formatting per data type**
   - `Int/BigInt/SmallInt/TinyInt` → integer string
   - `VarChar/Char/String` → string value
   - `Date` → `yyyy-MM-dd` format (days since epoch)
   - `Timestamp` → appropriate timestamp string
   - `Boolean` → `true`/`false`
   - `Float/Double` → decimal string
   - `Decimal` → decimal string

3. **Handle special cases**
   - Null partition values → use `__DEFAULT_PARTITION__`
   - Special characters in string values → URL encoding if needed (follow Java Paimon's behavior)

4. **Unit tests**
   - Single partition key
   - Multiple partition keys
   - Different data types (string, int, date)
   - Null partition value → `__DEFAULT_PARTITION__`

## Reference
- Java: [`PartitionPathUtils`](https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/utils/PartitionPathUtils.java)

## Affected files
- New file: `crates/paimon/src/spec/partition_utils.rs` (or integrated into existing module)

PR #133, opened 2026-03-14
feat: implement BinaryRow deserialization for partition bytes

### LINKED ISSUE #126: Implement BinaryRow deserialization to decode partition binary bytes
## Parent Issue
Part of #124 (support partitioned table)

## Background

In Paimon, `ManifestEntry._PARTITION` stores partition values as raw `BinaryRow` binary bytes (`Vec<u8>`). Currently, `BinaryRow` (`crates/paimon/src/spec/data_file.rs:30-51`) is a stub that only stores `arity` — it cannot hold or parse actual partition data.

To support partitioned table reading, we need `BinaryRow` to wrap the raw bytes and provide typed accessor methods.

## What needs to be done

1. **Enhance `BinaryRow` to hold actual binary data**
   - Add a `data: Vec<u8>` (or `Bytes`) field to back the row
   - Parse the binary layout: header (8 bytes) + null bit set (aligned) + fixed-length part (8 bytes per field) + variable-length part

2. **Implement typed getters**
   - `is_null_at(pos)` — check the null bit
   - `get_int(pos)` → `i32`
   - `get_long(pos)` → `i64`
   - `get_short(pos)` → `i16`
   - `get_byte(pos)` → `i8`
   - `get_boolean(pos)` → `bool`
   - `get_float(pos)` → `f32`
   - `get_double(pos)` → `f64`
   - `get_string(pos)` → `&str` / `String` (variable-length, read offset+length from fixed part, then read from variable part)
   - `get_binary(pos)` → `&[u8]`

3. **Add a constructor from raw bytes**
   - `BinaryRow::from_bytes(arity: i32, data: Vec<u8>) -> BinaryRow`

4. **Unit tests**
   - Round-trip: construct binary data manually, verify typed getters
   - Test null handling
   - Test variable-length fields (String, Binary)

## Reference
- Java: [`BinaryRow`](https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/data/BinaryRow.java)
- Java: [`BinaryRowSerializer`](https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/data/serializer/BinaryRowSerializer.java)

## Affected files
- `crates/paimon/src/spec/data_file.rs` — `BinaryRow` struct

QuakeWang:
### Purpose

Linked issue: close #126 

Implement `BinaryRow` deserialization primitives so partition binary bytes from Paimon manifests can be decoded in Rust.

This PR focuses on the `BinaryRow` capability itself only. Wiring manifest `_PARTITION` bytes into `TableScan` / `TableRead` is intentionally left to #131, and partition path generation is handled separately in #127.

### API and Format

 This change adds `BinaryRow` decoding APIs but does not change the external storage format.

  The implementation follows Java Paimon's existing `BinaryRow` binary layout, including null-bit layout and inline / variable-length encoding semantics.

### DIFF
--- a/crates/paimon/src/spec/data_file.rs
+++ b/crates/paimon/src/spec/data_file.rs
@@ -58,7 +58,7 @@ pub struct BinaryRow {
 
     /// Raw binary data backing this row. Empty when constructed via `new()`.
     /// Populated via `from_bytes()` for partition data from manifest entries.
-    #[serde(skip)]
+    #[serde(with = "serde_bytes")]
     data: Vec<u8>,
 }
 
@@ -524,13 +524,29 @@ mod tests {
     }
 
     #[test]
-    fn test_serde_roundtrip_backward_compat() {
-        // Verify BinaryRow::new(0) serde roundtrip is stable.
+    fn test_serde_roundtrip_empty() {
+        // Verify empty BinaryRow serde roundtrip is stable.
         let row = BinaryRow::new(0);
         let json = serde_json::to_string(&row).unwrap();
         let deserialized: BinaryRow = serde_json::from_str(&json).unwrap();
         assert_eq!(deserialized.arity(), 0);
         assert!(deserialized.is_empty());
+    }
+
+    #[test]
+    fn test_serde_roundtrip_populated() {
+        // Verify a populated BinaryRow roundtrips correctly with data intact.
+        let mut builder = BinaryRowBuilder::new(2);
+        builder.write_int(0, 42);
+        builder.write_string(1, "hello");
+        let row = builder.build();
+
+        let json = serde_json::to_string(&row).unwrap();
+        let deserialized: BinaryRow = serde_json::from_str(&json).unwrap();
+        assert_eq!(deserialized.arity(), row.arity());
+        assert_eq!(deserialized.data(), row.data());
+        assert_eq!(deserialized.get_int(0), 42);
+        assert_eq!(deserialized.get_string(1), "hello");
     }
 
     #[test]


crates/paimon/src/spec/data_file.rs
```diff
@@ -24,30 +24,203 @@ use std::fmt::{Display, Formatter};
...
-#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct BinaryRow {
     arity: i32,
     null_bits_size_in_bytes: i32,
+
+    /// Raw binary data backing this row. Empty when constructed via `new()`.
+    /// Populated via `from_bytes()` for partition data from manifest entries.
+    #[serde(skip)]
```
luoyuxia: why skip serde for this field?

QuakeWang: > why skip serde for this field?

The data field is populated at runtime via `BinaryRow::from_bytes()`, not through serde. In the Avro manifest, `_PARTITION` is stored as raw bytes (with a 4-byte arity prefix, per Java's `SerializationUtils.serializeBinaryRow`). The `#[serde(skip)]` is here to avoid breaking existing JSON serde for `DataFileMeta._MIN_KEY`/`_MAX_KEY`, which only serializes arity and `null_bits_size_in_bytes`.

luoyuxia: won't `SerializationUtils.serializeBinaryRow` serialize the data bytes?

QuakeWang: > won't `SerializationUtils.serializeBinaryRow` serialize the data bytes?

Yes, `SerializationUtils.serializeBinaryRow` does serialize the full row bytes (including the 4-byte arity prefix), but that is a separate binary codec used to fill a parent bytes field such as `_PARTITION`.

On the Rust side, `_PARTITION` is first deserialized into `ManifestEntry.partition: Vec<u8>`, and only then converted to `BinaryRow` via `BinaryRow::from_bytes(...)` at runtime.

`#[serde(skip)]` only affects `BinaryRow`'s own derived `Serialize`/`Deserialize`. I added it so introducing this runtime backing buffer does not change BinaryRow's current serde shape or the existing empty-row roundtrip behavior. If we ever need serde to roundtrip a populated `BinaryRow`, we should switch to a dedicated custom serializer instead of relying on the derived struct serde.

luoyuxia: Currently BinaryRow's Serialize/Deserialize is not used anywhere — DataSplit only derives Debug, and DataFileMeta._MIN_KEY/_MAX_KEY are Vec<u8>, not BinaryRow.

The #[serde(skip)] on data is problematic: if we later need DataSplit (which contains BinaryRow) to be serializable (e.g. for distributed task dispatch), the data field must participate in serde — otherwise the receiver gets an empty partition. Skipping it now risks a subtle bug later.

I'd suggest either:

Remove Serialize/Deserialize and #[serde(skip)] entirely since nothing uses them today, or
Keep Serialize/Deserialize but remove #[serde(skip)] (use #[serde(with = "serde_bytes")] instead) so that when DataSplit eventually derives Serialize/Deserialize, BinaryRow is already correct and we don't forget to fix this.

QuakeWang: > Currently BinaryRow's Serialize/Deserialize is not used anywhere — DataSplit only derives Debug, and DataFileMeta._MIN_KEY/_MAX_KEY are Vec, not BinaryRow.
> 
> The #[serde(skip)] on data is problematic: if we later need DataSplit (which contains BinaryRow) to be serializable (e.g. for distributed task dispatch), the data field must participate in serde — otherwise the receiver gets an empty partition. Skipping it now risks a subtle bug later.
> 
> I'd suggest either:
> 
> Remove Serialize/Deserialize and #[serde(skip)] entirely since nothing uses them today, or Keep Serialize/Deserialize but remove #[serde(skip)] (use #[serde(with = "serde_bytes")] instead) so that when DataSplit eventually derives Serialize/Deserialize, BinaryRow is already correct and we don't forget to fix this.

Good point, thanks for the review. I adopted option B by replacing `#[serde(skip)` with `#[serde(with = "serde_bytes")]`, so `BinaryRow.data` is preserved in Rust serde roundtrips.

I also added test_serde_roundtrip_populated to verify that a populated BinaryRow roundtrips with its backing bytes intact, in addition to the empty-row case.

PR #139, opened 2026-03-16
feat(spec): implement partition path generation

### LINKED ISSUE #127: Implement partition path generation (PartitionPathUtils)

QuakeWang:
### Purpose

Linked issue: close #127

### DIFF
--- a/crates/paimon/src/spec/data_file.rs
+++ b/crates/paimon/src/spec/data_file.rs
@@ -124,67 +124,82 @@ impl BinaryRow {
         (self.data[byte_index] & (1 << bit_offset)) != 0
     }
 
-    // ======================== Fixed-length getters ========================
+    // ======================== Internal read helpers ========================
 
     /// Byte offset of the field value at position `pos` within `self.data`.
     fn field_offset(&self, pos: usize) -> usize {
         self.null_bits_size_in_bytes as usize + pos * 8
     }
 
+    /// Bounds-checked slice read. Returns exactly `N` bytes starting at `offset`.
+    fn read_slice<const N: usize>(&self, offset: usize) -> crate::Result<[u8; N]> {
+        self.data
+            .get(offset..offset + N)
+            .and_then(|s| s.try_into().ok())
+            .ok_or_else(|| crate::Error::UnexpectedError {
+                message: format!(
+                    "BinaryRow: read {N} bytes at offset {offset} exceeds data length {}",
+                    self.data.len()
+                ),
+                source: None,
+            })
+    }
+
+    /// Bounds-checked single byte read.
+    fn read_byte_at(&self, offset: usize) -> crate::Result<u8> {
+        self.data
+            .get(offset)
+            .copied()
+            .ok_or_else(|| crate::Error::UnexpectedError {
+                message: format!(
+                    "BinaryRow: read 1 byte at offset {offset} exceeds data length {}",
+                    self.data.len()
+                ),
+                source: None,
+            })
+    }
+
     /// Read a little-endian i64 from `self.data` at the given byte offset.
-    fn read_i64_at(&self, offset: usize) -> i64 {
-        let bytes: [u8; 8] = self.data[offset..offset + 8]
-            .try_into()
-            .expect("BinaryRow: insufficient data for i64 read");
-        i64::from_le_bytes(bytes)
+    fn read_i64_at(&self, offset: usize) -> crate::Result<i64> {
+        self.read_slice::<8>(offset).map(i64::from_le_bytes)
     }
 
     /// Read a little-endian i32 from `self.data` at the given byte offset.
-    fn read_i32_at(&self, offset: usize) -> i32 {
-        let bytes: [u8; 4] = self.data[offset..offset + 4]
-            .try_into()
-            .expect("BinaryRow: insufficient data for i32 read");
-        i32::from_le_bytes(bytes)
+    fn read_i32_at(&self, offset: usize) -> crate::Result<i32> {
+        self.read_slice::<4>(offset).map(i32::from_le_bytes)
     }
 
-    pub fn get_boolean(&self, pos: usize) -> bool {
-        self.data[self.field_offset(pos)] != 0
+    // ======================== Fixed-length getters ========================
+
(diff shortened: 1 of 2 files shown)


crates/paimon/src/spec/data_file.rs
```diff
@@ -212,15 +212,123 @@ impl BinaryRow {
...
+        })
+    }
+
+    /// Read the unscaled value of a Decimal field as `i128`.
+    ///
+    /// - `precision <= 18` (compact): stored as `i64` in the fixed part.
+    /// - `precision > 18`: stored as big-endian two's complement bytes in the variable-length part.
+    ///
+    /// Reference: `BinaryRow.getDecimal` and `DecimalData.isCompact` in Java Paimon.
+    pub(crate) fn get_decimal_unscaled(&self, pos: usize, precision: u32) -> crate::Result<i128> {
```
luoyuxia, line 270: some get method return Result, but some get method like `get_binary` just panic if any error happen. 
I'd like to suggest to unify it to make all get method return  Result.
No try_get, only get, but return result. Such pattern has discussed in fluss-rust which we reach to agreement that return Result will be better.

crates/paimon/src/spec/data_file.rs
```diff
@@ -212,15 +212,123 @@ impl BinaryRow {
...
         let bytes = self.get_binary(pos);
         std::str::from_utf8(bytes).expect("BinaryRow: invalid UTF-8 in string field")
     }
+
+    /// Bounds-checked version of [`get_string`]. Returns `Err` on corrupted data or invalid UTF-8.
+    pub(crate) fn try_get_string(&self, pos: usize) -> crate::Result<&str> {
+        let bytes = self.try_get_binary(pos)?;
+        std::str::from_utf8(bytes).map_err(|e| crate::Error::UnexpectedError {
+            message: format!("BinaryRow: invalid UTF-8 in string field at pos {pos}: {e}"),
+            source: None,
```
luoyuxia: nit: source: Some(Box::new(e)),

crates/paimon/src/spec/partition_utils.rs
```diff
@@ -0,0 +1,955 @@
...
+/// # Arguments
+///
+/// * `partition_keys` - Partition column names, defining the order.
+/// * `schema_fields` - All fields from the table schema, used to look up types.
+/// * `row` - The partition `BinaryRow` whose arity matches `partition_keys.len()`.
+/// * `default_partition_name` - Value for null or blank partition values (e.g. `__DEFAULT_PARTITION__`).
+/// * `legacy_partition_name` - Whether to use legacy (Java `toString()`) formatting for timestamps.
+// TODO: remove after #131 consumes the pub(crate) API.
+#[allow(dead_code)]
+pub(crate) fn generate_partition_path(
```
luoyuxia: I'd like to suggest have a seperate method called `generate_part_values` like java since I believe such method should be shared by others in the future. And it make the code cleaner. You may also need to have a struct `InternalRowPartitionComputer`

crates/paimon/src/spec/partition_utils.rs
```diff
@@ -0,0 +1,955 @@
...
+
+    let value = match data_type {
+        DataType::Boolean(_) => row.get_boolean(pos).to_string(),
+        DataType::TinyInt(_) => row.get_byte(pos).to_string(),
+        DataType::SmallInt(_) => row.get_short(pos).to_string(),
+        DataType::Int(_) => row.get_int(pos).to_string(),
+        DataType::BigInt(_) => row.get_long(pos).to_string(),
+        // NOTE: Rust f32/f64 Display may differ from Java Float/Double.toString()
+        // on edge-case values. Acceptable since float partition keys are rare.
+        DataType::Float(_) => row.get_float(pos).to_string(),
```
luoyuxia: maybe we can return unsupported error directly since float partition keys are rare and there may be data loss risk in here.

crates/paimon/src/spec/partition_utils.rs
```diff
@@ -0,0 +1,955 @@
...
+
+    #[test]
+    fn test_empty_partition_keys() {
+        let row = BinaryRow::new(0);
+        let result = generate_partition_path(&[], &[], &row, DEFAULT_PARTITION_NAME, true).unwrap();
+        assert_eq!(result, "");
+    }
+
+    #[test]
+    fn test_single_string_partition() {
```
luoyuxia, line 734: nit:
the tests have a lot of repeated setup boilerplate (fields, keys, TestRowBuilder,
  generate_partition_path(...)), which makes the suite harder to extend and review. It would be nice to extract a small helper or two for the
  common single-column / assertion flow,

PR #151, opened 2026-03-23
feat: add code quality check script and license config

discivigour:
### Purpose

Add code quality check script and license config.

### API and Format

- ./dev/check.sh

### DIFF
--- a/dev/check.sh
+++ b/dev/check.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -e
+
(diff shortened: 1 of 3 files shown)


dev/check.sh
```diff
@@ -0,0 +1,102 @@
...
+
+echo "========================================"
+echo "Running code checks for paimon-rust"
+echo "========================================"
+
+# 1. Check license headers
+echo ""
+echo "[1/4] Checking license headers..."
+if command -v license-eye &> /dev/null; then
+    license-eye header check
```
QuakeWang, line 50: This project already uses [skywalking-eyes](https://github.com/apache/paimon-rust/blob/main/.github/workflows/ci.yml#L38) to check for licenses.

discivigour: Where is the skywalking-eyes? You mean ci.yml? I added this script to facilitate local verification.

QuakeWang: > Where is the skywalking-eyes? You mean ci.yml? I added this script to facilitate local verification.

Nice initiative! For the "facilitate local verification" goal, a [pre-commit](https://pre-commit.com/) hook might be more appropriate — it runs automatically on every commit, so contributors can't forget. Consider adding a `.pre-commit-config.yaml` for fmt/clippy checks. The Makefile targets and `.licenserc.yaml` are still valuable and can be kept. The `dev/check.sh` script could be simplified or removed if pre-commit covers the same ground.

discivigour: Thanks, I will try it.

discivigour: Thanks for the suggestion! I did look into pre-commit hooks and `.pre-commit-config.yaml` before going with `dev/check.sh`. 

I don't think it's necessary to introduce git hooks at present. Because it cannot be uploaded to the repository, or relies on external plugins, or is manually configured, and the current repository is not very large at present. 

dev/check.sh provides an easy, zero-dependency way for contributors to verify locally, and CI serves as the final safeguard. I think this combination is sufficient for the current project scale without introducing additional tooling overhead. We can revisit this if the project grows and the team scales.

PR #156, opened 2026-03-29
feat(spec): define Predicate data structure and PredicateBuilder

### LINKED ISSUE #154: feat: Define Predicate data structure and PredicateBuilder
### Search before asking

### Motivation

Predicate pushdown requires a reusable filter representation that can be shared by partition pruning, manifest statistics pruning, and data file statistics pruning.

This issue adds the foundational predicate types and field-mapping utilities with no behavior change in the read path.

### Solution

Add the following types to a new `crates/paimon/src/spec/predicate.rs` module:

- `Datum` — typed literal enum covering all Paimon partition-supported types (Bool, Int, Long, String, Date, Timestamp, Decimal, etc.)
- `PredicateOperator` — IsNull, IsNotNull, Eq, NotEq, Lt, LtEq, Gt, GtEq, In, NotIn
- `Predicate` — filter tree enum with `Leaf` (carrying column name, table schema index, `DataType`, operator, and literals), `And`, `Or`, `Not`, `AlwaysTrue`, `AlwaysFalse`
- `PredicateBuilder` — stores schema metadata, validates column references in builder methods (`equal()`, `less_than()`, etc.)
- `field_idx_to_partition_idx()` — maps table schema indices to partition row indices (`Vec<Option<usize>>`)
- Predicate composition helpers: `and()`, `or()`, `not()` with auto-flattening

Key design choice: `Leaf` carries both `index` (table schema position) and `data_type` because partition `BinaryRow` uses a separate index space and value extraction requires knowing the type. The mapping from table schema index to partition row index is handled by `field_idx_to_partition_idx()`.

### Anything else?

None

### Willingness to contribute

QuakeWang:
### Purpose

Linked issue: close #154 

Add foundational predicate types and field-mapping utilities for filter pushdown. This is infrastructure-only — no behavior change in the read path.

### API and Format

New public API added:
- Datum, PredicateOperator, Predicate, PredicateBuilder, field_idx_to_partition_idx

### DIFF
--- a/crates/paimon/src/spec/predicate.rs
+++ b/crates/paimon/src/spec/predicate.rs
@@ -37,7 +37,13 @@ use std::fmt;
 /// Each variant corresponds to one or more Paimon `DataType`s and carries the
 /// internal representation used by `BinaryRow`. This avoids untyped `Object`
 /// boxing (as in Java Paimon) and provides compile-time safety.
-#[derive(Debug, Clone, PartialEq)]
+///
+/// `PartialEq` is manually implemented so that `Decimal` uses mathematical
+/// equivalence (matching Java Paimon's `Decimal` which uses `compareTo() == 0`
+/// rather than `BigDecimal.equals` which is scale-sensitive),
+/// e.g. `Decimal(10, scale=1)` == `Decimal(100, scale=2)`
+/// because both represent `1.0`.
+#[derive(Debug, Clone)]
 pub enum Datum {
     Bool(bool),
     TinyInt(i8),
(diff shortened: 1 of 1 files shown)


QuakeWang:
> One noteworthy aspect is the equivalence comparison of Decimal. In Java, Decimal uses `compareTo` to achieve equivalence, which can maintain equality even at different scales.

Good catch! I have fixed this.

`Datum` now has a manual `PartialEq` impl instead of `derive(PartialEq)`. The `Decimal` variant normalizes both sides to a common scale before comparing unscaled values, matching Java `Decimal.equals()` which delegates to `compareTo() == 0`

For example, `Decimal(unscaled=10, scale=1)` and `Decimal(unscaled=100, scale=2)` are now considered equal since both represent `1.0`.

Issue #173, opened 2026-04-01
umbrella: datafusion support


JingsongLi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Motivation

TODO List:
- [x] Predicte pushdown
- [x] table scan refactor: bucket pruning, Limit with predicate
- [x] Data level stats pruning
- [x] CatalogProvider - Significant improvement in user experience, no need to manually register each table
- [x] Limit pushdown - easy to implement, greatly improving the interactive query experience
- [x] Nested Types - Expand the Scope of Queryable Tables
- [x] Schema Evolution: data reader.
- [x] Paimon Table types Cover: Data Evolution mode support
- [x] Time travel - snapshot id
- [x] Time travel - tag name
- [x] Use multiple-threads read manifest
- [x] filter pushdown to data read
- [x] Support data evolution row id filter
- [x] Python Datafusion API

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #176, opened 2026-04-01
feat: support catalog provider in datafusion

luoyuxia:
### Purpose

Linked issue: close #xxx
as a part of #173

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -67,10 +67,16 @@ impl CatalogProvider for PaimonCatalogProvider {
     }
 
     fn schema(&self, name: &str) -> Option<Arc<dyn SchemaProvider>> {
-        Some(Arc::new(PaimonSchemaProvider::new(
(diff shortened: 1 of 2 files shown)


crates/integrations/datafusion/src/catalog.rs
```diff
@@ -0,0 +1,142 @@
...
+        self
+    }
+
+    fn schema_names(&self) -> Vec<String> {
+        futures::executor::block_on(async {
+            self.catalog.list_databases().await.unwrap_or_default()
+        })
+    }
+
+    fn schema(&self, name: &str) -> Option<Arc<dyn SchemaProvider>> {
```
QuakeWang, line 69: `schema()` currently returns `Some(...)` for any name, which makes a non-existent database look like a resolvable schema. On the DataFusion side, the expected behavior is to return `None` when the schema does not exist. Otherwise, downstream code ends up turning “missing schema” into “table not found” or an empty schema. 

I reproduced this locally with a missing database. It would be better to check whether the database exists first and return None if it does not.

PR #177, opened 2026-04-01
feat: support limit push down in datafusion

luoyuxia:
### Purpose

Linked issue: close #xxx
as a part of https://github.com/apache/paimon-rust/issues/173

### DIFF
--- a/crates/paimon/src/table/source.rs
+++ b/crates/paimon/src/table/source.rs
@@ -22,8 +22,8 @@
 #![allow(dead_code)]
 
 use crate::spec::{BinaryRow, DataFileMeta};
+use crate::table::table_scan::group_by_overlapping_row_id;
 use serde::{Deserialize, Serialize};
-
 // ======================= DeletionFile ===============================
 
 /// Deletion file for a data file: describes a region in a file that stores deletion vector bitmap.
@@ -261,7 +261,7 @@ impl DataSplit {
         }
 
         // Merge overlapping row ID ranges and compute max row_count per group
-        let groups = merge_overlapping_row_id_ranges(&self.data_files);
+        let groups = group_by_overlapping_row_id(self.data_files.to_vec());
         let sum: i64 = groups
             .iter()
             .map(|group| group.iter().map(|f| f.row_count).max().unwrap_or(0))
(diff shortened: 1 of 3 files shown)


crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -133,9 +144,30 @@ impl ExecutionPlan for PaimonTableScan {
         };
 
         let stream = futures::stream::once(fut).try_flatten();
+
+        // Apply limit if specified
+        let limited_stream: Pin<Box<dyn Stream<Item = DFResult<RecordBatch>> + Send>> =
```
JingsongLi, line 154: If it's just a limit applied to this stream, will datafusion do it itself?

luoyuxia: Yes, DataFusion already does this on its own — it just stops early at the scan source in current pr. My original plan was to leave the limit pushdown to Paimon core for a follow-up PR, but we can also implement it directly in this PR if you prefer.

JingsongLi: You can continue to complete it, please note that there will be special logic for pushing down the limit in data evolution mode.

crates/paimon/src/table/table_scan.rs
```diff
@@ -313,6 +390,26 @@ impl<'a> TableScan<'a> {
...
         }
 
+        // --- Limit pushdown at manifest entry level (for append-only tables) ---
+        //
+        // For simple append-only tables without deletion vectors or data evolution,
+        // we can apply limit early at the manifest entry level. This is more efficient
+        // than waiting until splits are created.
+        //
+        // Reference: [AppendOnlyFileStoreScan.postFilterManifestEntries](https://github.com/apache/paimon/blob/release-1.3/paimon-core/src/main/java/org/apache/paimon/operation/AppendOnlyFileStoreScan.java#L91)
+        let limit_pushdown_at_manifest_level = self.table.schema.primary_keys().is_empty()
```
JingsongLi: Maybe we don't need this. Just do limit in `apply_limit_pushdown`.

crates/paimon/src/table/source.rs
```diff
@@ -339,3 +423,60 @@ impl Plan {
...
+
+// ======================= Helper Functions ===============================
+
+/// Merges overlapping row ID ranges into groups.
+///
+/// Files are sorted by `(start, end)` of their row ID range, then grouped
+/// by overlapping ranges. Two ranges overlap if `current.start <= previous_group_end`.
+///
+/// Reference: [RangeHelper.mergeOverlappingRanges()](https://github.com/apache/paimon/blob/release-1.3/paimon-common/src/main/java/org/apache/paimon/utils/RangeHelper.java#L59)
+fn merge_overlapping_row_id_ranges(files: &[DataFileMeta]) -> Vec<Vec<&DataFileMeta>> {
```
JingsongLi: Just reuse `group_by_overlapping_row_id`?

crates/paimon/src/table/table_scan.rs
```diff
@@ -246,6 +278,51 @@ impl<'a> TableScan<'a> {
...
 
+    /// Apply a limit-pushdown hint to the generated splits.
+    ///
+    /// Iterates through splits and accumulates `merged_row_count()` until the
+    /// limit hint is reached. Returns only the splits likely needed to satisfy
+    /// that hint.
+    ///
+    /// This does not guarantee an exact final row count. If any split's
+    /// `merged_row_count()` is `None` (for example because of unknown deletion
+    /// cardinality), all remaining splits are kept and the caller or query
```
JingsongLi: `all remaining splits are kept`?

crates/paimon/src/table/source.rs
luoyuxia: Thanks Not noticed it.

PR #190, opened 2026-04-02
feat(datafusion): support partition predicate pushdown

### LINKED ISSUE #172: feat(datafusion): support partition predicate pushdown
### Search before asking

### Description

The current DataFusion integration ignores pushed-down filters and always plans a full Paimon scan.

This task is to add partition predicate pushdown support in `paimon-datafusion` by translating exact partition-only DataFusion filters and wiring them into the existing Paimon scan planning path.

For a partitioned table, queries such as:

```sql
SELECT * FROM t WHERE dt = '2024-01-01'
SELECT * FROM t WHERE dt = '2024-01-01' AND hr = 10
```

should only plan matching partitions.

Unsupported or non-partition filters should remain as residual DataFusion filters.

Out of scope:
- data predicate pushdown into TableRead
- manifest stats pruning
- data file stats pruning
- limit pushdown

This task depends on the core partition predicate pruning support from issue #155 / PR #167.

### Willingness to contribute

QuakeWang:
### Purpose

Linked issue: close #172

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -37,10 +37,7 @@ github:
     discussions: true
     wiki: false
     projects: false
-  protected_branches:
-    main:
-      required_pull_request_reviews:
-        required_approving_review_count: 0
+  protected_branches: ~
   collaborators:
     - Xuanwo
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,13 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - name: Validate .asf.yaml
         run: python3 scripts/validate_asf_yaml.py
 
       - name: Check License Header
-        uses: apache/skywalking-eyes/header@v0.6.0
+        uses: apache/skywalking-eyes/header@v0.8.0
 
       - name: Install cargo-deny
         uses: taiki-e/install-action@v2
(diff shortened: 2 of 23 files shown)


crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -0,0 +1,525 @@
...
+        TableProviderFilterPushDown::Unsupported
+    }
+}
+
+pub(crate) fn build_pushed_predicate(
+    filters: &[Expr],
+    fields: &[DataField],
+    partition_keys: &[String],
+) -> Option<Predicate> {
+    let translator = PartitionFilterTranslator::new(fields, partition_keys);
```
JingsongLi: Why we just limit this to partition filter? Just pass all filters and let scan deal with this.

QuakeWang: > Why we just limit this to partition filter? Just pass all filters and let scan deal with this.

@JingsongLi  Thanks for the review. You're right, this is the wrong boundary.

The current implementation couples "which filters are translated/passed to scan" with "which filters are currently used for partition pruning". `TableScan` already accepts a full table-schema `Predicate` and extracts partition conjuncts internally, so restricting translation here to partition columns is too narrow.

I'll update the DataFusion side to pass all translatable filters down to scan, while keeping `Exact` only for predicates we can evaluate exactly today. Filters that are translatable but not fully enforced by the scan yet will remain `Inexact`.

JingsongLi: Cool~

QuakeWang: @JingsongLi I have updated, PTAL~

PR #193, opened 2026-04-02
feat: support data evolution table mode

JingsongLi:
### Purpose

Supports https://paimon.apache.org/docs/master/append-table/data-evolution/

Sub task of #173

### DIFF
--- a/crates/paimon/src/table/table_scan.rs
+++ b/crates/paimon/src/table/table_scan.rs
@@ -26,7 +26,7 @@ use crate::spec::{
     eval_row, field_idx_to_partition_idx, BinaryRow, CoreOptions, DataFileMeta, FileKind,
     IndexManifest, ManifestEntry, PartitionComputer, Predicate, Snapshot,
 };
-use crate::table::bin_pack::{pack_for_ordered, split_for_batch};
+use crate::table::bin_pack::split_for_batch;
 use crate::table::source::{DataSplitBuilder, DeletionFile, PartitionBucket, Plan};
 use crate::table::SnapshotManager;
 use crate::Error;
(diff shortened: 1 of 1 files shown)


crates/paimon/src/arrow/reader.rs
```diff
@@ -184,6 +188,259 @@ impl ArrowReader {
...
+        let table_field_names: Vec<String> =
+            table_fields.iter().map(|f| f.name().to_string()).collect();
+        let projected_column_names: Vec<String> = read_type
+            .iter()
+            .map(|field| field.name().to_string())
+            .collect();
+
+        Ok(try_stream! {
+            for split in &splits {
+                if split.raw_convertible() || split.data_files().len() == 1 {
```
QuakeWang, line 163: The raw-convertible / single-file branch yields `read_single_file()` directly, but this helper does not reorder columns back to `projected_column_names` after `ProjectionMask`, unlike the existing `read()` path.

Parquet returns file-schema order, not request order, so a projection like `["value", "id"]` will come back as `["id", "value"]` for single-file splits. That makes this path inconsistent with the current `read()` behavior, and it can also make the new projection test flaky when the plan contains single-file groups.

Should we preserve the same reorder logic here so that the data-evolution raw path matches the normal read path?

crates/paimon/src/arrow/reader.rs
```diff
@@ -184,6 +188,259 @@ impl ArrowReader {
...
+    if data_files.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Determine which columns each file provides and resolve conflicts by max_sequence_number.
+    // column_name -> (file_index, max_sequence_number)
+    let mut column_source: HashMap<String, (usize, i64)> = HashMap::new();
+
+    for (file_idx, file_meta) in data_files.iter().enumerate() {
+        let file_columns: Vec<String> = if let Some(ref wc) = file_meta.write_cols {
```
QuakeWang, line 291: Falling back to “all columns from the current table schema” when `write_cols == None` is not correct here.

One of the main data-evolution scenarios is reading old files after the table has added new columns. In that case, an old file only contains fields from the schema identified by `file_meta.schema_id`; using the current table schema here will incorrectly mark later-added columns as if this file already provided them. That affects winning-column resolution, and it can also drop entire row groups when the projection only contains newly added columns.

crates/paimon/src/table/table_scan.rs
```diff
@@ -321,26 +370,52 @@ impl<'a> TableScan<'a> {
...
-                    .with_snapshot(snapshot_id)
-                    .with_partition(partition_row.clone())
-                    .with_bucket(bucket)
-                    .with_bucket_path(bucket_path.clone())
-                    .with_total_buckets(total_buckets)
-                    .with_data_files(file_group);
-                if let Some(files) = data_deletion_files {
-                    builder = builder.with_data_deletion_files(files);
+            if data_evolution_enabled {
+                let file_groups = split_by_row_id(data_files);
```
QuakeWang: The split-generation logic here diverges from upstream `DataEvolutionSplitGenerator`.

Upstream does not simply group by equal `first_row_id` and emit one split per group. It first merges overlapping `row_id_range`s, then applies ordered bin packing using `target_split_size`/`open_file_cost`, and computes `rawConvertible` from the packed result. The current implementation introduces two regressions:
  1. `source.split.*` is effectively bypassed in data-evolution mode, so splits become much more fragmented.
  2. Grouping only by `first_row_id` misses the overlapping row-id-range case, which no longer matches upstream grouping semantics.

  I think we should align this with the Java split generator before merging the new read path.

crates/paimon/src/arrow/reader.rs
JingsongLi: I cannot get your point. Please raise an example.

QuakeWang: > I cannot get your point. Please raise an example.

My point is about old files after schema evolution.

If an old file was written when the table schema was `(id, name)`, and later the table becomes `(id, name, age)`, then `write_cols == None` should not mean that this old file contains `age` too. But the current fallback uses the current table schema, so it will treat the old file as if it also provides the newly added columns.

I think for `write_cols == None`, we should use the file schema from `file_meta.schema_id`, not the current table schema.

JingsongLi: The old file can be treaded as `(id, name, age)`, the age is null. Actually, this is a schema evolution.

I don't want to introduce the reading of old schema files because it should address issues such as column type changes.

QuakeWang: > The old file can be treaded as `(id, name, age)`, the age is null. Actually, this is a schema evolution.
> 
> I don't want to introduce the reading of old schema files because it should address issues such as column type changes.

@JingsongLi  I get your point. I agree we do not necessarily need to introduce historical schema reading here, especially if that expands the scope to type evolution.

My concern is mainly that, even under the “current schema + NULL for missing columns” semantics, the current fallback may still behave incorrectly in some cases. For example, when projecting only a newly added column, I think this path may return 0 rows instead of preserving the row count and filling NULLs.

So my point is less about requiring old-schema reads, and more about whether the current fallback already implements the expected schema-evolution behavior correctly.

JingsongLi: https://github.com/apache/paimon-rust/pull/193/changes/81665a180b6fd0d803fb2b6ea89951c47f8f9311
You mean these three lines code? We should remove it.

crates/paimon/src/table/table_scan.rs
JingsongLi: Packaging different rowids together makes the implementation very complex, which requires re grouping during read, which is not necessary for Rust.

crates/paimon/src/arrow/reader.rs
QuakeWang: Yes, those three lines are part of the issue, so removing them makes sense.

I just want to make sure this is sufficient: for the add-column case, if the projected column does not physically exist in the old files, we should still preserve the row count and fill NULL, instead of dropping rows.

JingsongLi: I found it quite troublesome to fill in NULL, and we also need the corresponding Arrow type, which we don't have here, so the current implementation doesn't have a corresponding column. In the future, when we need to support schema evolution, we will see how to change it.

QuakeWang: > I found it quite troublesome to fill in NULL, and we also need the corresponding Arrow type, which we don't have here, so the current implementation doesn't have a corresponding column. In the future, when we need to support schema evolution, we will see how to change it.

OK, that makes sense to me.

PR #195, opened 2026-04-02
feat: introduce time travel for data fusion

JingsongLi:
### Purpose

Introduce time travel support for the DataFusion integration, allowing users to query historical snapshots via standard SQL syntax FOR SYSTEM_TIME AS OF.

### Tests

provision.py & read_tables.rs

### DIFF
--- a/crates/paimon/src/table/snapshot_manager.rs
+++ b/crates/paimon/src/table/snapshot_manager.rs
@@ -154,6 +154,15 @@ impl SnapshotManager {
                 message: format!("snapshot JSON invalid: {e}"),
                 source: Some(Box::new(e)),
             })?;
+        if snapshot.id() != snapshot_id {
+            return Err(crate::Error::DataInvalid {
+                message: format!(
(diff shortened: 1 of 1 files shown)


crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -99,15 +154,60 @@ impl SnapshotManager {
...
-        if snapshot.id() != snapshot_id {
-            return Err(crate::Error::DataInvalid {
-                message: format!(
-                    "snapshot file id mismatch: LATEST points to {snapshot_id}, but file contains snapshot id {}",
-                    snapshot.id()
-                ),
-                source: None
-            });
-        }
+        Ok(snapshot)
```
QuakeWang, line 166: This `get_snapshot()` path now returns the deserialized snapshot directly, but it no longer checks `snapshot.id() == snapshot_id`.

That means a mismatched `snapshot-{id}` file will be accepted silently, and both the new time-travel path and the normal latest-snapshot path can read the wrong snapshot instead of failing fast.

I think we should keep the same id-consistency check here before merging.

PR #196, opened 2026-04-03
feat(scan): support data-level stats pruning in TableScan

### LINKED ISSUE #188: feat(scan): support data-level stats pruning in TableScan
### Search before asking

### Description

This is part of #173. The current scan path already supports partition pruning, but after that, `TableScan` still plans all candidate data files even when file-level stats already show that a filter cannot match.

This task is to add safe data-level stats pruning in the core scan layer based on `DataFileMeta` stats, so the planner can skip impossible files earlier.

Initial scope:
- use file-level min/max values and null counts
- support primitive columns
- support `=`, `!=`, `>`, `>=`, `<`, `<=`, `IS NULL`, `IS NOT NULL`, and `AND`
- fail open when stats are missing, unsupported, corrupted, or inconclusive

### Willingness to contribute

QuakeWang:
### Purpose

Linked issue: close #188

### DIFF
--- a/crates/integration_tests/tests/read_tables.rs
+++ b/crates/integration_tests/tests/read_tables.rs
@@ -1006,6 +1006,58 @@ async fn test_read_data_evolution_table_with_projection() {
 }
 
 // ---------------------------------------------------------------------------
+// Limit pushdown integration tests
+// ---------------------------------------------------------------------------
+
+/// Helper function to scan and read with limit pushdown.
+async fn plan_table(table: &paimon::Table, limit: Option<usize>) -> Plan {
+    let mut read_builder = table.new_read_builder();
+    if let Some(limit) = limit {
+        read_builder.with_limit(limit);
+    }
+    let scan = read_builder.new_scan();
+    scan.plan().await.expect("Failed to plan scan")
+}
+
+/// Test limit pushdown: when limit is smaller than total rows, fewer data files may be generated.
+#[tokio::test]
+async fn test_limit_pushdown() {
+    let catalog = create_file_system_catalog();
+
+    // Test limit pushdown for data evolution table
+    let table = get_table_from_catalog(&catalog, "data_evolution_table").await;
+
+    // Get full plan without limit
+    let full_plan = plan_table(&table, None).await;
+    let full_data_split_count: usize = full_plan.splits().iter().count();
+
+    // Get the plan with limit = 2
+    let limited_plan = plan_table(&table, Some(2)).await;
+    let limited_data_split_count: usize = limited_plan.splits().iter().count();
+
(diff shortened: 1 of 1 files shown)


crates/paimon/src/table/table_scan.rs
```diff
@@ -166,6 +168,72 @@ fn partition_matches_predicate(
...
+    fn null_count(&self, index: usize) -> Option<i64> {
+        self.null_counts.get(index).copied()
+    }
+
+    /// Check whether the stats rows have the expected number of fields.
+    ///
+    /// If either min or max BinaryRow has an arity different from
+    /// `expected_fields`, the stats were likely written in dense mode or
+    /// under a different schema — making index-based access unsafe.
+    fn arity_matches(&self, expected_fields: usize) -> bool {
```
JingsongLi, line 256: We should consider schema evolution here, otherwise may occur exception here.

crates/paimon/src/table/table_scan.rs
```diff
@@ -166,6 +168,86 @@ fn partition_matches_predicate(
...
+        let max_ok = self
+            .max_values
+            .as_ref()
+            .is_none_or(|r| r.arity() as usize == expected_fields);
+        let null_ok = self.null_counts.is_empty() || self.null_counts.len() == expected_fields;
+        min_ok && max_ok && null_ok
+    }
+}
+
+fn split_partition_and_data_predicates(
```
JingsongLi, line 291: Can you align this implementation to Java `splitPartitionPredicatesAndDataPredicates`?  https://github.com/apache/paimon/pull/7473 is important to many use cases.

crates/paimon/src/table/table_scan.rs
JingsongLi: Maybe it is better to merge this first: https://github.com/apache/paimon-rust/pull/197

We can finish schema evolution here. Read schema file to check evolution here.

QuakeWang: > Maybe it is better to merge this first: #197
> 
> We can finish schema evolution here. Read schema file to check evolution here.

Good idea, I will review #197 soon~

crates/integration_tests/tests/read_tables.rs
```diff
@@ -948,58 +1005,6 @@ async fn test_read_data_evolution_table_with_projection() {
     );
 }
 
-// ---------------------------------------------------------------------------
-// Limit pushdown integration tests
```
JingsongLi, line 952: Why remove limit tests?

JingsongLi: I'll try to fix it.

PR #197, opened 2026-04-03
feat: support schema evolution read with SchemaManager

JingsongLi:
### Purpose

Sub task of #173

Introduce SchemaManager to load versioned schema files (schema/schema-{id}) and use field-ID-based index mapping (aligned with Java Paimon's SchemaEvolutionUtil) for reading data files written under older schemas. This supports add-column (null fill) and type promotion (cast) scenarios.

Also align paimon_type_to_arrow with Python's from_paimon_type:
- TIME always maps to Time32(Millisecond)
- TIMESTAMP precision 0 maps to Second
- Add nested type support: Array, Map, Multiset, Row


crates/paimon/src/arrow/reader.rs
```diff
@@ -242,108 +326,213 @@ fn read_single_file_stream(
...
+            .collect();
 
-    // Build owned data for the stream closure.
-    let file_io = file_io.clone();
-    let split = split.clone();
-    let data_files: Vec<DataFileMeta> = data_files.to_vec();
-    let projected_column_names = projected_column_names.to_vec();
-    let output_batch_size = batch_size.unwrap_or(1024);
+        // Collect which file indices we need to open streams for.
+        let active_file_indices: Vec<usize> = file_read_columns.keys().copied().collect();
```
QuakeWang, line 499: I’m a bit worried about an edge case here.

If the projection only asks for a newly added column that none of the files in this split contain yet, `file_read_columns` stays empty, `active_file_indices` becomes empty, and the merge loop exits immediately. In that case, we don’t return a NULL-filled column, we just skip the whole row group.

That feels risky. In a `data evolution + add column` case, something like `SELECT new_col FROM ...` would silently lose the rows coming from older data. I think we still need to preserve the row count here and fill the projected column with NULLs.

crates/paimon/src/arrow/mod.rs
```diff
@@ -16,5 +16,113 @@
...
+        PaimonDataType::SmallInt(_) => ArrowDataType::Int16,
+        PaimonDataType::Int(_) => ArrowDataType::Int32,
+        PaimonDataType::BigInt(_) => ArrowDataType::Int64,
+        PaimonDataType::Float(_) => ArrowDataType::Float32,
+        PaimonDataType::Double(_) => ArrowDataType::Float64,
+        PaimonDataType::VarChar(_) | PaimonDataType::Char(_) => ArrowDataType::Utf8,
+        PaimonDataType::Binary(_) | PaimonDataType::VarBinary(_) => ArrowDataType::Binary,
+        PaimonDataType::Date(_) => ArrowDataType::Date32,
+        PaimonDataType::Time(_) => ArrowDataType::Time32(TimeUnit::Millisecond),
+        PaimonDataType::Timestamp(t) => {
```
QuakeWang, line 42: I can see the idea behind changing `TIMESTAMP(0)` to `Timestamp(Second)`, but right now this only changes the reader side. The DataFusion provider-side schema mapping is still on the old behavior.

At execution time we consume the batches from `read.to_arrow()` directly, while the provider schema still reports `TIMESTAMP(0)` as `Timestamp(Millisecond)`. So as soon as a table contains `TIMESTAMP(0)`, the declared schema and the actual batch schema no longer match.

It would be better to update both sides together here,  this leaves a pretty subtle trap in the DataFusion integration.

crates/paimon/src/arrow/reader.rs
JingsongLi: I will fix it and add tests.

crates/paimon/src/arrow/mod.rs
JingsongLi: I will revert it to Millisecond.

crates/paimon/src/arrow/mod.rs
```diff
@@ -16,5 +16,112 @@
...
 
 pub use crate::arrow::reader::ArrowReaderBuilder;
+
+use crate::spec::DataType as PaimonDataType;
+use arrow_schema::DataType as ArrowDataType;
+use arrow_schema::{Field as ArrowField, TimeUnit};
+use std::sync::Arc;
+
+/// Converts a Paimon [`DataType`](PaimonDataType) to an Arrow [`DataType`](ArrowDataType).
+pub fn paimon_type_to_arrow(dt: &PaimonDataType) -> crate::Result<ArrowDataType> {
```
luoyuxia, line 29: Thanks for that. 
Then we can remove method `paimon_data_type_to_arrow` in `crates/integrations/datafusion/src/schema.rs`

crates/paimon/src/arrow/reader.rs
```diff
@@ -189,16 +215,74 @@ impl ArrowReader {
     }
 }
 
+/// Build the target Arrow schema from the read type (Paimon DataFields).
+fn build_target_arrow_schema(read_type: &[DataField]) -> crate::Result<Arc<ArrowSchema>> {
```
luoyuxia: nit: move to arrow mod, and expose it so that paimon-rust/crates/integrations/datafusion/src/schema.rs can reuse it?

JingsongLi: Good point!

crates/paimon/src/arrow/mod.rs
JingsongLi: Good point!

PR #198, opened 2026-04-03
chore(provision): make Spark integration test provisioning idempotent

freedom3219:
### Purpose

Make Spark-based integration test fixture provisioning idempotent.

`dev/spark/provision.py` previously used `CREATE TABLE IF NOT EXISTS` together with unconditional `INSERT INTO` statements. Re-running provisioning against the same warehouse caused append/log tables to accumulate duplicate rows, which made the test fixture non-deterministic and broke integration tests.

### Brief change log

- Resolve the configured Paimon warehouse path from Spark config
- Clear the existing warehouse contents before provisioning test fixtures
- Keep the existing table definitions and fixture data unchanged
- Preserve the current reader behavior and test expectations while making repeated provisioning deterministic

This approach fits the current repository architecture because the integration tests assert exact fixture contents, while append tables are read as appended data rather than deduplicated business rows. Resetting the whole warehouse before provisioning is the most reliable way to keep the fixture stable across append tables, DV PK tables, and data evolution tables.

### Tests

Verified idempotent provisioning by running fixture setup twice against the same warehouse, then running:

- `cargo test -p paimon-integration-tests --test read_tables`
- `cargo test -p paimon-datafusion --test read_tables`

### DIFF
--- a/dev/spark/provision.py
+++ b/dev/spark/provision.py
@@ -20,11 +20,49 @@
 # Provisions Paimon tables into the warehouse (file:/tmp/paimon-warehouse)
 # for paimon-rust integration tests to read.
 
+import shutil
+from pathlib import Path
+from urllib.parse import unquote, urlparse
+
 from pyspark.sql import SparkSession
+
+
+def _warehouse_path_from_spark_conf(spark: SparkSession) -> Path:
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #199, opened 2026-04-03
feat: support time travel by tag name and doc it

JingsongLi:
### Purpose

Read data from a named tag by passing a string that is not a timestamp:
`SELECT * FROM paimon.default.my_table FOR SYSTEM_TIME AS OF 'my_tag'`
Tags are named snapshots created via Paimon's tag management (e.g., CALL sys.create_tag(...) in Spark). This is useful for pinning a stable version of the data for reproducible queries.

### DIFF
--- a/crates/paimon/src/table/tag_manager.rs
+++ b/crates/paimon/src/table/tag_manager.rs
@@ -1,0 +1,89 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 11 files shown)


crates/paimon/src/table/tag_manager.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Get the snapshot for a tag, or None if the tag file does not exist.
+    ///
+    /// Tag files are JSON with the same schema as Snapshot.
+    pub async fn get(&self, tag_name: &str) -> crate::Result<Option<Snapshot>> {
+        let path = self.tag_path(tag_name);
+        let input = self.file_io.new_input(&path)?;
+        if !input.exists().await? {
+            return Ok(None);
+        }
+        let bytes = input.read().await?;
```
XiaoHongbo-Hope: Suggestion: can we use try catch to handle file not found exception here like java to reduce IO? 

Image

JingsongLi: Fixed

luoyuxia (approved):
+1

PR #202, opened 2026-04-03
feat: support complex types

JingsongLi:
### Purpose

We have already supported complex types in previous PRs, but we did not test them.

Sub task of #173

### DIFF
--- a/crates/integrations/datafusion/tests/read_tables.rs
+++ b/crates/integrations/datafusion/tests/read_tables.rs
@@ -586,13 +586,16 @@ async fn test_read_complex_type_table_via_datafusion() {
 
     assert_eq!(rows[0].0, 1);
     assert_eq!(rows[0].1, "[1, 2, 3]");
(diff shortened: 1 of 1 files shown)


crates/integrations/datafusion/tests/read_tables.rs
```diff
@@ -523,3 +523,76 @@ async fn test_data_evolution_drop_column_null_fill() {
...
+                arr_fmt.value(i).to_string(),
+                map_fmt.value(i).to_string(),
+                row_fmt.value(i).to_string(),
+            ));
+        }
+    }
+    rows.sort_by_key(|(id, _, _, _)| *id);
+
+    assert_eq!(rows[0].0, 1);
+    assert_eq!(rows[0].1, "[1, 2, 3]");
```
luoyuxia, line 588: do we need to assert rows[0].2 which is the map type:  map('a', 10, 'b', 20)

JingsongLi: Fixed

PR #203, opened 2026-04-04
feat: replace apache-avro with serde_avro_fast and parallelize manifest reads

JingsongLi:
### Purpose

Switch Avro deserialization from apache-avro (Value intermediate repr) to serde_avro_fast (direct bytes→struct), eliminating redundant allocations for ~10-20x deserialization speedup. Read manifest files concurrently with buffer_unordered(64) instead of sequentially.

Sub task of #173

### DIFF
--- a/crates/paimon/src/spec/manifest_entry.rs
+++ b/crates/paimon/src/spec/manifest_entry.rs
@@ -28,6 +28,9 @@ pub struct Identifier {
     pub bucket: i32,
     pub level: i32,
     pub file_name: String,
+    pub extra_files: Vec<String>,
+    pub embedded_index: Option<Vec<u8>>,
+    pub external_path: Option<String>,
 }
 
 /// Entry of a manifest file, representing an addition / deletion of a data file.
@@ -90,6 +93,9 @@ impl ManifestEntry {
             bucket: self.bucket,
             level: self.file.level,
             file_name: self.file.file_name.clone(),
+            extra_files: self.file.extra_files.clone(),
+            embedded_index: self.file.embedded_index.clone(),
+            external_path: self.file.external_path.clone(),
         }
     }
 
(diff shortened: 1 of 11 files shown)


crates/paimon/src/spec/manifest_entry.rs
```diff
@@ -28,6 +28,9 @@ pub struct Identifier {
     pub bucket: i32,
     pub level: i32,
     pub file_name: String,
+    pub extra_files: Vec<String>,
```
JingsongLi, line 31: Identifier must be aligned with Java, otherwise there will be exceptions.

crates/paimon/src/table/table_scan.rs
```diff
@@ -134,22 +140,19 @@ fn build_deletion_files_map(
 /// The identifier must be rich enough to match Paimon's file identity, otherwise a delete
 /// for one file version can incorrectly remove another with the same file name.
 fn merge_manifest_entries(entries: Vec<ManifestEntry>) -> Vec<ManifestEntry> {
-    let mut deleted_entry_keys = HashSet::new();
-    let mut added_entries = Vec::new();
-
+    let mut map: IndexMap<Identifier, ManifestEntry> = IndexMap::with_capacity(entries.len());
```
luoyuxia: I'm curious why this change is needed.

The previous implementation mostly follows py-paimon. With the current code, we may read two manifest entries like [Delete(id), Add(id)].
Because manifests are read concurrently, Delete(id) may be observed first. In that case, the later Add(id) would be kept by the current IndexMap + shift_remove logic.

However, IIUC, that is incorrect: once Delete(id) exists, the corresponding Add(id) should be filtered out rather than kept.

JingsongLi: There aren't many deleted entries, it doesn't seem very meaningful to modify it to read twice, as it will repeatedly read the manifest containing the deleted entry.

If we don't read it twice, the current implementation is quite reasonable, we only need to traverse it once.

JingsongLi: But I think the previous implementation is indeed safer. I will revert here.

PR #204, opened 2026-04-04
feat(python): introduce pypaimon core and DataFusion catalog integration

### LINKED ISSUE #189: introduce python binding to expose datafusion intergration
### Search before asking

### Description

as per-title, we already has datafusion, we can expose it to python so that pypaimon cli can use it.

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #189

Subtask of #173

Introduce the Rust-powered core for `pypaimon` and provide Python-side DataFusion integration for querying Paimon tables.

This change adds a new `bindings/python` package, exposes a `PaimonCatalog` that can be registered into a native DataFusion `SessionContext`, and updates the Python API/docs to focus on catalog-based integration for simple table queries.

It also includes a runtime fallback on the Rust DataFusion integration side to handle DataFusion FFI callbacks that may run without an entered Tokio runtime. This works around the current gap discussed in https://github.com/apache/datafusion/issues/16312.

### Brief change log

- add the new `bindings/python` package with `maturin`-based build configuration
- expose the Rust-powered core for `pypaimon`
- add Python DataFusion integration through `PaimonCatalog`
- align the Python API with native DataFusion catalog registration via `ctx.register_catalog_provider("paimon", catalog)`
- simplify the Python-side test to cover normal table queries only and drop time-travel coverage for now
- update Python README and package description to describe the Rust-powered core and DataFusion integration
- add Python typing support for DataFusion catalog export
- add `datafusion` as a Python dev dependency for local testing
- add a runtime fallback in `paimon-datafusion` for catalog and scan-planning FFI paths when DataFusion invokes callbacks without a Tokio runtime
- centralize the runtime fallback helpers and document that they are a temporary workaround until the upstream FFI runtime propagation issue is resolved

### Tests

- `uv run --no-sync maturin develop`
- `uv run --no-sync pytest tests/test_datafusion.py -q`

### API and Format

- adds a new Python package under `bindings/python`
- exposes `PaimonCatalog` for DataFusion integration in Python
- Python users register the catalog with native DataFusion `SessionContext`
- time travel is not exposed through the current Python DataFusion catalog integration
- no storage format change

### Documentation

- add `bindings/python/README.md`
- add `bindings/python/project-description.md`
- document that this project builds the Rust-powered core for `pypaimon`
- document DataFusion integration for querying Paimon tables from Python

### DIFF
--- a/dev/spark/README.md
+++ b/dev/spark/README.md
@@ -1,3 +1,22 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
 # Spark Provisioning for Integration Tests
 
 This directory contains the Spark + Paimon setup that provisions test tables into `/tmp/paimon-warehouse`.


bindings/python/project-description.md
```diff
@@ -0,0 +1,69 @@
...
+    "metastore": "rest",
+    "uri": "http://localhost:8080",
+    "warehouse": "my_warehouse",
+})
+
+ctx = SessionContext()
+ctx.register_catalog_provider("paimon", catalog)
+```
+
+Time travel queries are not supported in the Python binding at this time.
```
luoyuxia, line 69: **Why time travel is not supported in this PR**

  - This Python integration intentionally uses native DataFusion `SessionContext` plus `CatalogProvider`.
  - That path works well for normal table queries, but time travel requires planner-level extension rather than simple catalog/table
    registration.
  - DataFusion Python/FFI does not currently expose a suitable planner extension point for this, so we cannot cleanly wire time-travel semantics
    into native SessionContext here.

  **What would be needed to support time travel**

  - Either introduce a custom context that owns the planner extension,
  - or wait for / contribute planner-level registration support in DataFusion Python/FFI,
  - or expose time travel through a separate explicit API instead of native SQL/catalog resolution.

  **Why not use a custom context now**

  - The goal of this PR is to stay aligned with native DataFusion usage and keep the Python integration lightweight and predictable.
  - If we introduce a custom context and still want a user experience close to native SessionContext, we would need to re-expose a large part of
    the SessionContext API surface there, which is relatively heavy to implement and maintain.
  - It would also introduce another API model and make the overall Python experience less consistent with native DataFusion.
  - If time travel becomes a strong requirement, we can revisit it and design dedicated support separately.

.github/workflows/ci.yml
```diff
@@ -105,6 +105,13 @@ jobs:
           RUST_LOG: DEBUG
           RUST_BACKTRACE: full
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
JingsongLi: ASF cannot execute this uses.

luoyuxia: Thanks. I have updated.

bindings/python/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2021 Datafuse Labs
```
JingsongLi: This is wrong... Also modify project LICENSE file.

crates/integrations/datafusion/src/runtime.rs
```diff
@@ -0,0 +1,65 @@
...
+pub(crate) fn block_on_with_runtime<F>(
+    future: F,
+    runtime_error: &'static str,
+    panic_error: &'static str,
+) -> F::Output
+where
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    let run = move || block_on_new_runtime(future, runtime_error);
```
JingsongLi: Why not use global `OnceLock<Runtime>` to avoid creating everytime.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -82,19 +83,6 @@ impl TableProvider for PaimonTableProvider {
         TableType::Base
     }
 
-    fn supports_filters_pushdown(
```
JingsongLi, line 85: Why? Revert this?

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -125,7 +113,16 @@ impl TableProvider for PaimonTableProvider {
             read_builder.with_limit(limit);
         }
         let scan = read_builder.new_scan();
-        let plan = scan.plan().await.map_err(to_datafusion_error)?;
+        // DataFusion's Python FFI may poll `TableProvider::scan()` without an active
+        // Tokio runtime. `scan.plan()` can reach OpenDAL/Tokio filesystem calls while
+        // reading Paimon metadata, so we must provide a runtime here instead of
+        // assuming the caller already entered one.
+        let plan = await_with_runtime(
```
JingsongLi: `scan()` itself is an asynchronous fn, and the DataFusion FFI uses the runtime handle passed by FFI_CatalogProvider to poll it, so there is no need to do a runtime fallback here.

bindings/python/src/runtime.rs
```diff
@@ -0,0 +1,32 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use std::sync::OnceLock;
+
+use tokio::runtime::{Handle, Runtime};
+
+static RUNTIME: OnceLock<Runtime> = OnceLock::new();
+
+pub fn runtime() -> Handle {
```
JingsongLi: Can this just use paimon_datafusion::runtime::runtime?

crates/integrations/datafusion/src/table/mod.rs
luoyuxia: No behavior change here — I didn’t remove  supports_filters_pushdown(). I only moved it to match the method order in the TableProvider trait, to make IDE happy

bindings/python/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+This project builds the Rust-powered core for [PyPaimon](https://paimon.apache.org/docs/master/pypaimon/overview/) while also providing DataFusion integration for querying Paimon tables.
+
+## Usage
+
+For DataFusion queries, use the native `SessionContext` and register a `PaimonCatalog`:
+
+```python
+from datafusion import SessionContext
+from pypaimon_core.datafusion import PaimonCatalog
```
JingsongLi: maybe use pypaimon_rust.datafusion?

crates/integrations/datafusion/src/table/mod.rs
luoyuxia: I initially thought the same, but turns out not, Seems not all async FFI entry points propagate the runtime —
Without the await_with_runtime fallback, scan.plan() would panic at runtime for no active runtime on the thread.

bindings/python/README.md
luoyuxia: Good point. The current name pypaimon_core was inspired by pyiceberg-core, but given that pypaimon already exists as the pure-Python package, pypaimon_core could be misleading. pypaimon_rust might be more straightforward — it clearly indicates this is the Rust-backed implementation, similar to how cryptography has cryptography-rust. The mixed Python/Rust naming is a bit unusual but arguably more honest about what the package actually is.

JingsongLi (approved):
+1

PR #205, opened 2026-04-04
feat(scan): add bucket pruning, DV/postpone filtering, and DE group pruning

JingsongLi:
### Purpose

Subtask of #173

Major scan optimizations for TableScan:

- Push partition/data predicate filters into read_all_manifest_entries for early concurrent pruning during manifest reading
- Add manifest-file-level partition stats pruning before reading manifest files
- DV level-0 filtering now only applies to primary-key tables; non-PK tables with DV keep level-0 files
- Filter out postpone bucket (bucket < 0) entries for PK tables
- Data evolution group-level predicate filtering: after group_by_overlapping_row_id, merged stats across overlapping files allow pruning entire groups
- Bucket predicate filtering: extract bucket key predicates (Eq/In), compute target buckets via MurmurHash3 (seed=42, word-aligned), and skip manifest entries whose bucket is not in the target set.
- Disable limit pushdown when data predicates exist

### Tests

Tests: unit tests for bucket computation, predicate extraction, and target bucket calculation; integration tests for PK-without-DV, non-PK-with-DV, postpone bucket, data evolution filtering, and bucket predicate filtering.

### DIFF
--- a/crates/integrations/datafusion/src/physical_plan/scan.rs
+++ b/crates/integrations/datafusion/src/physical_plan/scan.rs
@@ -16,10 +16,8 @@
 // under the License.
 
 use std::any::Any;
-use std::pin::Pin;
 use std::sync::Arc;
 
-use datafusion::arrow::array::RecordBatch;
 use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
 use datafusion::error::Result as DFResult;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
@@ -27,7 +25,7 @@ use datafusion::physical_expr::EquivalenceProperties;
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::{DisplayAs, ExecutionPlan, Partitioning, PlanProperties};
-use futures::{Stream, StreamExt, TryStreamExt};
+use futures::{StreamExt, TryStreamExt};
 use paimon::table::Table;
 use paimon::DataSplit;
 
@@ -128,7 +126,6 @@ impl ExecutionPlan for PaimonTableScan {
         let table = self.table.clone();
         let schema = self.schema();
         let projected_columns = self.projected_columns.clone();
-        let limit = self.limit;
 
         let fut = async move {
             let mut read_builder = table.new_read_builder();


crates/paimon/src/spec/binary_row.rs
```diff
@@ -0,0 +1,752 @@
...
+    pub fn hash_code(&self) -> i32 {
+        hash_by_words(&self.data)
+    }
+
+    /// Build a BinaryRow from typed Datum values using `BinaryRowBuilder`.
+    pub fn from_datums(datums: &[(&crate::spec::Datum, &crate::spec::DataType)]) -> Option<Self> {
+        let arity = datums.len() as i32;
+        let mut builder = BinaryRowBuilder::new(arity);
+
+        for (pos, (datum, _data_type)) in datums.iter().enumerate() {
```
XiaoHongbo-Hope: `from_datums` currently ignores `_data_type`, but Java Paimon's `BinaryRow` encoding depends on the declared logical type and precision.

Ref: https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/data/BinaryRow.java#L204-L213

JingsongLi: Fixed

dev/spark/README.md
```diff
@@ -0,0 +1,41 @@
+# Spark Provisioning for Integration Tests
+
+This directory contains the Spark + Paimon setup that provisions test tables into `/tmp/paimon-warehouse`.
```
JingsongLi, line 3: Add this file to let AI know how to run Integration Tests with colima docker.

crates/paimon/src/spec/stats.rs
```diff
@@ -15,9 +15,32 @@
...
+/// Flattens nullable array with nullable items into `Vec<i64>`, treating null items as 0.
+fn deserialize_null_counts<'de, D>(deserializer: D) -> Result<Vec<i64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let opt: Option<Vec<Option<i64>>> = Option::deserialize(deserializer)?;
+    Ok(opt
+        .unwrap_or_default()
+        .into_iter()
+        .map(|v| v.unwrap_or(0))
```
XiaoHongbo-Hope: Seems this will make null item in null_counts as 0, which is different actually.  Null should mean "unknown",  and 0 means 'this is no null value'.

crates/paimon/src/table/table_scan.rs
```diff
@@ -691,8 +691,14 @@ impl<'a> TableScan<'a> {
             }
         }
 
-        // Apply limit pushdown to reduce the number of splits if possible
-        let splits = self.apply_limit_pushdown(splits);
+        // Apply limit pushdown only when there are no data predicates.
+        // With data predicates, merged_row_count() reflects pre-filter row counts,
+        // so stopping early could return fewer rows than the limit after filtering.
+        let splits = if data_predicates.is_empty() {
```
luoyuxia, line 691: good catch on it.  I'm wondering whether java&python has same issue?

JingsongLi: In fact, Java engines do not push limits when there is a filter. When pushing filters, Java engines require connectors to return which filters can be fully effective before pushing limits.

I checked datafusion and it doesn't have this complex filter push logic.

crates/paimon/src/table/table_scan.rs
```diff
@@ -63,22 +68,145 @@ async fn read_manifest_list(
...
+                                if !targets.contains(&entry.bucket()) {
+                                    return false;
+                                }
+                            }
+                        }
+                        if let Some(pred) = partition_predicate {
+                            match partition_matches_predicate(entry.partition(), pred) {
+                                Ok(false) => return false,
+                                Ok(true) => {}
+                                Err(_) => {}
```
luoyuxia: Is it expected to ignore the error here? If any error occurs at this point (for example, a temporary I/O error), could it cause an entry that actually does not match the predicate to be kept and returned?

crates/paimon/src/table/table_scan.rs
luoyuxia: Thanks. Then, I think we should also remove this limit logic in `crates/integrations/datafusion/src/physical_plan/scan.rs` 
```
// Enforce the final LIMIT at the DataFusion execution layer.
        let limited_stream: Pin<Box<dyn Stream<Item = DFResult<RecordBatch>> + Send>> =
            if let Some(limit) = limit {
                let mut remaining = limit;
                Box::pin(stream.try_filter_map(move |batch| {
                    futures::future::ready(if remaining == 0 {
                        Ok(None)
                    } else if batch.num_rows() <= remaining {
                        remaining -= batch.num_rows();
                        Ok(Some(batch))
                    } else {
                        let limited_batch = batch.slice(0, remaining);
                        remaining = 0;
                        Ok(Some(limited_batch))
                    })
                }))
            } else {
                Box::pin(stream)
            };

        Ok(Box::pin(RecordBatchStreamAdapter::new(
            self.schema(),
            limited_stream,
        )))
```

luoyuxia (approved):
+1

PR #224, opened 2026-04-07
feat(datafusion): support fetch contract in PaimonTableScan

QuakeWang:
### Purpose

Linked issue: close #220 

Align DataFusion scan pruning with the core-owned limit-hint design.

Instead of introducing a DataFusion-only fetch contract, this PR now reuses core-owned split pruning by passing a conservative limit hint into `read_builder.with_limit(...)` when it is safe to do so.

### Brief change log

  - keep the shared split-pruning rule in `table_scan`
  - reuse that core-owned pruning from DataFusion via `read_builder.with_limit(...)`
  - only push a limit hint when there are no filters or all pushed filters are `Exact`
  - fail open for inexact or residual filters instead of adding a scan-side exact fetch contract
  - simplify `PaimonTableScan` back to regular scan execution without scan-side fetch handling
  - update plan and integration tests to cover safe limit hints and the absence of scan-side `fetch=...`

### DIFF
--- a/crates/integrations/datafusion/src/filter_pushdown.rs
+++ b/crates/integrations/datafusion/src/filter_pushdown.rs
@@ -20,15 +20,72 @@ use datafusion::logical_expr::expr::InList;
 use datafusion::logical_expr::{Between, BinaryExpr, Expr, Operator, TableProviderFilterPushDown};
 use paimon::spec::{DataField, DataType, Datum, Predicate, PredicateBuilder};
 
-pub(crate) fn classify_filter_pushdown(
+#[derive(Debug)]
+struct SingleFilterAnalysis {
+    translated_predicates: Vec<Predicate>,
+    has_untranslated_residual: bool,
+}
+
+#[derive(Debug)]
+pub(crate) struct FilterPushdownAnalysis {
+    pub(crate) pushed_predicate: Option<Predicate>,
+    pub(crate) has_untranslated_residual: bool,
+}
+
+fn analyze_filter(filter: &Expr, fields: &[DataField]) -> SingleFilterAnalysis {
+    let translator = FilterTranslator::new(fields);
+    if let Some(predicate) = translator.translate(filter) {
+        return SingleFilterAnalysis {
+            translated_predicates: vec![predicate],
+            has_untranslated_residual: false,
+        };
+    }
+
+    SingleFilterAnalysis {
+        translated_predicates: split_conjunction(filter)
+            .into_iter()
+            .filter_map(|expr| translator.translate(expr))
+            .collect(),
+        has_untranslated_residual: true,
+    }
+}
+
+pub(crate) fn analyze_filters(filters: &[Expr], fields: &[DataField]) -> FilterPushdownAnalysis {
+    let mut translated_predicates = Vec::new();
+    let mut has_untranslated_residual = false;
+
+    for filter in filters {
+        let analysis = analyze_filter(filter, fields);
+        translated_predicates.extend(analysis.translated_predicates);
+        has_untranslated_residual |= analysis.has_untranslated_residual;
+    }
+
+    FilterPushdownAnalysis {
+        pushed_predicate: if translated_predicates.is_empty() {
+            None
+        } else {
+            Some(Predicate::and(translated_predicates))
+        },
+        has_untranslated_residual,
+    }
+}
+
+#[cfg(test)]
+pub(crate) fn build_pushed_predicate(filters: &[Expr], fields: &[DataField]) -> Option<Predicate> {
+    analyze_filters(filters, fields).pushed_predicate
+}
+
+pub(crate) fn classify_filter_pushdown<F>(
     filter: &Expr,
     fields: &[DataField],
-    partition_keys: &[String],
-) -> TableProviderFilterPushDown {
+    is_exact_filter_pushdown: F,
+) -> TableProviderFilterPushDown
+where
+    F: Fn(&Predicate) -> bool,
+{
     let translator = FilterTranslator::new(fields);
-    if translator.translate(filter).is_some() {
-        let partition_translator = FilterTranslator::for_allowed_columns(fields, partition_keys);
-        if partition_translator.translate(filter).is_some() {
+    if let Some(predicate) = translator.translate(filter) {
+        if is_exact_filter_pushdown(&predicate) {
             TableProviderFilterPushDown::Exact
         } else {
             TableProviderFilterPushDown::Inexact
@@ -40,21 +97,6 @@ pub(crate) fn classify_filter_pushdown(
         TableProviderFilterPushDown::Inexact
     } else {
         TableProviderFilterPushDown::Unsupported
-    }
-}
-
-pub(crate) fn build_pushed_predicate(filters: &[Expr], fields: &[DataField]) -> Option<Predicate> {
-    let translator = FilterTranslator::new(fields);
-    let pushed: Vec<_> = filters
-        .iter()
-        .flat_map(split_conjunction)
-        .filter_map(|filter| translator.translate(filter))
-        .collect();
-
-    if pushed.is_empty() {
-        None
-    } else {
-        Some(Predicate::and(pushed))
     }
 }
 
@@ -75,7 +117,6 @@ fn split_conjunction(expr: &Expr) -> Vec<&Expr> {
 
 struct FilterTranslator<'a> {
     fields: &'a [DataField],
-    allowed_columns: Option<&'a [String]>,
     predicate_builder: PredicateBuilder,
 }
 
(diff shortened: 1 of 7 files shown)


crates/paimon/src/table/table_scan.rs
```diff
@@ -380,51 +419,8 @@ impl<'a> TableScan<'a> {
...
-    /// limit hint is reached. Returns only the splits likely needed to satisfy
-    /// that hint.
-    ///
-    /// This does not guarantee an exact final row count. If a split's
-    /// `merged_row_count()` is `None` (for example because of unknown deletion
-    /// cardinality), that split is kept even though its contribution to the
-    /// limit is unknown. Planning may still stop early later if the
-    /// accumulated known `merged_row_count()` reaches the limit, and the
-    /// caller or query engine must enforce the final LIMIT.
     fn apply_limit_pushdown(&self, splits: Vec<DataSplit>) -> Vec<DataSplit> {
```
JingsongLi, line 442: prune_splits_by_limit_hint is same to apply_limit_pushdown.

We can just make `apply_limit_pushdown` public?

crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -16,75 +16,307 @@
...
+        return PartitionPlan {
+            planned_partitions: split_buckets.into_iter().map(Arc::from).collect(),
+            partition_fetches: row_buckets
+                .into_iter()
+                .map(|rows| Some(rows.into_iter().sum()))
+                .collect(),
+        };
+    }
+
+    let tail_start = tail_start.expect("tail_start checked above");
```
JingsongLi: Just unwrap()

crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -16,75 +16,307 @@
...
+}
+
+pub(crate) fn build_partition_plan(
+    full_splits: &[DataSplit],
+    target_partitions: usize,
+    limit_hint: Option<usize>,
+    fetch: Option<usize>,
+) -> PartitionPlan {
+    if let Some(fetch) = fetch {
+        let selected_splits = prune_splits_by_limit_hint(full_splits.iter().cloned(), Some(fetch));
```
JingsongLi: Why not use read_builder.with_limit here?

crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -124,6 +357,35 @@ impl ExecutionPlan for PaimonTableScan {
         Ok(self)
     }
 
+    fn with_fetch(&self, fetch: Option<usize>) -> Option<Arc<dyn ExecutionPlan>> {
+        if self.scan_fetch_mode != ScanFetchMode::Exact {
+            return None;
+        }
+
+        let fetch = fetch.filter(|fetch| *fetch > 0);
```
JingsongLi: Why filter fetch > 0 here?

crates/integrations/datafusion/src/physical_plan/scan.rs
QuakeWang: I did not switch this to `read_builder.with_limit(...)` here because the provider stage needs to retain the full split plan for later physical `with_fetch()` rebuilds. For `OFFSET + LIMIT`, DataFusion may push `skip + fetch`, so narrowing the provider scan too early would lose the extra skip budget. I added a short comment here to make that intent explicit.

crates/integrations/datafusion/src/physical_plan/scan.rs
QuakeWang: Good catch. I removed the `fetch > 0` normalization so `Some(0)` is preserved as part of the physical fetch contract, and I added tests to cover the zero-fetch path.

crates/integrations/datafusion/src/physical_plan/scan.rs
JingsongLi: Maybe can we support fetch in the table_scan?

QuakeWang: I think this is a reasonable direction longer-term, and that is also why I moved the shared split-pruning rule into `table_scan` as `prune_splits_by_limit_hint(...)`.

For this PR, I kept the full physical fetch contract in the DataFusion layer because it also needs DataFusion-specific behavior (`with_fetch()`, repartitioned scan planning, per-partition fetch budgets, and exact stream truncation). Moving all of that into `table_scan` would expand the scope beyond issue #220.

If we later want a cleaner core abstraction for fetch-aware split planning, I’d be happy to follow up on that separately.

JingsongLi: I still think it should be placed on `table_scan`. If you think it's a meaningful feature, otherwise it's just a standalone ability for datafusion, then its significance is not significant.

BTW, currently, there is no engine that supports fetch optimization.

QuakeWang: I aligned the latest commit with this direction. The PR no longer introduces a DataFusion-only fetch contract: `TableProvider::scan()` now only passes a conservative limit hint via `read_builder.with_limit(...)` when there are no filters or all pushed filters are `Exact`, and `PaimonTableScan` no longer carries `with_fetch()` / per-partition fetch budgets / exact stream truncation.

I also updated the plan tests to assert scan-side `limit=...` without scan-side `fetch=...` for `OFFSET + LIMIT`. So the remaining change is just core-owned limit-hint pruning reused by DataFusion, not a standalone fetch optimization feature. If we want true fetch optimization later, I agree it should be introduced from `table_scan` first.

crates/paimon/src/table/table_scan.rs
```diff
@@ -310,6 +310,48 @@ pub struct TableScan<'a> {
...
+/// limit hint is reached. Returns only the splits likely needed to satisfy
+/// that hint.
+///
+/// This does not guarantee an exact final row count. If a split's
+/// `merged_row_count()` is `None` (for example because of unknown deletion
+/// cardinality), that split is kept even though its contribution to the
+/// limit is unknown. Planning may still stop early later if the
+/// accumulated known `merged_row_count()` reaches the limit, and the
+/// caller or query engine must enforce the final LIMIT.
+pub fn prune_splits_by_limit_hint(
```
JingsongLi: Do we need to modify this? Just let it as it is?

QuakeWang: Good point.

I kept this logic internal in `TableScan` and removed the extra public helper. I also added the regression coverage back for `limit == 0` and unknown `merged_row_count()` so the original limit-hint behavior stays unchanged.

crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -58,6 +58,22 @@ pub(crate) fn build_pushed_predicate(filters: &[Expr], fields: &[DataField]) ->
     }
 }
 
+/// Whether it is safe to pass a row-count hint down to paimon-core planning.
+///
+/// This stays intentionally narrow: the hint is only safe when there are no
+/// filters, or when every filter is exact at the table-provider boundary.
+pub(crate) fn can_push_down_limit_hint(
```
JingsongLi: I think table_scan has already made this judgment.

crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -58,6 +58,22 @@ pub(crate) fn build_pushed_predicate(filters: &[Expr], fields: &[DataField]) ->
...
+    filters: &[Expr],
+    fields: &[DataField],
+    partition_keys: &[String],
+) -> bool {
+    filters.is_empty()
+        || filters.iter().all(|filter| {
+            classify_filter_pushdown(filter, fields, partition_keys)
```
JingsongLi: Another point is, can we add this judgment to read_builder.with_filter? We can clean datafusion connector here.

QuakeWang: I moved the scan-owned part of this judgment into paimon-core. The remaining DataFusion-side check is only for residual filters that are not visible to `ReadBuilder::with_filter(...)`, because `with_filter` only receives the translated Paimon `Predicate`.

Unsupported or inexact DataFusion filters may still remain above the scan, and that residual part is still needed to decide whether `LIMIT` can be pushed safely. So I kept only this residual-filter guard in the DataFusion connector.

JingsongLi: We can just provide a method in read_builder.rs, `exacted_filter_pushdown` return bool.
You can just use this method to datafusion `supports_filters_pushdown`. For limit, you can just push it always.

And then what else do we need to do?

JingsongLi (approved):
+1

Issue #232, opened 2026-04-10
Introduce WriteBuilder


JingsongLi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Motivation

Introduce WriteBuilder to Paimon Table:
- [ ] Table Commit.
- [ ] Table Write.
- [ ] Table Update.
- [ ] DataFusion Integration.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #233, opened 2026-04-10
feat(table): Add commit pipeline with SnapshotCommit abstraction

JingsongLi:
### Purpose

Subtask of #232

Implement the table write and commit infrastructure, including:
- SnapshotCommit trait with RenamingSnapshotCommit (filesystem) and RESTSnapshotCommit (REST catalog) implementations
- TableCommit with retry logic, append/overwrite/truncate support, partition statistics generation, and row tracking
- WriteBuilder as the entry point for creating TableCommit instances, with overwrite mode configured at construction time
- RESTEnv to hold REST catalog context (API client, identifier, uuid)
- CommitMessage, PartitionStatistics, and ManifestList types
- SnapshotManager extensions for atomic snapshot commit and latest hint
- BinaryRow write_datum and datums_to_binary_row utilities
- CoreOptions accessors for bucket, commit retry, and row tracking

Reference: pypaimon write/commit pipeline

### DIFF
--- a/crates/integrations/datafusion/src/physical_plan/scan.rs
+++ b/crates/integrations/datafusion/src/physical_plan/scan.rs
@@ -248,6 +248,7 @@ mod tests {
     use paimon::spec::{
         BinaryRow, DataType, Datum, IntType, PredicateBuilder, Schema as PaimonSchema, TableSchema,
     };
+    use paimon::table::Table;
     use std::fs;
     use tempfile::tempdir;
     use test_utils::{local_file_path, test_data_file, write_int_parquet_file};
@@ -298,6 +299,7 @@ mod tests {
             Identifier::new("test_db", "test_table"),
             "/tmp/test-table".to_string(),
             table_schema,
+            None,
         )
     }
 
@@ -329,6 +331,7 @@ mod tests {
             Identifier::new("default", "t"),
             table_path,
             table_schema,
+            None,
         );
 
         let split = paimon::DataSplitBuilder::new()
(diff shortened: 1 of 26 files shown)


jerry-024 (commented):
## `commit_snapshot` rename path silently overwrites on local filesystem

opendal's `rename` uses POSIX semantics — it silently overwrites the target if it already exists. Java Paimon's `FileIO.rename()` returns `false` when the target exists, and additionally wraps the operation in `lock.runWithLock(() -> !fileIO.exists(newSnapshotPath) && callable.call())`.

paimon-rust does neither check. On local filesystem with concurrent writers:

1. Writer A writes temp-A, renames to `snapshot-1` → succeeds
2. Writer B writes temp-B, renames to `snapshot-1` → also succeeds (POSIX overwrite)
3. Both return `Ok(true)`, `try_commit` stops retrying for both
4. Writer A's snapshot content is gone, its manifest lists are orphaned — silent data loss

**Suggested fix**: Add an `exists()` check before rename, matching Java's pattern:

```rust
// Check before rename to avoid silent overwrite (opendal uses POSIX rename semantics)
if self.file_io.exists(&target_path).await? {
    let _ = self.file_io.delete_file(&tmp_path).await;
    return Ok(false);
}
match self.file_io.rename(&tmp_path, &target_path).await {
    ...
}
```

This narrows the race window significantly. For full mutual exclusion on backends without atomic rename, an external lock mechanism (like Java's `Lock` interface) would be needed.

jerry-024 (approved):
+1

PR #237, opened 2026-04-11
feat(datafusion): Add DDL support with PRIMARY KEY constraint syntax

JingsongLi:
### Purpose

- Add PaimonDdlHandler for CREATE TABLE, ALTER TABLE (ADD/DROP/RENAME COLUMN, RENAME TABLE)
- Add PaimonTableFactory for CREATE EXTERNAL TABLE via DataFusion TableProviderFactory
- Extend PaimonCatalogProvider/SchemaProvider with CREATE/DROP SCHEMA and DROP TABLE
- Add arrow_to_paimon_type and arrow_fields_to_paimon to paimon::arrow for Arrow-to-Paimon type conversion
- Support PRIMARY KEY (col, ...) constraint syntax in CREATE TABLE DDL


crates/integrations/datafusion/src/ddl.rs
```diff
@@ -0,0 +1,387 @@
...
+        if statements.len() != 1 {
+            return Err(DataFusionError::Plan(
+                "Expected exactly one SQL statement".to_string(),
+            ));
+        }
+
+        match &statements[0] {
+            Statement::CreateTable(create_table) => self.handle_create_table(create_table).await,
+            Statement::AlterTable {
+                name, operations, ..
```
littlecoder04: Statement::AlterTable has an if_exists flag, but it is ignored here.

JingsongLi: Good catch.

crates/integrations/datafusion/src/table_factory.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) -> DFResult<Arc<dyn TableProvider>> {
+        if !cmd.location.is_empty() {
+            return Err(DataFusionError::Plan(
+                "LOCATION is not supported for Paimon tables. Table path is determined by the catalog warehouse.".to_string(),
+            ));
+        }
+
+        let identifier = resolve_identifier(&cmd.name)?;
+
+        // Build Paimon schema from the CREATE EXTERNAL TABLE command.
```
littlecoder04: My understanding is that both CREATE TABLE and CREATE EXTERNAL TABLE create managed Paimon tables here, rather than true external tables. It looks like two SQL entry points for creating the managed Paimon tables. Is that right?

JingsongLi: Good point, let' ban EXTERNAL.

littlecoder04:
+1

littlecoder04 (approved):

leaves12138 (approved):
Nice work. The DDL support is well-structured and the mock-based tests in `ddl.rs` are thorough.

**Highlights:**

1. **`PaimonDdlHandler`** cleanly intercepts CREATE TABLE and ALTER TABLE while delegating everything else to the inner `SessionContext`. The three-part name resolution (`catalog.db.table` → strip catalog prefix) is correct.

2. **Arrow↔Paimon type conversion** in `crates/paimon/src/arrow/mod.rs` covers all the common types with roundtrip tests. The mapping of `Utf8/LargeUtf8/Utf8View → VarChar` and `Binary/LargeBinary/BinaryView → VarBinary` is sensible.

3. **Integration tests** in `ddl_tests.rs` using `FileSystemCatalog` + `TempDir` provide good end-to-end coverage for CREATE TABLE with PK, partitions, and WITH options.

- `sql_data_type_to_arrow` doesn't handle ARRAY, MAP, or ROW types — those will hit the `_ => Err(...)` fallback. Users would need to use programmatic API for complex types in DDL. Worth a doc mention.
- The `register_table` impl in `catalog.rs` just acknowledges (`Ok(Some(table))`). This is fine since the actual table creation happens via `PaimonDdlHandler`, but a short comment explaining the lifecycle (DDL handler creates → DataFusion calls register_table → we ack) would help future readers.

+1, good to merge.

PR #238, opened 2026-04-12
feat(table): Add list_all metadata APIs and Tag struct

plusplusjiajia:
Add list_all-style accessors to the snapshot, schema, and tag managers and introduce a Tag type that mirrors Java's tag/Tag.java. These are the foundation for exposing Paimon system tables ($snapshots, $schemas, $tags, $options) through downstream query engines such as the DataFusion integration.

### DIFF
--- a/crates/paimon/src/error.rs
+++ b/crates/paimon/src/error.rs
@@ -109,6 +109,16 @@ pub enum Error {
     },
 }
 
+impl Error {
+    /// Whether this error wraps an `opendal::ErrorKind::NotFound`.
+    pub fn is_not_found(&self) -> bool {
+        matches!(
+            self,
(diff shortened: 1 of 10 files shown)


JingsongLi:
Thanks @plusplusjiajia for the contribution.

Let's implement it in a different style. System tables are relatively simple implementation paths. We can give each feature an independent PR, so that we can better know which APIs are necessary and which are unnecessary.

plusplusjiajia:
> Let's implement it in a different style. System tables are relatively simple implementation paths. We can give each feature an independent PR, so that we can better know which APIs are necessary and which are unnecessary.

Thanks @JingsongLi , makes sense. I'll close this PR and resubmit as small feature-driven PRs, each driven from the DataFusion side so every PR ends with a runnable SELECT * FROM t$xxx.  Order: $options → $schemas → $snapshots → $tags.

PR #240, opened 2026-04-12
feat(datafusion): Add $options system table

plusplusjiajia:
Adds support for the `<table>$options` Paimon system table to the DataFusion integration, mirroring Java `OptionsTable`.

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -145,43 +145,6 @@ impl PaimonSchemaProvider {
     pub fn new(catalog: Arc<dyn Catalog>, database: String) -> Self {
         PaimonSchemaProvider { catalog, database }
     }
-
-    /// Resolves `<base>$<system_name>` into a system table provider.
-    ///
-    /// Unknown system names return `Ok(None)` (DataFusion reports "table not
-    /// found"). When the system name is registered but the base table is
-    /// missing, an explicit error is returned so users can tell the two cases
-    /// apart in error messages.
-    async fn load_system_table(
-        &self,
-        base: &str,
-        system_name: &str,
-    ) -> DFResult<Option<Arc<dyn TableProvider>>> {
-        if !system_tables::is_registered(system_name) {
-            return Ok(None);
-        }
-
-        let catalog = Arc::clone(&self.catalog);
-        let database = self.database.clone();
-        let base_owned = base.to_string();
-        let system_name_owned = system_name.to_string();
-        await_with_runtime(async move {
-            let identifier = Identifier::new(database, base_owned.clone());
-            match catalog.get_table(&identifier).await {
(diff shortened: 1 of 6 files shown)


crates/integrations/datafusion/src/system_tables/options.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        _filters: &[Expr],
+        _limit: Option<usize>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        // Re-read latest schema each scan so in-place schema evolution is visible (Java parity).
```
JingsongLi: In-place? Rust has no cache, so you can just get schema from Table.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -112,6 +145,43 @@ impl PaimonSchemaProvider {
...
         PaimonSchemaProvider { catalog, database }
     }
+
+    /// Resolves `<base>$<system_name>` into a system table provider.
+    ///
+    /// Unknown system names return `Ok(None)` (DataFusion reports "table not
+    /// found"). When the system name is registered but the base table is
+    /// missing, an explicit error is returned so users can tell the two cases
+    /// apart in error messages.
+    async fn load_system_table(
```
JingsongLi: Put these methods to system_tables/mod.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -130,8 +200,13 @@ impl SchemaProvider for PaimonSchemaProvider {
     }
 
     async fn table(&self, name: &str) -> DFResult<Option<Arc<dyn TableProvider>>> {
+        let (base, system_name) = split_object_name(name);
+        if let Some(system_name) = system_name {
+            return self.load_system_table(base, system_name).await;
```
JingsongLi: You can create a wrap_to_system_table method to return table.

crates/integrations/datafusion/src/system_tables/options.rs
plusplusjiajia: > In-place? Rust has no cache, so you can just get schema from Table.

Thanks for catching this!  Now scan() reads self.table.schema().options() directly.

crates/integrations/datafusion/src/catalog.rs
plusplusjiajia: > Put these methods to system_tables/mod.

Thanks for the suggestion!  Moved.

crates/integrations/datafusion/src/catalog.rs
plusplusjiajia: > You can create a wrap_to_system_table method to return table.

Good point!

JingsongLi (approved):
+1

PR #246, opened 2026-04-15
refactor(read): split ArrowReader into DataFileReader, DataEvolutionReader and TableRead

JingsongLi:
### Purpose

Move the monolithic arrow/reader.rs into three focused modules under table:
- data_file_reader.rs: regular Parquet file reading with schema evolution
- data_evolution_reader.rs: column-merge reading for data evolution mode
- table_read.rs: TableRead extracted from read_builder.rs

Also replaces assert! with proper error returns and adds bounds checking in attach_row_id.


jerry-024 (approved):
+1

PR #249, opened 2026-04-15
feat: add primary-key table read/write support with sort-merge deduplication

JingsongLi:
### Purpose

Add KV file reader/writer, sort-merge reader with LoserTree, and RowKind support to enable reading and writing primary-key tables. Includes integration tests and DataFusion pk_tables tests.


crates/paimon/src/table/table_read.rs
```diff
@@ -76,17 +77,37 @@ impl<'a> TableRead<'a> {
...
+                                .sequence_fields()
+                                .iter()
+                                .map(|s| s.to_string())
+                                .collect(),
+                        },
+                    );
+                    return reader.read(data_splits);
+                }
+                crate::spec::MergeEngine::FirstRow => {
+                    // Fall through to DataFileReader — scan already skips level-0
```
jerry-024, line 104: `MergeEngine::FirstRow` still falls back to `DataFileReader` here, so the new support only works because `TableScan` hides level-0 files. That means fresh first-row writes remain invisible until compaction, which makes the latest snapshot incomplete rather than actually implementing first-row read semantics. If we want to stay behavior-compatible with the Java API, I think this constraint should be made explicit and tested. In Java, the analogous behavior is scoped to `DataTableBatchScan` as a batch-scan policy; it is not a general first-row merge reader implementation.

crates/paimon/src/table/table_scan.rs
```diff
@@ -461,6 +461,21 @@ impl<'a> TableScan<'a> {
         let deletion_vectors_enabled = core_options.deletion_vectors_enabled();
         let data_evolution_enabled = core_options.data_evolution_enabled();
 
+        let has_primary_keys = !self.table.schema().primary_keys().is_empty();
+
+        // Skip level-0 files for PK tables when:
+        // - DV mode: level-0 files are unmerged, DV handles dedup at higher levels
+        // - FirstRow engine without DV: reads go through DataFileReader (no merge),
+        //   so only compacted (level > 0) files are safe to read directly
+        let skip_level_zero = if has_primary_keys {
```
jerry-024: This looks broader than the Java implementation. In Java, level-0 skipping is kept inside `DataTableBatchScan`, while non-read paths still use the full file set: overwrite scanning goes through `CommitScanner.readOverwriteChanges`, and writer restore computes max sequence from all restored files in `AbstractFileStoreWrite.createWriterContainer`. Here `skip_level_zero` is pushed into the shared manifest-planning path, so overwrite / truncate and `TableWrite::scan_partition_sequence_numbers` also stop seeing live level-0 files. That changes behavior relative to the Java API and can lead to missed deletes or sequence-number bootstrap from an incomplete state for first-row / DV tables.

crates/paimon/src/table/table_read.rs
JingsongLi: This is Paimon's protocol.

crates/paimon/src/table/table_scan.rs
JingsongLi: Good point!

jerry-024 (approved):
+1

PR #250, opened 2026-04-15
feat(blob): add BlobType groundwork and preserve DDL semantics

QuakeWang:
### Purpose

This PR adds the first groundwork slice for BlobType in `paimon-rust`.

It introduces `BlobType` into the Rust type system, preserves `BLOB` in DataFusion DDL, and explicitly rejects unsupported write paths.

This PR is only a groundwork step for #228, and I will implement `.blob` read path for the next pr.

### Brief change log

  - add `DataType::Blob` and `BlobType`
  - support BlobType schema serialization / deserialization and add serde fixtures
  - expose BlobType as Arrow `Binary` in the existing type conversion surface
  - preserve `BLOB` in DataFusion DDL schema translation, instead of silently degrading it to `VARBINARY`
  - add tests for `CREATE TABLE` / `ALTER TABLE ... ADD COLUMN ... BLOB`
  - explicitly reject BlobType in current write paths
  - explicitly keep BlobType unsupported for partition keys and parquet row filter literals

### API and Format

API change:
  - add a new logical type `BlobType` / `DataType::Blob`

Format impact:
  - schema serde now supports `BLOB`
  - no data file format change in this PR
  - `.blob` file read support is not included in this PR


crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -324,20 +319,87 @@ impl PaimonSqlHandler {
...
+            datafusion::sql::sqlparser::ast::ColumnOption::NotNull
+        )
+    })
+}
+
+/// Convert a sqlparser SQL data type to a Paimon data type.
+///
+/// DDL schema translation must use this function instead of going through Arrow,
+/// because Arrow cannot preserve logical distinctions such as `BLOB` vs `VARBINARY`.
+fn sql_data_type_to_paimon_type(
```
JingsongLi, line 347: If we have this, we should remove `sql_data_type_to_arrow`.

QuakeWang: @JingsongLi Thanks for the review. `sql_data_type_to_arrow` has been removed, and DDL type translation now goes directly through `sql_data_type_to_paimon_type`.

JingsongLi (approved):
+1

PR #251, opened 2026-04-16
feat(blob): add create-time schema contract checks

QuakeWang:
### Purpose

Linked issue: part of #228

This PR hardens the create-time schema contract for `BLOB` columns.

After `BlobType` groundwork landed, Rust still relied on scattered runtime rejections for some invalid table shapes. This change moves the basic table-creation checks to schema construction so invalid  `BLOB` table definitions fail earlier and more consistently.

This PR is intentionally limited to create-time validation. It does not add `.blob` file-format dispatch or read path support.

### Brief change log

  - add create-time `BLOB` validation in `Schema::new()`
  - require `data-evolution.enabled = true` when top-level `BLOB` columns exist
  - reject tables whose top-level columns are all `BLOB`
  - reject top-level `BLOB` columns in partition keys
  - add schema-level tests for the accepted and rejected cases
  - update the existing `table_write` test fixture to satisfy the new schema contract


JingsongLi (approved):
+1

PR #259, opened 2026-04-17
feat: support reading blob files in data evolution flow

QuakeWang:
### Purpose

Linked issue: close #228 

Add the minimal read-only `.blob` read path for `BlobType`, so Rust can read Java-generated blob files in the existing data-evolution flow.

### Brief change log

  - add `.blob` format dispatch
  - add dedicated `.blob` reader for `BlobType`
  - support blob footer/index parsing, local row selection, and Arrow `Binary` output
  - extend `DataEvolutionReader` with rolling blob merge support via internal blob source planning
  - add Java compatibility coverage for blob format and rolling blob e2e


crates/paimon/src/table/data_evolution_reader.rs
```diff
@@ -343,113 +267,1289 @@ impl DataEvolutionReader {
...
+                ),
+                source: None,
+            })?;
+
+            let source_idx = if let Some(&existing_idx) = blob_source_indices.get(&field_id) {
+                existing_idx
+            } else {
+                let source_idx = sources.len();
+                sources.push(FieldSource::BlobBunch {
+                    bunch: BlobBunch::new(expected_row_count, false),
```
JingsongLi: Always false here?

crates/blob_test_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+    file_bytes
+}
+
+pub(crate) fn write_blob_file(path: &Path, rows: &[Option<&[u8]>]) {
+    let file_bytes = build_blob_file_bytes(rows);
+    fs::write(path, file_bytes)
+        .unwrap_or_else(|e| panic!("Failed to write blob test file {path:?}: {e}"));
+}
+
+fn encode_delta_varints(values: &[i64]) -> Vec<u8> {
```
JingsongLi: Duplicated in blob.rs

crates/paimon/src/arrow/format/blob.rs
```diff
@@ -0,0 +1,748 @@
...
+            &RecordBatchOptions::new().with_row_count(Some(positions.len())),
+        )
+        .map_err(|e| Error::UnexpectedError {
+            message: format!("Failed to build empty blob RecordBatch: {e}"),
+            source: Some(Box::new(e)),
+        });
+    }
+
+    let mut builder = BinaryBuilder::new();
+    for &position in positions {
```
JingsongLi: Is it possible to read blobs in parallel?

crates/paimon/src/table/data_evolution_reader.rs
QuakeWang: I removed the unused `row_id_push_down` flag and simplified `BlobBunch::new`.

crates/blob_test_utils.rs
QuakeWang: I removed the duplicated helper in `blob.rs` tests and reused `blob_test_utils::encode_delta_varints`.

crates/paimon/src/arrow/format/blob.rs
QuakeWang: I changed blob payload reads to bounded parallel reads with `buffered(...)` while preserving output order, and added a test for it.

JingsongLi (approved):
+1

PR #260, opened 2026-04-17
feat: add Vortex columnar file format support as optional feature

JingsongLi:
### Purpose

Add read/write support for the Vortex file format behind a `vortex` feature flag. The writer uses a background task with kanal channel for streaming writes, and the reader supports predicate pushdown and row selection. Also introduces a configurable `file.format` option (default "parquet") to replace hardcoded file extensions across all writer paths.

### DIFF
--- a/crates/paimon/src/arrow/format/vortex.rs
+++ b/crates/paimon/src/arrow/format/vortex.rs
@@ -130,10 +130,11 @@ impl FormatFileReader for VortexFormatReader {
         // Build the target Arrow schema for the projected fields.
         let target_schema = crate::arrow::build_target_arrow_schema(read_fields)?;
 
-        // Empty projection (e.g. SELECT COUNT(*)): return a single zero-column batch
-        // with the correct row count, matching Parquet/ORC/Avro behavior.
         if read_fields.is_empty() {
-            let row_count = vortex_file.row_count() as usize;
+            let row_count = match &row_selection {
+                Some(ranges) => ranges.iter().map(|r| r.count() as usize).sum(),
+                None => vortex_file.row_count() as usize,
+            };
             let batch = RecordBatch::try_new_with_options(
                 target_schema,
                 vec![],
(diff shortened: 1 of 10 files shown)


crates/paimon/src/arrow/format/vortex.rs
```diff
@@ -0,0 +1,1303 @@
...
+        // Build projection expression for requested fields.
+        let projected_names: Vec<&str> = read_fields.iter().map(|f| f.name()).collect();
+
+        let mut scan_builder = vortex_file.scan().map_err(|e| Error::DataInvalid {
+            message: format!("Failed to create Vortex scan: {e}"),
+            source: None,
+        })?;
+
+        // Apply column projection.
+        if !projected_names.is_empty() {
```
XiaoHongbo-Hope: DataFusion can push projection=[] for SELECT COUNT(*). Here empty projection is treated as no projection, so Vortex still returns actual columns, and RecordBatch validation later fails against the empty requested schema.

crates/paimon/src/arrow/format/vortex.rs
```diff
@@ -0,0 +1,1367 @@
...
+        let target_schema = crate::arrow::build_target_arrow_schema(read_fields)?;
+
+        // Empty projection (e.g. SELECT COUNT(*)): return a single zero-column batch
+        // with the correct row count, matching Parquet/ORC/Avro behavior.
+        if read_fields.is_empty() {
+            let row_count = vortex_file.row_count() as usize;
+            let batch = RecordBatch::try_new_with_options(
+                target_schema,
+                vec![],
+                &arrow_array::RecordBatchOptions::new().with_row_count(Some(row_count)),
```
XiaoHongbo-Hope, line 141: This early return ignores predicate filters for empty-projection reads. Empty-projection reads can return the full file
  row count instead of the filtered row count.

JingsongLi: We should consider `row_selection`. Predicates are inherently designed for coarse-grained filtering; unlike in Python, the returned data is not filtered on a per-row basis.

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #261, opened 2026-04-17
remove stale raw_convertible builder calls

liang767339:
Clean up stale test setup in the data evolution reader tests after `raw_convertible` is now derived automatically
  from split data files.

  ### Brief change log

  - Remove obsolete `with_raw_convertible(false)` calls from the data evolution reader tests.
  - Align the test setup with the current split reading behavior.
  - Keep the change scoped to tests only, with no production logic change.

  ### Tests

  Not run.

  ### API and Format

  ### Documentation

  No documentation updates are required.


JingsongLi (approved):
+1

PR #262, opened 2026-04-18
feat: Enable exact COUNT(*) pushdown via partition statistics

JingsongLi:
### Purpose

Add filter_exact tracking to PaimonTableScan so DataFusion's aggregate_statistics optimizer can eliminate scans for COUNT(*) queries when row counts are precisely known.

- Add filter_exact field to PaimonTableScan and PaimonScanBuilder
- Return Precision::Exact(num_rows) when merged_row_count is known for all splits, no limit is applied, and filter is exact
- Fix filter_exact computation: no predicate should be considered exact
- Refactor build_paimon_scan into PaimonScanBuilder struct to satisfy clippy::too_many_arguments
- Add count_pushdown.rs integration tests covering:
  - COUNT(*) without filter (pushes down)
  - COUNT(*) with data filter (does not push down)
  - COUNT(*) with partition filter (pushes down)
  - COUNT(*) with mixed partition+data filter (does not push down)
  - COUNT(*) on empty/single-row tables (pushes down)
- Run all paimon-datafusion tests in CI instead of just read_tables

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -125,15 +125,6 @@ impl TableSchema {
         new_schema
     }
 
-    /// Create a copy of this schema with specified options removed.
-    pub fn copy_removing_options(&self, keys: &[&str]) -> Self {
-        let mut new_schema = self.clone();
-        for key in keys {
-            new_schema.options.remove(*key);
-        }
-        new_schema
-    }
-
     /// Apply a list of schema changes and return a new schema with incremented ID.
     pub fn apply_changes(&self, changes: Vec<crate::spec::SchemaChange>) -> crate::Result<Self> {
         let mut new_schema = self.clone();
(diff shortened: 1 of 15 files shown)


crates/paimon/src/spec/schema.rs
```diff
@@ -125,6 +125,31 @@ impl TableSchema {
         new_schema
     }
 
+    /// Apply a list of schema changes and return a new schema with incremented ID.
+    pub fn apply_changes(&self, changes: Vec<crate::spec::SchemaChange>) -> crate::Result<Self> {
```
XiaoHongbo-Hope, line 129: It looks like schema change validation is missing on this path. For example, changes to primary keys,
  partition keys, `data-evolution.enabled`, and other schema-related options seem to bypass revalidation
  here.

JingsongLi: Currently, there is no systematic validation yet, which we will look at later. Actually, it is not that important. The actual validation occurs on the REST server.

XiaoHongbo-Hope: > 

Ok, looks good to me.

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #264, opened 2026-04-19
feat(datafusion): Add $snapshots system table

plusplusjiajia:
### Purpose
  Mirrors Java SnapshotsTable: exposes `<table>$snapshots` with 14 columns matching Java `SnapshotsTable.TABLE_TYPE`.

### DIFF
--- a/crates/paimon/src/table/snapshot_manager.rs
+++ b/crates/paimon/src/table/snapshot_manager.rs
@@ -151,10 +151,19 @@ impl SnapshotManager {
         self.find_by_list_files(i64::min).await
     }
 
-    /// List all snapshot ids sorted ascending.
+    /// List all snapshot ids sorted ascending. Returns an empty vector when
+    /// the snapshot directory does not exist.
     pub async fn list_all_ids(&self) -> crate::Result<Vec<i64>> {
         let snapshot_dir = self.snapshot_dir();
-        let statuses = self.file_io.list_status(&snapshot_dir).await?;
+        let statuses = match self.file_io.list_status(&snapshot_dir).await {
+            Ok(s) => s,
+            Err(crate::Error::IoUnexpected { ref source, .. })
+                if source.kind() == opendal::ErrorKind::NotFound =>
+            {
+                return Ok(Vec::new());
+            }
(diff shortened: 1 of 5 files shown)


crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -90,26 +91,20 @@ impl SnapshotManager {
         id_str.trim().parse().ok()
     }
 
-    /// List snapshot files and find the id using the given reducer (min or max).
-    async fn find_by_list_files(&self, reducer: fn(i64, i64) -> i64) -> crate::Result<Option<i64>> {
```
JingsongLi, line 94: Why modify this?

plusplusjiajia: The refactor shares the directory-scan logic between the existing find_by_list_files and the new list_all_ids, I will revert it to keep this PR scoped to $snapshots.

crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -150,6 +151,28 @@ impl SnapshotManager {
         self.find_by_list_files(i64::min).await
     }
 
+    /// List all snapshot ids sorted ascending.
+    pub async fn list_all_ids(&self) -> crate::Result<Vec<i64>> {
+        let snapshot_dir = self.snapshot_dir();
+        let statuses = self.file_io.list_status(&snapshot_dir).await?;
```
luoyuxia: Seems it will return error when the snapshot dir doesn't exists. Is it expected to return error instead of empty vec?

luoyuxia (commented):
Thanks for the pr. Left minor comments. PTAL

crates/paimon/src/table/snapshot_manager.rs
plusplusjiajia: @luoyuxia  Good catch, thanks! Updated to return an empty vec when the
  snapshot directory doesn't exist.

luoyuxia (approved):

JingsongLi (approved):
+1

PR #265, opened 2026-04-19
build: check in Cargo.lock

plusplusjiajia:
### Purpose
Committing `Cargo.lock` is no longer considered an antipattern for libraries: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

Peer ASF project apache/opendal also commits its lockfile.
e.g., opendal has it: https://github.com/apache/opendal/blob/main/core/Cargo.lock

paimon-rust is a mixed workspace producing distributable artifacts (`bindings/c` cdylib+staticlib, `bindings/python` cdylib used by pypaimon_rust, consumed by `bindings/go`), so a committed lockfile guarantees reproducible builds of those artifacts.


JingsongLi:
As a library crate, paimon-rust generally doesn't commit `Cargo.lock`. The official Rust documentation also recommends this. If the team wants reproducible CI, commits need to be updated regularly with `cargo.lock`, which adds an extra maintenance burden.

plusplusjiajia:
Thanks for the review, @JingsongLi ! Makes sense, closing this. Fair points on the library convention and maintenance burden.

PR #266, opened 2026-04-19
[WIP ]ci: use rustc 1.95.0 to satisfy constant_time_eq@0.4.3

JingsongLi:
### Purpose

Fix https://github.com/apache/paimon-rust/actions/runs/24619981104/job/71988803916

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - uses: dtolnay/rust-toolchain@stable
+        with:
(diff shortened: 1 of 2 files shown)


QuakeWang (commented):
This fixes the immediate CI breakage, but it also raises the effective Rust version we validate in CI from the workspace-declared `rust-version = "1.86.0"` to 1.95.0. 

After this change, CI will no longer catch regressions against our stated MSRV, so the project contract becomes inconsistent. We should either update `Cargo.toml` to 1.95.0 in this PR as well, or keep CI aligned with the declared MSRV and fix the dependency drift instead.

PR #267, opened 2026-04-19
feat: add Lumina vector index read infrastructure

XiaoHongbo-Hope:
### Purpose

Subtask of #227


crates/paimon/src/lumina/reader.rs
```diff
@@ -0,0 +1,316 @@
...
+            (distances, labels)
+        };
+
+        let id_to_scores = collect_results(&labels, &distances, effective_k, index_metric);
+        if id_to_scores.is_empty() {
+            return Ok(None);
+        }
+
+        Ok(Some(Box::new(DictBasedScoredIndexResult::new(
+            id_to_scores,
```
JingsongLi: Here can just return ids like tantivy. We don't need to introduce `DictBasedScoredIndexResult`.

crates/paimon/src/globalindex/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+use roaring::RoaringTreemap;
+use std::collections::BinaryHeap;
+use std::sync::Arc;
+
+pub type ScoreGetter = Arc<dyn Fn(u64) -> f32 + Send + Sync>;
+
+pub trait GlobalIndexResult: Send + Sync {
```
JingsongLi: Remove this file.

JingsongLi (commented):
You can just implement whole feature up to datafusion integration in this PR.

crates/paimon/Cargo.toml
```diff
@@ -39,6 +39,7 @@ storage-fs = ["opendal/services-fs"]
 storage-oss = ["opendal/services-oss"]
 storage-s3 = ["opendal/services-s3"]
 storage-hdfs = ["opendal/services-hdfs-native"]
+lumina = ["libloading"]
```
JingsongLi: `libloading` is already an indirect dependency of the project (brought in via other crates). If you add it as a direct dependency, it won't introduce any new crates to the build—it simply adds an explicit entry to your `Cargo.toml`. The impact on the package size is essentially negligible.

So, no need to add lumina feature.

crates/paimon/src/lumina/reader.rs
```diff
@@ -0,0 +1,316 @@
...
+    ) -> crate::Result<()> {
+        if self.searcher.is_some() {
+            return Ok(());
+        }
+
+        let index_meta = LuminaIndexMeta::deserialize(&self.io_meta.metadata)?;
+
+        let mut searcher_options = strip_lumina_options(&self.options);
+        for (k, v) in index_meta.options().iter() {
+            searcher_options.insert(k.to_string(), v.to_string());
```
luoyuxia: is it expected that `searcher_options` will overwrite `self.options` which i assume is the option passed by users.

luoyuxia (commented):
@XiaoHongbo-Hope Thanks for the pr. I left minor comments. PTAL

XiaoHongbo-Hope:
> You can just implement whole feature up to datafusion integration in this PR.

Sure, added. And fixed other comments two.

crates/paimon/src/lumina/reader.rs
XiaoHongbo-Hope: > is it expected that `searcher_options` will overwrite `self.options` which i assume is the option passed by users.

Thanks, fixed

JingsongLi (approved):
+1

PR #269, opened 2026-04-20
feat(datafusion): add copy-on-write DML support for append-only tables

JingsongLi:
### Purpose

Implement CopyOnWriteMergeWriter that rewrites entire data files for row-level updates and deletes on append-only tables. Add DELETE, UPDATE execution and enhance MERGE INTO with COW strategy, along with comprehensive integration tests.


crates/integrations/datafusion/src/merge_into.rs
```diff
@@ -582,29 +1074,320 @@ fn extract_source_ref(source: &TableFactor) -> DFResult<(String, Option<String>)
...
+    if partition_keys.is_empty() {
+        return Ok(None);
+    }
+
+    let cols = partition_keys
+        .iter()
+        .map(|k| format!("{s_alias}.\"{k}\""))
+        .collect::<Vec<_>>()
+        .join(", ");
+    let sql = format!("SELECT DISTINCT {cols} FROM {source_ref} AS {s_alias}");
```
jerry-024, line 1348: **Must fix: partitioned MERGE is incorrect.** This assumes the source has columns named identically to the target's partition keys. If source lacks them → runtime `column not found`; if source has same-named columns with different meaning, or the ON maps a differently-named source column (`t.dt = s.event_date`), `partition_set` silently omits partitions the MERGE actually touches — rows that should UPDATE get inserted as NOT MATCHED. Minimum fix: fall back to `Ok(None)` (full-partition scan). Proper fix: resolve source→target partition mapping from the ON condition.

crates/integrations/datafusion/src/merge_into.rs
```diff
@@ -84,7 +113,470 @@ pub(crate) fn is_row_id_conflict(err: &DataFusionError) -> bool {
...
+
+        for mc in &clauses.matched {
+            // Build WHERE clause: exclude rows consumed by previous clauses, then apply this predicate
+            let mut conditions: Vec<String> = Vec::new();
+            for prev in &consumed_predicates {
+                conditions.push(format!("NOT ({prev})"));
+            }
+            if let Some(ref pred) = mc.predicate {
+                conditions.push(pred.clone());
+                consumed_predicates.push(pred.clone());
```
jerry-024, line 442: **Must fix: multi-clause ordering is wrong.** When `mc.predicate` is `None` (unconditional clause), nothing is pushed into `consumed_predicates`, so later clauses re-match the same rows. Combined with `cow_writer.rs` strict checks, legal SQL like `WHEN MATCHED THEN UPDATE ... / WHEN MATCHED AND s.flag=1 THEN UPDATE ...` fails at commit with `duplicate UPDATE operations` or `both DELETE and UPDATE`. Fix: when predicate is `None`, `consumed_predicates.push("TRUE".into())`, or reject same-kind clauses after an unconditional one (unreachable by SQL semantics). Same bug in the inserts loop at line 825 — fix both.

crates/integrations/datafusion/src/update.rs
```diff
@@ -141,6 +157,158 @@ async fn execute_update_once(
...
+    update: &Update,
+    writer: &mut CopyOnWriteMergeWriter,
+) -> DFResult<u64> {
+    let select_parts: Vec<String> = std::iter::once("\"__paimon_file_idx\"".to_string())
+        .chain(std::iter::once("\"__paimon_row_offset\"".to_string()))
+        .chain(
+            columns
+                .iter()
+                .zip(exprs.iter())
+                .map(|(col, expr)| format!("{expr} AS \"__upd_{col}\"")),
```
jerry-024: **Must fix: inconsistent identifier quoting.** Here `col` is raw `id.value` (unescaped) wrapped as `"__upd_{col}"`. Elsewhere (`insert_select_clause`) `ins.columns` comes from `Ident::to_string()` (already quoted) then gets wrapped again in `"{field}"` — double-quoted. Same statement emits inconsistent SQL for a column containing `"`, and a user with DDL privileges who creates a column named `x" FROM t; DROP TABLE t; --` gets genuine injection. Fix: one helper `fn quote_identifier(name: &str) -> String { Ident { value: name.into(), quote_style: Some('"') }.to_string() }` and route every identifier interpolation through it. Remove the double-wrap in `insert_select_clause`.

crates/integrations/datafusion/src/update.rs
```diff
@@ -141,6 +157,158 @@ async fn execute_update_once(
...
+
+/// Single attempt of CoW UPDATE execution.
+async fn execute_cow_update_once(
+    ctx: &SessionContext,
+    update: &Update,
+    table: &Table,
+) -> DFResult<DataFrame> {
+    let (columns, exprs) = extract_set_assignments(update)?;
+
+    let table_ref = update.table.to_string();
```
jerry-024, line 189: **Must fix: `UPDATE t AS x` parses but fails at runtime.** `update.table.to_string()` keeps the alias in `table_ref`, but the inner query selects from the generated `__cow_target_N` (no alias), so `WHERE x.c = 1` resolves against a non-existent name and errors with `column x.c not found`. Same pattern in `delete.rs` (`extract_delete_table_ref` drops the alias but the user's WHERE still uses it). Minimum acceptable fix: explicitly reject `alias.is_some()` at both entry points with `DataFusionError::Plan("table alias in UPDATE/DELETE is not yet supported")`. Full fix: register the temp table under the user's alias.

crates/integrations/datafusion/src/delete.rs
```diff
@@ -0,0 +1,141 @@
...
+        .await
+        .map_err(to_datafusion_error)?;
+
+    let (has_data, cow_table_name) = register_cow_target_table(ctx, table, &writer).await?;
+    if !has_data {
+        return ok_result(ctx, 0);
+    }
+
+    let result = execute_cow_delete_inner(ctx, &cow_table_name, delete, &mut writer).await;
+    let _ = ctx.deregister_table(&cow_table_name);
```
jerry-024: **Must fix: MemTable leaks on async cancellation.** Sequential `let _ = ctx.deregister_table(...)` is not RAII. If the future is cancelled (timeout, panic mid-SQL, ctx drop) between register and deregister, the `MemTable` (holds all target rows) stays in the `SessionContext`. `COW_TABLE_COUNTER` is monotonically increasing so names never collide — the leak is silent and unbounded on long-lived contexts. Fix: wrap `cow_table_name` in a Drop guard that calls `ctx.deregister_table` synchronously in `Drop`. Same pattern in `update.rs:200` and the merge_into register path.

jerry-024 (commented):
Overall design is sound (CoW writer + `validate_deleted_files` + retry loop, concurrent-DELETE conflict detection works). Five must-fix items inline before merge.

jerry-024 (approved):
+1

PR #270, opened 2026-04-21
feat(blob): add blob descriptor write support for append-only tables

JingsongLi:
### Purpose

Introduce BlobDescriptor serialization/deserialization, AppendBlobFileWriter for writing blob format files, blob format writer implementation, and descriptor mode in the reader that returns BlobDescriptor references instead of inline data. Update row tracking in commit to use per-column counters for blob files. Allow BlobType in append-only tables (still unsupported with primary keys).

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 const DELETION_VECTORS_ENABLED_OPTION: &str = "deletion-vectors.enabled";
 const DATA_EVOLUTION_ENABLED_OPTION: &str = "data-evolution.enabled";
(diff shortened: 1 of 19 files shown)


crates/paimon/src/spec/core_options.rs
```diff
@@ -369,6 +378,24 @@ impl<'a> CoreOptions<'a> {
...
+    pub fn blob_as_descriptor(&self) -> bool {
+        self.options
+            .get(BLOB_AS_DESCRIPTOR_OPTION)
+            .map(|v| v.eq_ignore_ascii_case("true"))
+            .unwrap_or(false)
+    }
+
+    /// Comma-separated BLOB field names stored as serialized BlobDescriptor
+    /// bytes inline in normal data files (no .blob files for these fields).
+    pub fn blob_descriptor_fields(&self) -> HashSet<String> {
```
jerry-024, line 393: `blob-descriptor-field` needs schema-level validation before we use this set. In Java, every configured field must exist and must be a top-level `BLOB` field. Here we only parse the option string, so typos or nested / non-BLOB fields are silently accepted and we diverge from Java behavior. Can we validate this during schema construction / table initialization instead of letting it flow into the writer path?

crates/paimon/src/table/table_write.rs
```diff
@@ -249,13 +260,18 @@ impl TableWrite {
             ))
         };
 
+        let has_blob_fields = schema.fields().iter().any(|f| {
+            f.data_type().contains_blob_type() && !blob_descriptor_fields.contains(f.name())
```
jerry-024: This is broader than the Java blob contract. `contains_blob_type()` recurses into nested types, so a column like `ROW<blob BLOB>` now flips `has_blob_fields` and routes the whole top-level column into `AppendBlobFileWriter`, but the blob writer still only accepts a single top-level `BinaryArray`. That makes the new nested-blob acceptance path a false positive: the schema is accepted, then the write path fails at runtime. I think this needs to be tightened to top-level `DataType::Blob(_)` only (matching Java `BlobType.fieldsInBlobFile`) or explicitly rejected during schema validation.

jerry-024 (approved):
+1

PR #271, opened 2026-04-21
feat(datafusion): reject typed columns in PARTITIONED BY clause

JingsongLi:
### Purpose

PARTITIONED BY should only accept column references, not column definitions with types. Extract and validate the clause before sqlparser parsing since GenericDialect requires types in column defs.

### DIFF
--- a/crates/integrations/datafusion/src/sql_handler.rs
+++ b/crates/integrations/datafusion/src/sql_handler.rs
@@ -24,7 +24,7 @@
 //! SCHEMA, DROP TABLE, etc.) to the underlying [`SessionContext`].
 //!
 //! Supported DDL:
-//! - `CREATE TABLE db.t (col TYPE, ..., PRIMARY KEY (col, ...)) [PARTITIONED BY (col TYPE, ...)] [WITH ('key' = 'val')]`
+//! - `CREATE TABLE db.t (col TYPE, ..., PRIMARY KEY (col, ...)) [PARTITIONED BY (col, ...)] [WITH ('key' = 'val')]`
 //! - `ALTER TABLE db.t ADD COLUMN col TYPE`
 //! - `ALTER TABLE db.t DROP COLUMN col`
 //! - `ALTER TABLE db.t RENAME COLUMN old TO new`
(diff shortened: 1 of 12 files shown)


crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -361,6 +373,143 @@ impl PaimonSqlHandler {
...
+                }
+            }
+            b'-' if i + 1 < len && bytes[i + 1] == b'-' => {
+                i += 2;
+                while i < len && bytes[i] != b'\n' {
+                    i += 1;
+                }
+            }
+            _ => {
+                if sql[i..].len() >= 11 && sql[i..i + 11].eq_ignore_ascii_case("PARTITIONED") {
```
jerry-024: This loop advances `i` one byte at a time, but then slices `sql[i..]` and `sql[i..i + 11]`. As soon as the input contains a non-ASCII character outside `...` or `--` comments, `i` can land in the middle of a UTF-8 codepoint and panic before we return a `DataFusionError`. That makes this a blocker for me.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -92,8 +91,9 @@ impl PaimonSqlHandler {
     /// Execute a SQL statement. ALTER TABLE is handled by Paimon directly;
     /// everything else is delegated to DataFusion.
     pub async fn sql(&self, sql: &str) -> DFResult<DataFrame> {
+        let (rewritten_sql, partition_keys) = extract_partition_by(sql)?;
```
jerry-024: Running `extract_partition_by` before parsing every statement changes behavior beyond `CREATE TABLE`. The scanner currently skips only single-quoted strings and `--` comments, so `PARTITIONED BY` inside `/* ... */` comments or quoted identifiers can be misdetected and turn otherwise valid SQL into `Expected ( after PARTITIONED BY`. I think this needs to be restricted to `CREATE TABLE` parsing or made fully tokenizer/AST-aware before merge.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -361,6 +378,171 @@ impl PaimonSqlHandler {
     }
 }
 
+/// Quick check whether the SQL looks like a CREATE TABLE statement.
+fn looks_like_create_table(sql: &str) -> bool {
+    let trimmed = sql.trim_start();
+    let upper = &trimmed[..trimmed.len().min(30)];
```
jerry-024: This gate fixes the earlier regression for most non-DDL statements, but I still see two blocking issues here. First, `&trimmed[..trimmed.len().min(30)]` can panic on a UTF-8 boundary when byte 30 lands in the middle of a multibyte character (for example: `SELECT aaaaaaaaaaaaaaaaaaaa中文 FROM t`). Second, this helper skips only leading whitespace, not leading `/* ... */` or `-- ...` comments, so valid SQL like `/* note */ CREATE TABLE ... PARTITIONED BY (dt)` will bypass `extract_partition_by` and fall back to the original GenericDialect parse failure. I think this helper needs to stay byte-based and comment-aware, or be replaced with a tokenizer/parser-aware check.

jerry-024 (approved):
+1

PR #273, opened 2026-04-22
fix(datafusion): add error logging to catalog methods that silently swallow errors

shyjsarah:
…wallow errors

The CatalogProvider/SchemaProvider trait methods schema_names(), schema(), table_names(), and table_exist() have no Result return type, so errors from remote catalog operations were silently discarded. Add log::error! to surface these failures for debugging.

### Purpose
  Linked issue: N/A (no existing issue — upstream trait limitation)

  Add `log::error!` to 4 `CatalogProvider`/`SchemaProvider` trait methods that silently discard errors from remote catalog operations, making failures visible for debugging.

  ### Brief change log

  - `schema_names()`: replace `unwrap_or_default()` with `match` + `log::error!` on `Err`
  - `schema()`: replace `Err(_) => None` with `log::error!` before returning `None`
  - `table_names()`: replace `unwrap_or_default()` with `match` + `log::error!` on `Err`
  - `table_exist()`: replace `Err(_) => false` with `log::error!` before returning `false`
  - Add `log = "0.4"` dependency to `paimon-datafusion`

  These 4 methods have no `Result` return type in the upstream DataFusion trait, so logging is the best we can do. A proper fix requires changing the upstream trait signatures (to
  be proposed separately as a breaking API change on apache/datafusion).

  ### Tests

  - Existing tests pass (`cargo test -p paimon-datafusion`)
  - Manual: set `RUST_LOG=error` and trigger a catalog error to verify log output

  ### API and Format

  No API or format changes.

  ### Documentation

  No documentation changes needed.

### DIFF
--- a/crates/integrations/datafusion/Cargo.toml
+++ b/crates/integrations/datafusion/Cargo.toml
@@ -35,6 +35,7 @@ async-trait = "0.1"
 chrono = "0.4"
 constant_time_eq = { workspace = true }
 datafusion = { workspace = true }
+log = "0.4"
 paimon = { path = "../../paimon" }
 futures = "0.3"
 serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #274, opened 2026-04-22
feat(datafusion): add session-scoped dynamic options via SET/RESET

JingsongLi:
### Purpose

Support `SET 'paimon.key' = 'value'` and `RESET 'paimon.key'` to set session-level dynamic options that are automatically merged into table scans via `copy_with_options`. Non-paimon-prefixed SET/RESET falls through to DataFusion.

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -53,18 +53,10 @@ impl Debug for PaimonCatalogProvider {
 }
 
 impl PaimonCatalogProvider {
-    /// Creates a new [`PaimonCatalogProvider`].
+    /// Creates a new [`PaimonCatalogProvider`] with shared dynamic options.
     ///
     /// All data is loaded lazily when accessed.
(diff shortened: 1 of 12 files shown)


crates/integrations/datafusion/src/catalog.rs
```diff
@@ -54,7 +57,18 @@ impl PaimonCatalogProvider {
     ///
     /// All data is loaded lazily when accessed.
     pub fn new(catalog: Arc<dyn Catalog>) -> Self {
-        PaimonCatalogProvider { catalog }
+        Self::with_dynamic_options(catalog, Default::default())
```
jerry-024: PaimonCatalogProvider::new() now creates its own fresh DynamicOptions, and PaimonSqlHandler::new() does the same. That means the default public construction path does not actually share session state: SET paimon.* mutates the handler-local map, but later table resolution here reads from a different map. Unless every caller switches to with_dynamic_options(...) on both sides, the feature is effectively broken for the documented/default API.

jerry-024 (approved):
+1

PR #280, opened 2026-04-23
feat(datafusion): add Hive-style INSERT OVERWRITE PARTITION support and simplify API

JingsongLi:
### Purpose

Add static partition overwrite via PARTITION clause (e.g. INSERT OVERWRITE t PARTITION (dt = 'value') SELECT ...). Partial partition specs use predicate-based filtering to overwrite all matching sub-partitions. Simplify PaimonSqlHandler::new to 3 args (ctx, catalog, name) with internal DynamicOptions management, catalog registration, and relation planner setup. Remove redundant overwrite_partitions from CommitEntriesPlan, unifying all overwrite filtering through PartitionFilter.

### DIFF
--- a/crates/integrations/datafusion/src/sql_handler.rs
+++ b/crates/integrations/datafusion/src/sql_handler.rs
@@ -436,10 +436,9 @@ impl PaimonSqlHandler {
             .await
             .map_err(to_datafusion_error)?;
 
-        let partition_exprs = insert
-            .partitioned
-            .as_ref()
-            .expect("guarded by match: partitioned is Some and non-empty");
+        let partition_exprs = insert.partitioned.as_ref().ok_or_else(|| {
+            DataFusionError::Plan("INSERT OVERWRITE PARTITION requires a PARTITION clause".into())
+        })?;
         let partition_fields = table.schema().partition_fields();
         let static_partitions =
             parse_static_partitions(partition_exprs, &partition_fields, table.schema().fields())?;
@@ -456,7 +455,7 @@ impl PaimonSqlHandler {
             .filter(|f| !static_partitions.contains_key(f.name()))
             .count();
 
-        if let Some(first) = batches.iter().find(|b| b.num_rows() > 0) {
+        if let Some(first) = batches.first() {
             if first.num_columns() != expected_source_cols {
                 return Err(DataFusionError::Plan(format!(
                     "Source query has {} columns, but expected {} non-partition columns",
@@ -1059,9 +1058,6 @@ fn parse_number_datum(n: &str, data_type: &PaimonDataType, negate: bool) -> DFRe
                 DataFusionError::Plan(format!("Invalid DOUBLE: {e}"))
             })?))
         }
-        _ if negate => Err(DataFusionError::Plan(format!(
-            "Cannot negate value for type {data_type:?}"
-        ))),
         _ => Err(DataFusionError::Plan(format!(
             "Cannot convert {n} to {data_type:?}"
         ))),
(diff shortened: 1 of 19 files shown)


crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -399,6 +421,88 @@ impl PaimonSqlHandler {
...
+        }
+
+        let messages = tw.prepare_commit().await.map_err(to_datafusion_error)?;
+        let commit = wb.new_commit();
+
+        commit
+            .overwrite(messages, Some(static_partitions))
+            .await
+            .map_err(to_datafusion_error)?;
+
```
XiaoHongbo-Hope, line 555: `PARTITION (dt)` is all-dynamic, but this path still passes `Some(static_partitions)` into `overwrite()`.
  With an empty map from `parse_static_partitions()`, this will drop all existing partitions.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -399,6 +421,88 @@ impl PaimonSqlHandler {
...
+            .new_write()
+            .map_err(to_datafusion_error)?
+            .with_overwrite();
+        let mut row_count = 0u64;
+        let mut col_checked = false;
+
+        while let Some(batch_result) = stream.next().await {
+            let batch = batch_result?;
+            if batch.num_rows() == 0 {
+                continue;
```
XiaoHongbo-Hope, line 516: This skips column-count/type validation for an empty source result. For example, `INSERT OVERWRITE t
  PARTITION (dt = '2024-01-01') SELECT id FROM t WHERE false` should fail because the source only produces`id` while the target still expects `id, name`, but this path skips validation and the later `overwrite()` still truncates the target partition.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -399,6 +421,88 @@ impl PaimonSqlHandler {
...
+                if batch.num_columns() != expected_source_cols {
+                    return Err(DataFusionError::Plan(format!(
+                        "Source query has {} columns, but expected {} non-partition columns",
+                        batch.num_columns(),
+                        expected_source_cols
+                    )));
+                }
+                col_checked = true;
+            }
+            let augmented = append_partition_columns(
```
luoyuxia, line 531: This path is currently treating the source columns as positional "all non-static table fields in schema order". That breaks `INSERT OVERWRITE ... PARTITION (...)` when the statement uses an explicit target column list, because this handler never applies `insert.columns` / `after_columns` before calling `append_partition_columns`.

For example, if the table schema is `(dt, id, name)` and the statement is:

`INSERT OVERWRITE t (name, id) PARTITION (dt = '2024-01-01') VALUES ('alice', 1)`

then the source columns here are `(name, id)`, but `append_partition_columns` will consume them as `(id, name)` in table-schema order. That means we either get an unexpected cast failure (`'alice' -> id`) or silently write the wrong values when the types are compatible.

We should preserve target-column mapping here and reorder/project the source batch by column name before appending static partition columns.

crates/paimon/src/table/bucket_assigner.rs
```diff
@@ -97,4 +97,10 @@ impl BucketAssignerEnum {
             Self::CrossPartition(a) => a.prepare_commit_index(file_io, index_dir).await,
         }
     }
+
+    pub fn set_overwrite(&mut self, is_overwrite: bool) {
+        if let Self::Dynamic(a) = self {
```
XiaoHongbo-Hope, line 102: set_overwrite() does not affect CrossPartitionAssigner. With cross-partition PK tables and merge-engine='first-row', INSERT OVERWRITE ... PARTITION can delete old data without writing the replacement
row, causing data loss.

JingsongLi: CrossPartitionAssigner cannot support well to OVERWRITE. The semantic will be very strange.

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #283, opened 2026-04-25
feat(datafusion): Add $tags system table

plusplusjiajia:
### Purpose
Add the `$tags` Paimon system table to the DataFusion integration.  Mirrors Java [TagsTable](https://github.com/apache/paimon/blob/release-1.3/paimon-core/src/main/java/org/apache/paimon/table/system/TagsTable.java).

  ```sql
  SELECT * FROM paimon.default.my_table$tags;


JingsongLi (approved):
+1

PR #287, opened 2026-04-25
feat(datafusion): Add $manifests system table

plusplusjiajia:
### Purpose
 Add the `$manifests` Paimon system table to the DataFusion integration. Mirrors Java
  [`ManifestsTable`](https://github.com/apache/paimon/blob/release-1.3/paimon-core/src/main/java/org/apache/paimon/table/system/ManifestsTable.java).

  ```sql
  SELECT * FROM paimon.default.my_table$manifests;


JingsongLi (commented):
Please rebase and update `datafusion.md` too.

plusplusjiajia:
> Please rebase and update `datafusion.md` too.

@JingsongLi Good catch! Done, rebased and added $manifests (plus backfilled $snapshots/$tags) to the System Tables section.

JingsongLi (approved):
+1

PR #288, opened 2026-04-26
feat(spec, api, catalog): add Catalog::list_partitions[_paged] for RESTCatalog

plusplusjiajia:
### Purpose
Add `Catalog::list_partitions` and `list_partitions_paged` for catalogs that track partitions via metastore (e.g. `RESTCatalog`). Mirrors Java
  [`Catalog`](https://github.com/apache/paimon/blob/release-1.4/paimon-core/src/main/java/org/apache/paimon/catalog/Catalog.java) interface and
  [`AbstractCatalog`](https://github.com/apache/paimon/blob/release-1.4/paimon-core/src/main/java/org/apache/paimon/catalog/AbstractCatalog.java) default impl.

  - **spec**: new `Partition` struct (11 fields, JSON wire-format compatible with Java)
  - **api**: `RESTApi::list_partitions[_paged]` + `ListPartitionsResponse` + `partitions(db,t)` URL builder
  - **catalog**: trait gains `list_partitions[_paged]` with default impls (return empty / single full page)
  - **catalog/rest**: `RESTCatalog` overrides both, swallowing `NotImplemented` so callers can fall back to scanning manifest entries

### DIFF
--- a/crates/paimon/src/api/api_response.rs
+++ b/crates/paimon/src/api/api_response.rs
@@ -258,6 +258,29 @@ impl ListTablesResponse {
     }
 }
 
+/// Response for listing partitions.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ListPartitionsResponse {
+    /// List of partitions.
+    pub partitions: Option<Vec<crate::spec::Partition>>,
+    /// Token for the next page.
+    pub next_page_token: Option<String>,
+}
+
+impl ListPartitionsResponse {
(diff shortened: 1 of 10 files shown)


JingsongLi:
It would be best to understand why we need to support this.

plusplusjiajia:
> It would be best to understand why we need to support this.

@JingsongLi Good question. The motivation is the upcoming $partitions system table — without this, it can only surface 5 of the 9 columns Java exposes (audit fields like created_at/created_by/updated_by/options come from RESTCatalog.listPartitions).

JingsongLi (approved):
+1

PR #290, opened 2026-04-27
feat(datafusion): support branch_manager and system table

xuzifu666:
### Purpose

Linked issue: close #xxx

Branch management capabilities are frequently used in feature-driven scenarios, but are not currently supported in the Rust version. 
This PR adds support for branch_manager and branch system table capabilities in the datafusion module.

### DIFF
--- a/crates/paimon/src/table/branch_manager.rs
+++ b/crates/paimon/src/table/branch_manager.rs
@@ -110,8 +110,8 @@ impl BranchManager {
 
     /// Check if a branch exists.
     pub async fn branch_exists(&self, branch_name: &str) -> crate::Result<bool> {
-        let names = self.list_all().await?;
-        Ok(names.iter().any(|n| n == branch_name))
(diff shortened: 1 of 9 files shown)


crates/paimon/src/table/branch_manager.rs
```diff
@@ -0,0 +1,577 @@
...
+                message: format!("Branch name '{}' already exists.", branch_name),
+                source: None,
+            });
+        }
+        Ok(())
+    }
+
+    /// Check if a branch exists.
+    pub async fn branch_exists(&self, branch_name: &str) -> crate::Result<bool> {
+        let names = self.list_all().await?;
```
JingsongLi: Do not list all for `exists`, performance bad.

xuzifu666: Yes, it has been changed to determine based on branch name.

JingsongLi (approved):
+1

PR #291, opened 2026-04-27
feat(datafusion): add CALL procedure support for tag and rollback operations

JingsongLi:
### Purpose

//! CALL procedure support for Paimon tables.
//!
//! Supported procedures:
//! - `CALL sys.create_tag(table => '...', tag => '...', snapshot_id => ...)`
//! - `CALL sys.delete_tag(table => '...', tag => '...')`
//! - `CALL sys.rollback_to(table => '...', snapshot_id => ... | tag => '...')`
//! - `CALL sys.rollback_to_timestamp(table => '...', timestamp => ...)`
//! - `CALL sys.create_tag_from_timestamp(table => '...', tag => '...', timestamp => ...)`

### DIFF
--- a/crates/integrations/datafusion/src/procedures.rs
+++ b/crates/integrations/datafusion/src/procedures.rs
@@ -37,9 +37,92 @@ use datafusion::sql::sqlparser::ast::{
     FunctionArguments, ObjectName, Value as SqlValue,
 };
 use paimon::catalog::{Catalog, Identifier};
+use paimon::spec::Snapshot;
 use paimon::table::{SnapshotManager, Table, TagManager};
 
 use crate::error::to_datafusion_error;
+
+/// Resolve a snapshot by id: try live snapshot file first, then fall back to tag metadata.
+async fn resolve_snapshot_by_id(
+    sm: &SnapshotManager,
+    tm: &TagManager,
+    snapshot_id: i64,
+) -> DFResult<Snapshot> {
+    if let Ok(snap) = sm.get_snapshot(snapshot_id).await {
+        return Ok(snap);
+    }
+    let tags = tm.list_all().await.map_err(to_datafusion_error)?;
+    for (_, snap) in &tags {
+        if snap.id() == snapshot_id {
+            return Ok(snap.clone());
+        }
+    }
+    Err(DataFusionError::Plan(format!(
+        "Snapshot '{snapshot_id}' does not exist in live files or tag metadata"
+    )))
+}
+
+/// Find the earliest snapshot with commit time >= timestamp_millis,
+/// considering both live snapshots and tag-retained snapshots.
+async fn later_or_equal_from_all(
+    sm: &SnapshotManager,
+    tm: &TagManager,
+    timestamp_millis: i64,
+) -> DFResult<Option<Snapshot>> {
+    let live = sm
+        .later_or_equal_time_millis(timestamp_millis)
+        .await
+        .map_err(to_datafusion_error)?;
+    let tags = tm.list_all().await.map_err(to_datafusion_error)?;
+    let tag_candidate = tags
+        .into_iter()
+        .map(|(_, snap)| snap)
+        .filter(|s| (s.time_millis() as i64) >= timestamp_millis)
+        .min_by_key(|s| s.time_millis());
+    match (live, tag_candidate) {
+        (Some(a), Some(b)) => {
+            if a.time_millis() <= b.time_millis() {
+                Ok(Some(a))
+            } else {
+                Ok(Some(b))
+            }
+        }
+        (Some(a), None) => Ok(Some(a)),
+        (None, Some(b)) => Ok(Some(b)),
+        (None, None) => Ok(None),
+    }
+}
+
+/// Find the latest snapshot with commit time <= timestamp_millis,
+/// considering both live snapshots and tag-retained snapshots.
+async fn earlier_or_equal_from_all(
+    sm: &SnapshotManager,
+    tm: &TagManager,
+    timestamp_millis: i64,
+) -> DFResult<Option<Snapshot>> {
+    let live = sm
+        .earlier_or_equal_time_millis(timestamp_millis)
+        .await
+        .map_err(to_datafusion_error)?;
+    let tags = tm.list_all().await.map_err(to_datafusion_error)?;
+    let tag_candidate = tags
+        .into_iter()
+        .map(|(_, snap)| snap)
+        .filter(|s| (s.time_millis() as i64) <= timestamp_millis)
+        .max_by_key(|s| s.time_millis());
+    match (live, tag_candidate) {
+        (Some(a), Some(b)) => {
+            if a.time_millis() >= b.time_millis() {
+                Ok(Some(a))
+            } else {
+                Ok(Some(b))
+            }
+        }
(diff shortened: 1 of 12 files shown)


crates/integrations/datafusion/src/procedures.rs
```diff
@@ -0,0 +1,416 @@
...
+        .transpose()?;
+
+    let (sm, tm) = managers(&table);
+    if tm.tag_exists(tag_name).await.map_err(to_datafusion_error)? {
+        return Err(DataFusionError::Plan(format!(
+            "Tag '{tag_name}' already exists"
+        )));
+    }
+    let snapshot = if let Some(id) = snapshot_id {
+        sm.get_snapshot(id).await.map_err(to_datafusion_error)?
```
jerry-024: This only resolves `snapshot_id` from the live snapshot files. In Java Paimon, creating a tag from a snapshot id also works when that snapshot file has already expired but is still retained by an existing tag. With this implementation, `CALL sys.create_tag(..., snapshot_id => ...)` incorrectly fails for tag-retained snapshots, which defeats one of the main retention use cases for tags. Please resolve the id from both snapshot files and tag metadata.

crates/integrations/datafusion/src/procedures.rs
```diff
@@ -0,0 +1,416 @@
...
+    let table = get_table(catalog, catalog_name, args).await?;
+    let tag_name = require_arg(args, "tag")?;
+    let ts_str = require_arg(args, "timestamp")?;
+    let timestamp: i64 = ts_str
+        .parse()
+        .map_err(|_| DataFusionError::Plan(format!("Invalid timestamp: '{ts_str}'")))?;
+
+    let (sm, tm) = managers(&table);
+    let snapshot = sm
+        .later_or_equal_time_millis(timestamp)
```
jerry-024: This should also consider snapshots retained only by existing tags. Java Paimon first checks live snapshots and then scans tag snapshots to choose the earliest snapshot whose commit time is >= the requested timestamp. This implementation only sees live snapshot files, so after expiration it can choose a later snapshot or fail even though a suitable tag-retained snapshot exists.

crates/integrations/datafusion/src/procedures.rs
```diff
@@ -0,0 +1,416 @@
...
+                "rollback_to requires either 'snapshot_id' or 'tag' argument".to_string(),
+            ));
+        }
+    } else {
+        let (sm, tm) = managers(&table);
+        if let Some(id_str) = args.get("snapshot_id") {
+            let id: i64 = id_str
+                .parse()
+                .map_err(|_| DataFusionError::Plan(format!("Invalid snapshot_id: '{id_str}'")))?;
+            sm.get_snapshot(id).await.map_err(to_datafusion_error)?;
```
jerry-024: Same issue here as in `create_tag`: rollback by `snapshot_id` only accepts a live snapshot file. Java Paimon falls back to tag metadata when the requested snapshot file has expired but the snapshot is still tag-retained, and then rolls back through the tag path. As written, `CALL sys.rollback_to(..., snapshot_id => ...)` will reject valid tag-retained snapshots after expiration.

crates/paimon/src/table/snapshot_manager.rs
```diff
@@ -282,40 +282,97 @@ impl SnapshotManager {
...
+
     /// Returns the snapshot whose commit time is earlier than or equal to the given
     /// `timestamp_millis`. If no such snapshot exists, returns None.
     ///
-    /// Uses binary search over snapshot IDs (assumes monotonically increasing commit times).
+    /// Uses binary search over the actual snapshot ID list to handle gaps from deleted snapshots.
     ///
     /// Reference: [SnapshotManager.earlierOrEqualTimeMills](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java)
-    pub async fn earlier_or_equal_time_mills(
+    pub async fn earlier_or_equal_time_millis(
```
jerry-024, line 348: This renames the public API from `earlier_or_equal_time_mills` to `earlier_or_equal_time_millis`. Internal callers are updated, but downstream users of the crate calling the old public method will get a compile break. Can we keep a deprecated compatibility wrapper with the old name that forwards to this method?

crates/integrations/datafusion/src/procedures.rs
```diff
@@ -0,0 +1,416 @@
...
+    ctx: &SessionContext,
+    catalog: &Arc<dyn Catalog>,
+    catalog_name: &str,
+    args: &HashMap<String, String>,
+) -> DFResult<DataFrame> {
+    let table = get_table(catalog, catalog_name, args).await?;
+    let tag_str = require_arg(args, "tag")?;
+
+    let (_, tm) = managers(&table);
+    for tag_name in tag_str.split(',') {
```
jerry-024, line 311: This loop diverges from Java in two ways:

1. **Java is more lenient.** `Table.deleteTags(tagStr)` calls `TagManager.deleteTag` for each name, and Java's `deleteTag` logs `WARN "Tag '...' doesn't exist."` and returns silently when the tag is missing. The Rust version returns `Err` instead.
2. **Rust leaves partial state.** Calling with `tag => 'v1,v2,bad'` deletes `v1` and `v2`, then errors on `bad`. The caller sees an error and may assume nothing changed, but two tags are already gone. Same hazard for empty entries from `'v1,,v2'`.

Suggest either matching Java's warn-and-continue semantics, or doing a two-pass: split + trim + non-empty check + `tag_exists` check across all names first, then delete.

Java reference: [`Table.deleteTags`](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/table/Table.java) default impl + `TagManager.deleteTag`.

jerry-024 (commented):
Two semantic divergences from the Java reference. Details inline.

crates/integrations/datafusion/src/procedures.rs
```diff
@@ -0,0 +1,416 @@
...
+            return Err(DataFusionError::Plan(format!(
+                "Tag '{tag_name}' does not exist"
+            )));
+        }
+        tm.delete(tag_name).await.map_err(to_datafusion_error)?;
+    }
+    ok_result(ctx)
+}
+
+async fn clean_larger_than(
```
jerry-024, line 324: Missing long-lived changelog cleanup. Java's [`RollbackHelper.cleanLargerThan`](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/table/RollbackHelper.java) does three things:

1. `cleanSnapshots` — covered here (steps 1 + 2)
2. `cleanLongLivedChangelogs` — **missing**
3. `cleanTags` — covered here (step 3)

For tables with a changelog producer, this leaves changelog files newer than the rollback target on disk and never updates the long-lived-changelog `LATEST` hint (or sets it to `-1` when everything is cleaned). After rollback, orphan changelog entries can break incremental scans and confuse readers that follow the changelog hint.

The Java implementation also computes `to = max(earliest, retained_id + 1)` so it doesn't try to delete already-expired snapshots; the Rust loop is fine on that front because `list_all_ids()` only returns existing files, but worth keeping the structure aligned with Java for future maintenance.

jerry-024 (approved):
+1

PR #292, opened 2026-04-28
feat(datafusion): add TRUNCATE TABLE and DROP PARTITION SQL support

JingsongLi:
### Purpose

Wire existing TableCommit::truncate_table() and truncate_partitions() APIs to the DataFusion SQL layer, supporting:
- TRUNCATE TABLE db.t
- TRUNCATE TABLE db.t PARTITION (col = val, ...)
- ALTER TABLE db.t DROP PARTITION (col = val, ...)

### DIFF
--- a/crates/integrations/datafusion/src/sql_handler.rs
+++ b/crates/integrations/datafusion/src/sql_handler.rs
@@ -29,6 +29,9 @@
 //! - `ALTER TABLE db.t DROP COLUMN col`
 //! - `ALTER TABLE db.t RENAME COLUMN old TO new`
 //! - `ALTER TABLE db.t RENAME TO new_name`
+//! - `ALTER TABLE db.t DROP PARTITION (col = val, ...)`
+//! - `TRUNCATE TABLE db.t`
+//! - `TRUNCATE TABLE db.t PARTITION (col = val, ...)`
 
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -45,7 +48,7 @@ use datafusion::prelude::{DataFrame, SessionContext};
 use datafusion::sql::sqlparser::ast::{
     AlterTableOperation, ColumnDef, CreateTable, CreateTableOptions, Delete, Expr as SqlExpr,
     FromTable, Insert, Merge, ObjectName, RenameTableNameKind, Reset, ResetStatement, Set,
-    SqlOption, Statement, TableFactor, TableObject, Update, Value as SqlValue,
+    SqlOption, Statement, TableFactor, TableObject, Truncate, Update, Value as SqlValue,
 };
 use datafusion::sql::sqlparser::dialect::GenericDialect;
 use datafusion::sql::sqlparser::parser::Parser;
@@ -195,6 +198,7 @@ impl PaimonSqlHandler {
                 }
                 self.ctx.sql(sql).await
             }
+            Statement::Truncate(truncate) => self.handle_truncate_table(truncate).await,
             _ => self.ctx.sql(sql).await,
         }
     }
@@ -325,6 +329,16 @@ impl PaimonSqlHandler {
                             changes.push(SchemaChange::set_option(key.value.clone(), v));
                         }
                     }
+                }
+                // DropPartitions is a data operation (not a schema change), so we handle it
+                // separately and return early — it cannot be combined with schema changes.
+                // `if_exists` is intentionally ignored: the underlying overwrite is a no-op
+                // when the partition doesn't exist, which matches IF EXISTS semantics.
+                AlterTableOperation::DropPartitions {
+                    partitions,
+                    if_exists: _,
+                } => {
+                    return self.handle_drop_partitions(&identifier, partitions).await;
                 }
                 other => {
                     return Err(DataFusionError::Plan(format!(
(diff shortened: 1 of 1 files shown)


crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -973,6 +1057,60 @@ fn extract_options(opts: &CreateTableOptions) -> DFResult<Vec<(String, String)>>
...
+        };
+
+        if !partition_keys.iter().any(|k| k == &col_name) {
+            return Err(DataFusionError::Plan(format!(
+                "Column '{col_name}' is not a partition column"
+            )));
+        }
+
+        let field = field_map.get(col_name.as_str()).ok_or_else(|| {
+            DataFusionError::Plan(format!("Column '{col_name}' not found in table schema"))
```
jerry-024, line 1119: **HIGH: silent wrong-partition deletion on incomplete specs.**

This builds `partition` only from keys the user supplied. The map flows through `TableCommit::truncate_partitions` → `partitions_to_bytes`:

```rust
let datum = p.get(key).cloned().flatten();  // missing key → None
```

`None` is encoded into the partition filter as `IS NULL`. So on a `(dt, region)` partitioned table, `DROP PARTITION (dt = '2026-04-28')` becomes a filter `dt='2026-04-28' AND region IS NULL` — it does **not** delete all `dt='2026-04-28'` partitions like Hive/Spark prefix semantics would. If a NULL-region partition exists, it gets deleted (silent wrong-target). If not, silent no-op. Either way, user intent is mishandled with no error.

Two acceptable fixes:
- **Reject incomplete specs explicitly:** after the loop, `if partition.len() != partition_keys.len()` → `Err` listing the missing keys.
- **Implement true prefix-match deletion:** requires a partition-filter primitive that matches on a subset of keys; not just a downstream-API tweak.

For reference, Java's Spark [`PaimonPartitionManagement.toPaimonPartitions`](https://github.com/apache/paimon/blob/master/paimon-spark/paimon-spark-common/src/main/scala/org/apache/paimon/spark/PaimonPartitionManagement.scala) does `require(partitionFieldCount <= partitionKeys.length)` and projects the partition row type so prefix specs work as Hive expects.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -556,6 +570,76 @@ impl PaimonSqlHandler {
...
+        })?;
+        let identifier = self.resolve_table_name(&target.name)?;
+        let table = self
+            .catalog
+            .get_table(&identifier)
+            .await
+            .map_err(to_datafusion_error)?;
+
+        let wb = table.new_write_builder();
+        let commit = wb.new_commit();
```
jerry-024, line 594: **MEDIUM: empty `PARTITION ()` clause falls through to a full-table truncate.**

If `truncate.partitions` is `Some(empty_vec)` (whatever AST shape sqlparser produces for a degenerate parser edge case), this branch is skipped and execution drops into `commit.truncate_table()` below — the entire table is wiped despite the user writing a `PARTITION` clause. Opposite of intent.

Suggest tightening:

```rust
if let Some(partitions) = &truncate.partitions {
    if partitions.is_empty() {
        return Err(DataFusionError::Plan(
            "PARTITION clause requires at least one column = value".to_string(),
        ));
    }
    // ... existing branch
    return ok_result(...);
}
commit.truncate_table().await?;
```

This way only `TRUNCATE TABLE t` (no PARTITION clause at all, i.e. `truncate.partitions` is `None`) reaches `truncate_table()`.

crates/integrations/datafusion/src/sql_handler.rs
```diff
@@ -326,6 +330,16 @@ impl PaimonSqlHandler {
                         }
                     }
                 }
+                // DropPartitions is a data operation (not a schema change), so we handle it
+                // separately and return early — it cannot be combined with schema changes.
```
jerry-024: **MEDIUM: `if_exists` dropped at three levels — none honored.**

1. **Inner `AlterTableOperation::DropPartitions { if_exists: _ }`** (this match arm): partition-level IF EXISTS is explicitly bound and discarded. The doc comment claims this "matches IF EXISTS semantics" because the underlying overwrite is a no-op on missing partitions (verified — `truncate_partitions` early-returns on empty resolved entries). But that makes plain `DROP PARTITION` behave identically to `DROP PARTITION IF EXISTS`. Spark's `AlterTableDropPartitionCommand` errors when IF EXISTS is omitted and the partition doesn't exist — this PR diverges silently.

2. **Outer `Statement::AlterTable { if_exists, .. }`**: returning early from this branch into `handle_drop_partitions` skips whatever outer IF EXISTS handling the rest of `handle_alter_table` does. `ALTER TABLE IF EXISTS missing_table DROP PARTITION (...)` will fail with a hard error from `catalog.get_table(...)` instead of being a silent no-op.

3. **`Statement::Truncate { if_exists, .. }`** (handler at line 570): sqlparser parses the flag (confirmed in `sqlparser::ast::Truncate`), but `handle_truncate_table` never reads it. `TRUNCATE TABLE IF EXISTS missing_table` errors instead of no-oping.

Suggested fix: at each layer, read the flag, and on `get_table` returning NotFound (or by `list_tables` first), short-circuit to `ok_result` when `if_exists` is set, otherwise propagate the error.

jerry-024 (commented):
Three must-fix items, posted inline. The HIGH one (partial partition spec) is silent wrong-data deletion, not just a behavior nuance.

jerry-024 (approved):
+1

PR #294, opened 2026-05-01
feat(datafusion): Add $partitions system table

plusplusjiajia:
### Purpose
  Add the `$partitions` Paimon system table to the DataFusion integration. Mirrors Java
  [`PartitionsTable`](https://github.com/apache/paimon/blob/release-1.4/paimon-core/src/main/java/org/apache/paimon/table/system/PartitionsTable.java).

  ```sql
  SELECT * FROM paimon.default.my_table$partitions;


JingsongLi (approved):
+1

PR #295, opened 2026-05-01
fix(datafusion): wrap system table scan IO in await_with_runtime to support FFI

plusplusjiajia:
### Purpose
 System tables panic with `there is no reactor running` when accessed via FFI (e.g. pypaimon SQL), because the FFI path doesn't carry a Tokio runtime context across the boundary. Rust `#[tokio::test]` doesn't expose this since it always has a runtime entered.

  Wraps the async IO in `scan()` of 5 IO-dependent system tables (`$schemas`, `$snapshots`, `$tags`, `$branches`, `$manifests`) in `crate::runtime::await_with_runtime`, matching the existing pattern in `catalog.rs` / `table/mod.rs` / `vector_search.rs`. `$options` is unchanged (no IO).

### DIFF
--- a/crates/integrations/datafusion/src/system_tables/branches.rs
+++ b/crates/integrations/datafusion/src/system_tables/branches.rs
@@ -79,31 +79,11 @@ impl TableProvider for BranchesTable {
         _filters: &[Expr],
         _limit: Option<usize>,
     ) -> DFResult<Arc<dyn ExecutionPlan>> {
-        let bm = BranchManager::new(
-            self.table.file_io().clone(),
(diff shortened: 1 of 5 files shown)


JingsongLi (commented):
Can we create tests in Python?

plusplusjiajia:
> Can we create tests in Python?

@JingsongLi Thanks for the suggestion! Opened https://github.com/apache/paimon/pull/7762 with Python e2e tests covering all 5 system tables. PTAL when you have a moment.

JingsongLi (approved):
+1

PR #306, opened 2026-05-03
feat: expose catalog metadata api for pypaimon_rust

### LINKED ISSUE #284: Expose catalog metadata API (list_databases, list_tables, schema) for pypaimon_rust
### Search before asking

### Motivation

Paimon-rust's python bindings (path: bindings/python) currently expose only a DataFusion entry point:

```python
catalog = PaimonCatalog({"metastore": "filesystem", "warehouse": "..."})
ctx = SessionContext()
ctx.register_catalog("paimon", catalog)  # only access path
ctx.sql("SELECT * FROM paimon.mydb.orders")
```

This makes it impossible to use paimon-rust as a drop-in replacement for pypaimon in metadata-browsing and schema-introspection use cases. The following operations that pypaimon supports today are entirely missing:

- Listing databases and tables
- Fetching table schema (columns, types, primary keys, partition keys, table options)

### Solution

Add the following methods to PaimonCatalog and introduce three new wrapper classes:

```python
# PaimonCatalog (additions)
catalog.list_databases() -> list[str]
catalog.list_tables(database_name: str) -> list[str]
catalog.get_table(identifier: str) -> PyTable        # "db.table" format

# PyTable (new)
table.identifier() -> str
table.location()   -> str
table.schema()     -> PyTableSchema

# PyTableSchema (new)
schema.fields()         -> list[PyDataField]
schema.partition_keys() -> list[str]
schema.primary_keys()   -> list[str]
schema.options()        -> dict[str, str]            # e.g. bucket, changelog-mode
schema.comment()        -> str | None

# PyDataField (new)
field.name()        -> str
field.field_type()  -> str                           # e.g. "INT NOT NULL", "TIMESTAMP(6)"
field.is_nullable() -> bool
field.description() -> str | None
```

Intended usage (parity with pypaimon):
```python
catalog = PaimonCatalog({"metastore": "filesystem", "warehouse": "/tmp/paimon"})

for db in catalog.list_databases():
    for tbl in catalog.list_tables(db):
        table  = catalog.get_table(f"{db}.{tbl}")
        schema = table.schema()
        print(schema.fields(), schema.primary_keys(), schema.options())
```

### Anything else?

_No response_

### Willingness to contribute

SML0127:
### Purpose

Linked issue: close https://github.com/apache/paimon-rust/issues/284

### Brief change log

- added list_databases, list_tables, and get_table methods to existing PaimonCatalog.
- introduced new PyO3 wrapper classes for metadata inspection: Table, TableSchema, and DataField.

### API and Format

- Python-only API additions. No changes to Rust public APIs or storage formats.


leaves12138 (changes requested):
Thanks for the PR. The runtime binding implementation looks generally good, and the existing CI is green. I also verified locally that `cargo test -q -p pypaimon_rust --no-default-features` passes after the first build.

One blocking issue: this package ships `py.typed` and `bindings/python/python/pypaimon_rust/datafusion.pyi`, but the stub file still only exposes `PaimonCatalog.__init__` and `__datafusion_catalog_provider__`. The new public APIs added by this PR (`list_databases`, `list_tables`, `get_table`, plus the new `Table`, `TableSchema`, and `DataField` classes and their methods) are missing from the type stubs.

That means dynamic Python calls work, but type-aware users and IDEs will still see the old API. Please update `bindings/python/python/pypaimon_rust/datafusion.pyi` to match the new exported Python surface. It would also be helpful to cover this with a small stub/import check if the project already has a place for that.

Also, GitHub currently reports the PR as not mergeable (`mergeable_state=dirty`), so it will need a rebase/update before merge.

SML0127:
@leaves12138 Thanks for the comment. I've updated and rebbased the branch.

JingsongLi (commented):
Thanks for the contribution! The catalog metadata API additions look good overall.

Review comments:

1. **Type stubs updated** — I see this PR now includes updates to `datafusion.pyi` with `DataField`, `TableSchema`, `Table`, and the new `PaimonCatalog` methods. This addresses the concern from the previous review. Good.

2. **`get_table` identifier parsing**: The `splitn(2, '.')` approach works for simple cases but will fail for database names containing dots (e.g., `my.db.my_table`). While this is unlikely in practice, consider documenting that database names must not contain dots, or using the first dot as the separator (which is what you're already doing with `splitn(2, '.')`). The current implementation is fine.

3. **Missing observability APIs in this PR's `Table` class (vs PR #307)**: This PR exposes `identifier()`, `location()`, and `schema()` only. The observability methods (`latest_snapshot`, `list_snapshots`, etc.) are in the follow-up #307. This is a clean separation — metadata in #306, observability in #307.

4. **Test coverage**: `test_list_databases_and_tables` covers the happy path well. Consider adding a test for `get_table` with an invalid identifier format (e.g., missing dot) to verify the `PyValueError` is raised correctly.

5. **Mergeable state**: As the previous reviewer noted, this PR has merge conflicts (`mergeable_state=dirty`). Please rebase onto main before this can be merged.

6. **Architecture note**: Storing both `catalog: Arc<dyn Catalog>` and `provider: Arc<PaimonCatalogProvider>` in `PaimonCatalog` is slightly redundant since `PaimonCatalogProvider` internally also holds the catalog. But it's a clean approach that avoids reaching through the provider, so it's fine.

Please rebase and this should be ready to merge.

JingsongLi (approved):
+1

leaves12138 (approved):
Thanks for the updates. The previous blocking concern about the Python type stubs has been addressed, and the PR now exposes the new catalog metadata APIs consistently in both the runtime binding and datafusion.pyi. I also checked the PR merge ref locally with cargo fmt --all --check, cargo check -p pypaimon_rust --no-default-features, and cargo test -p pypaimon_rust --no-default-features; all passed. GitHub CI is green as well. +1.

PR #307, opened 2026-05-03
expose table observability API for pypaimon_rust

### LINKED ISSUE #285: Expose paimon table observability API (snapshots, partitions, tags) for pypaimon_rust
### Search before asking

### Motivation

Beyond catalog metadata (issue https://github.com/apache/paimon-rust/issues/284), pypaimon_rust also lacks table-level introspection APIs that pypaimon exposes today: snapshot history, partition layout, and tags are not accessible from Python.

### Solution

Add the following APIs to PyTable and introduce new wrapper classes:

```python
# PyTable (additions)
table.latest_snapshot()  -> PySnapshot | None
table.list_snapshots()   -> list[PySnapshot]   # full history, newest first

# PySnapshot (new)
snapshot.id()                  -> int
snapshot.commit_time_ms()      -> int         # Snapshot::time_millis(), epoch ms
snapshot.total_record_count()  -> int | None  # None if not tracked
snapshot.delta_record_count()  -> int | None  # None if not tracked
snapshot.commit_kind()         -> str         # "APPEND" | "COMPACT" | "OVERWRITE" | "ANALYZE"

# PyTable (additions)
table.list_partitions()  -> list[dict[str, str]]
# e.g. [{"dt": "2024-01-01", "hr": "10"}, ...]

table.partition_stats()  -> list[PyPartitionStat]

# PyPartitionStat (new)
stat.partition()         -> dict[str, str]
stat.record_count()      -> int
stat.file_count()        -> int
stat.total_size_bytes()  -> int

# PyTable (addition)
table.list_tags() -> list[PyTag]

# PyTag (new)
tag.name()        -> str
tag.snapshot_id() -> int
```

Intended usage:
```python
table = catalog.get_table("mydb.orders")
snap = table.latest_snapshot()

if snap:
    print(snap.commit_time_ms(), snap.total_record_count(), snap.commit_kind())

for stat in table.partition_stats():
    print(stat.partition(), stat.record_count(), stat.total_size_bytes())

for tag in table.list_tags():
    print(tag.name(), tag.snapshot_id())
```

### Anything else?

_No response_

### Willingness to contribute

SML0127:
### Purpose

Linked issue: close https://github.com/apache/paimon-rust/issues/285

### Brief change log

- added PartitionStat and Table::partition_stats / list_partitions.
- exposed latest_snapshot, list_snapshots, list_partitions, partition_stats, and list_tags


SML0127:
I'll rebase this pr once https://github.com/apache/paimon-rust/pull/306 is merged

JingsongLi (commented):
Thanks for the contribution! This extends the Python bindings with useful observability APIs (snapshots, tags, partitions).

A few review comments:

1. **PR #306 dependency**: This PR builds on #306 (catalog metadata API). Given that #306 still has requested changes outstanding (type stubs and rebase), this PR should wait for #306 to merge first. Consider rebasing once #306 is resolved.

2. **`partition_stat.rs` — manifest entry scan approach**: The `read_all_manifest_entries` implementation reads ALL manifests from both `base_manifest_list` and `delta_manifest_list`, then aggregates stats by partition key bytes. For tables with many manifests, this could be expensive. Consider:
   - Adding a note about expected performance characteristics.
   - Or in a follow-up, exploring whether the `$partitions` system table path in DataFusion could be reused instead of duplicating the scan logic.

3. **`partition_value_to_string`** — the fallback branch `_ => row.get_string(pos_i)` will silently misinterpret binary types (VARBINARY, BINARY) as UTF-8 strings. Consider returning a hex encoding or explicitly erroring for unsupported partition key types.

4. **Snapshot ordering**: `list_snapshots()` reverses the list (`into_iter().rev()`) so newest is first. The Python test asserts `snaps[0].id() == snap.id()` — which is correct. But this ordering convention should be documented in the Python type stubs or docstring.

5. **Missing type stubs**: The new `Snapshot`, `Tag`, and `PartitionStat` classes need `.pyi` entries in `datafusion.pyi` for IDE support. Same issue as raised on #306.

6. **Thread safety of `runtime().block_on()`**: The `list_snapshots`, `latest_snapshot`, `list_tags`, and `partition_stats` methods all call `runtime().block_on(...)` from Python. If called from within an async Python context (e.g., inside an `asyncio` event loop), this will panic. Consider documenting this limitation.

Overall the implementation looks correct and the test coverage is reasonable. The main blocker is the dependency on #306 and the missing type stubs.

SML0127:
Hi @JingsongLi 
I've rebased the branch. Could you take a look?

crates/paimon/src/table/partition_stat.rs
```diff
@@ -0,0 +1,212 @@
...
+fn aggregate_partition_stats(
+    entries: &[ManifestEntry],
+    partition_keys: &[String],
+    partition_fields: &[DataField],
+) -> crate::Result<Vec<PartitionStat>> {
+    let mut grouped: HashMap<Vec<u8>, Accum> = HashMap::new();
+    for entry in entries {
+        let bucket = grouped.entry(entry.partition().to_vec()).or_default();
+        let file = entry.file();
+        let live_rows = file.row_count - file.delete_row_count.unwrap_or(0);
```
JingsongLi: `delete_row_count` should not be subtracted when producing these partition statistics. In Paimon, `DataFileMeta::row_count` is the physical row count and `delete_row_count` is only the number of DELETE row-kind records inside that file (`rowCount = add_row_count + delete_row_count`). The existing commit/catalog partition statistics paths aggregate `file.row_count` directly, so this new API will under-report PK/changelog files and still will not represent the true logical live row count. Please keep this consistent with the existing partition statistics semantics, or document/expose a separately computed logical count.

crates/paimon/src/table/partition_stat.rs
```diff
@@ -0,0 +1,212 @@
...
+    if keys.is_empty() {
+        return Ok(map);
+    }
+    let row = BinaryRow::from_serialized_bytes(bytes)?;
+    for (i, key) in keys.iter().enumerate() {
+        let dt = fields
+            .iter()
+            .find(|f| f.name() == key)
+            .map(|f| f.data_type());
+        let value = if row.is_null_at(i) {
```
JingsongLi: This custom decoder returns partition values that differ from the values used everywhere else in the table. For example null partitions should use `partition.default-name`, non-legacy date/time/timestamp partitions should be formatted with `PartitionComputer`, and decimals should use the configured scale instead of the raw unscaled integer. As a result `list_partitions()` / `partition_stats()` can report partition specs that do not match the catalog/system-table view for the same table. Please reuse `PartitionComputer::generate_part_values` here (or expose a helper) so the observability API uses the same partition formatting rules as writes and catalog partition listing.

JingsongLi:
This PR currently conflicts with the latest main branch. I confirmed it locally with `git merge-tree --write-tree origin/main codex/review-pr-307`, which reports a content conflict in `crates/paimon/src/table/mod.rs`. Please rebase on current main and resolve the conflict before this can be merged.

crates/paimon/src/table/partition_stat.rs
SML0127: Thanks for the feedback, fixed to use  `file.row_count`  directly as suggested.

crates/paimon/src/table/partition_stat.rs
SML0127: Thanks for the feedback. I replaced the custom decoder with `PartitionComputer::generate_part_values()` as suggested

SML0127:
Hi @JingsongLi 
Thanks you for comments. I resolved all above issues. PTAL

bindings/python/python/pypaimon_rust/datafusion.pyi
```diff
@@ -36,10 +36,59 @@ class TableSchema:
...
+    def latest_snapshot(self) -> Optional[Snapshot]:
+        """
+        Warning: This method blocks on a DataFusion runtime.
+        Calling this from an active asyncio event loop will result in a panic.
+        """
+        ...
+    def list_snapshots(self) -> List[Snapshot]:
+        """
+        Returns all snapshots ordered newest first (descending by ID).
+
```
JingsongLi: This line has trailing whitespace, which makes `git diff --check origin/main...HEAD` fail. Please remove it.

crates/paimon/src/table/partition_stat.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    Ok(all_entries)
+}
+
+fn aggregate_partition_stats(
+    entries: &[ManifestEntry],
+    computer: &PartitionComputer,
+) -> crate::Result<Vec<PartitionStat>> {
+    let mut grouped: HashMap<Vec<u8>, Accum> = HashMap::new();
+    for entry in entries {
```
JingsongLi: This should probably aggregate the same live manifest entries as the existing catalog partition listing path instead of summing every raw ADD/DELETE entry. `catalog::list_partitions_from_file_system` reads these manifests and calls `merge_active_entries` before counting files/rows, which keeps the stats based on the active file set by identifier. With the current signed accumulation, duplicate ADD entries or tombstones/compaction records can drift from scan/catalog semantics and produce different partition counts. Can we reuse that live-entry logic here (or call `merge_active_entries` before aggregating) so `partition_stats()` and `Catalog::list_partitions()` stay consistent?

SML0127:
fix ci bugs in this pr https://github.com/apache/paimon-rust/pull/418

crates/paimon/src/table/partition_stat.rs
```diff
@@ -0,0 +1,171 @@
...
+fn aggregate_partition_stats(
+    entries: Vec<ManifestEntry>,
+    computer: &PartitionComputer,
+) -> crate::Result<Vec<PartitionStat>> {
+    // Mirror catalog::partition_listing::list_partitions_from_file_system:
+    // drop DELETE-shadowed entries first so stats are based on the active
+    // (live) file set, consistent with scan and catalog partition semantics.
+    let active = merge_active_entries(entries);
+
+    let mut grouped: HashMap<Vec<u8>, Accum> = HashMap::new();
```
JingsongLi: This makes the public `partition_stats()` / `list_partitions()` order nondeterministic. The existing catalog fallback uses a `BTreeMap<Vec<u8>, _>` when grouping partition bytes, so callers see a stable partition order across runs; here the `HashMap` iteration order can vary between processes and make the Python API awkward to test or compare. Could we use the same `BTreeMap` grouping (or sort `out` by the serialized partition bytes before returning) so the observability API stays deterministic and consistent with `Catalog::list_partitions()`?

SML0127: Thanks for the great feedback! 
I used btreemap for deterministic partition ordering.

bindings/python/tests/test_datafusion.py
```diff
@@ -685,3 +685,40 @@ def test_list_databases_and_tables():
...
+
+    # Non-partitioned, non-tagged table: both should be empty but well-typed.
+    parts = table.list_partitions()
+    stats = table.partition_stats()
+    tags = table.list_tags()
+
+    assert isinstance(parts, list)
+    assert isinstance(stats, list)
+    assert isinstance(tags, list)
+    # simple_log_table has no partition keys -> partition_stats yields a single
```
QuakeWang, line 901: This test is too weak for the main risk in this PR. `partition_stats()` is only smoke-tested on a non-partitioned table, and the test accepts either 0 or 1 empty partition, so it would not catch regressions in `merge_active_entries`, partition value formatting, or per-partition aggregation. Could we add a temporary-warehouse test with a real partitioned table and assert concrete `list_partitions()` / `partition_stats()` values, including record/file/size counts?

JingsongLi:
Looks good to me.

bindings/python/tests/test_datafusion.py
SML0127: Thanks for the feedback. I added  `test_partition_stats_with_partitioned_table()`  and  `test_partition_stats_excludes_overwritten_partition()`  to cover the cases you mentioned.

crates/paimon/src/table/partition_stat.rs
```diff
@@ -0,0 +1,174 @@
...
+        all_entries.extend(entries);
+    }
+    Ok(all_entries)
+}
+
+fn aggregate_partition_stats(
+    entries: &[ManifestEntry],
+    computer: &PartitionComputer,
+) -> crate::Result<Vec<PartitionStat>> {
+    let mut grouped: HashMap<Vec<u8>, Accum> = HashMap::new();
```
leaves12138: This still groups raw manifest entries without calling `merge_active_entries`, and it still uses `HashMap`. The new overwrite test only exercises a simple ADD/DELETE pair where signed sums happen to work; duplicate ADD entries present across base/delta manifests remain double-counted, and return order remains randomized between processes. Please reuse the live-entry semantics from `list_partitions_from_file_system` (`merge_active_entries` followed by `BTreeMap` aggregation) and add a duplicate-entry regression.

leaves12138 (changes requested):
Thanks for adding the partitioned and overwrite tests. The Python binding compiles and formatting passes locally, but the current head still leaves the two manifest-aggregation issues unresolved.

`partition_stats()` continues to aggregate raw ADD/DELETE entries directly and groups them in a `HashMap`. The overwrite test only covers an ordinary ADD/DELETE pair where signed arithmetic happens to match the live set; it does not cover duplicate ADD entries across base/delta manifests, which are double-counted here but collapsed by `merge_active_entries`. The `HashMap` iteration also keeps the public result order nondeterministic despite the earlier reply saying it was switched to a `BTreeMap`.

Please mirror `catalog::list_partitions_from_file_system`: merge active entries first, then aggregate the remaining ADD files in a `BTreeMap<Vec<u8>, _>`. A focused synthetic duplicate-entry test would guard the actual failure mode better than the current overwrite-only case.

crates/paimon/src/table/partition_stat.rs
SML0127: Thank you for the detailed feedback. Fixed `aggregate_partition_stats` to call  `merge_active_entries`  first, then accumulate live ADD entries in a  BTreeMap .

JingsongLi (approved):
+1

leaves12138 (approved):
Reviewed the latest changes. Looks good to me.

PR #308, opened 2026-05-05
feat(datafusion): populate partition stats in $manifests system table

QuakeWang:
### Purpose

The DataFusion `$manifests` system table already defines `min_partition_stats` and `max_partition_stats`, but currently returns `NULL` for both columns. This change decodes the partition stats stored in `ManifestFileMeta` and renders them with Java-compatible row cast semantics.

### Brief change log

  - Add a DataFusion-internal formatter for Java-style `BinaryRow` row-to-string output.
  - Populate `$manifests.min_partition_stats` and `$manifests.max_partition_stats`.
  - Avoid formatting partition stats when the columns are not projected.
  - Handle empty stats bytes for unpartitioned tables, all-null partition stats, and unknown stats.
  - Add focused formatter tests and a DataFusion integration test for partition stats.


JingsongLi (commented):
Can you also update doc?

QuakeWang:
> Can you also update doc?

@JingsongLi I have updated this, PTAL :)

JingsongLi (approved):
+1

PR #309, opened 2026-05-06
feat(datafusion): add catalog-level temporary table support

JingsongLi:
### Purpose

This PR introduces catalog-level temporary table and view support for the DataFusion integration. Previously, temporary tables were not scoped to a specific Paimon catalog. This change allows users to create, query, and drop temporary tables and views within any registered Paimon catalog using standard SQL syntax.

Also add Python API:
```python
import pyarrow as pa

batch = pa.record_batch([[1, 2], ["alice", "bob"]], names=["id", "name"])

ctx.register_batch("paimon.default.my_temp", batch)

batches = ctx.sql("SELECT * FROM paimon.default.my_temp")

# Drop it via SQL when no longer needed
ctx.sql("DROP TEMPORARY TABLE paimon.default.my_temp")
```

### Brief change log

- Catalog-level temp storage: Added temp_tables to PaimonCatalogProvider (backed by RwLock<HashMap<String, MemorySchemaProvider>>) so that temporary tables are organized per database inside each catalog.
- SchemaProvider resolution: Updated PaimonSchemaProvider to check temporary tables first, falling back to real Paimon tables. This ensures temp tables can shadow persistent tables without naming conflicts.
- SQLContext APIs: Added register_temp_table, deregister_temp_table, and temp_table_exist for programmatic temp table management with flexible TableReference names.
- SQL parsing: Extended SQLContext::sql to intercept and handle:
     - CREATE TEMPORARY TABLE (col TYPE, ...)
     - CREATE TEMPORARY TABLE ... AS SELECT ...
     - CREATE TEMPORARY VIEW ... AS SELECT ...
     - DROP TEMPORARY TABLE ...
- Time-travel queries: Refactored VERSION AS OF / TIMESTAMP AS OF execution to register a UUID-based temporary table internally, avoiding global namespace pollution.
- Docs & tests: Renamed datafusion.md to sql.md, documented all new SQL syntax, and added comprehensive integration tests.

### Tests

- Added crates/integrations/datafusion/tests/sql_context_tests.rs covering:
     - CREATE TEMPORARY TABLE with column definitions
     - CREATE TEMPORARY TABLE ... AS SELECT
     - CREATE TEMPORARY VIEW ... AS SELECT
     - DROP TEMPORARY TABLE / DROP TEMPORARY VIEW
     - IF NOT EXISTS semantics
     - Temp table shadowing and lifecycle
- Updated existing test suites (read_tables.rs, merge_into_tests.rs, delete_tests.rs, etc.) to align with the new catalog resolution logic.

### API and Format

- New public APIs on SQLContext:
     - register_temp_table(name, Arc<dyn TableProvider>)
     - deregister_temp_table(name) -> Option<Arc<dyn TableProvider>>
     - temp_table_exist(name) -> bool
- New public APIs on PaimonCatalogProvider:
     - register_temp_table(database, table_name, table)
     - deregister_temp_table(database, table_name)
     - temp_table_exist(database, table_name) -> bool
- No changes to the underlying Paimon storage format.

### Documentation

- Renamed docs/src/datafusion.md to docs/src/sql.md and updated mkdocs.yml.
- Documented temporary table SQL syntax, SQLContext usage examples, and catalog registration patterns.


crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1498,6 +1942,115 @@ fn datum_to_constant_array(
...
+
+/// Extract `VERSION AS OF <n>` or `VERSION AS OF '<tag>'` from a SQL string.
+///
+/// Looks for the pattern (case-insensitive):
+/// - `... VERSION AS OF <number>` — numeric snapshot ID
+/// - `... VERSION AS OF '<tag>'` — tag name (quoted string)
+///
+/// Returns the table name, version/tag value, and byte range of the full clause.
+fn extract_version_as_of(sql: &str) -> Option<VersionAsOfInfo> {
+    let lower = sql.to_lowercase();
```
jerry-024, line 2030: **P0 — Must fix (Java Paimon compatibility)**

`lower.find("version as of ")` is a naive string search that will fail on valid SQL patterns that work in Java Paimon (Flink/Spark):

1. **JOIN two time-travel tables**: `SELECT * FROM t1 VERSION AS OF 1 JOIN t2 VERSION AS OF 2 ON ...` — only the first match is found; the second is silently ignored.
2. **Subquery / CTE**: `WITH cte AS (SELECT * FROM t VERSION AS OF 1) SELECT * FROM cte` — `clause_range` replacement corrupts the SQL structure.
3. **String literal false positive**: `WHERE note = 'version as of 1'` — matches inside a string constant.
4. **Table alias after clause**: `FROM t VERSION AS OF 1 AS tt` — replacement range doesn't account for the alias.

In Java Paimon, all of these are handled correctly by the SQL parser. Users migrating SQL from Flink/Spark will hit unexpected failures.

**Suggestion:** Consider a two-pass approach — first extract and strip time-travel clauses with awareness of quoted strings and parenthesis nesting, then hand the cleaned SQL to `Parser::parse_sql`. At minimum, add a quoted-string skip to avoid false positives inside string literals.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -275,6 +405,130 @@ impl SQLContext {
...
+        // Replace the original table name + time-travel clause with just the UUID name
+        let rewritten_sql = format!(
+            "{}{}{}",
+            &sql[..clause_range.0],
+            uuid_name,
+            &sql[clause_range.1..]
+        );
+
+        // Register the provider under the UUID temp table name
+        self.register_temp_table(uuid_name.as_str(), provider)?;
```
jerry-024: **P1 — Temp table leak on panic**

If `self.ctx.sql()` panics, this temp table is never cleaned up. This PR already introduces `TempTableTracker` as an RAII guard for exactly this pattern (used in `execute_cow_delete_once`, `execute_cow_merge_once`, etc.), but it's not used here.

Suggested fix:
```rust
let mut tracker = TempTableTracker::new(self);
self.register_temp_table(uuid_name.as_str(), provider)?;
tracker.register(&uuid_name);

let result = self.ctx.sql(&rewritten_sql).await;
// tracker auto-deregisters on drop
result
```

docs/src/sql.md
```diff
@@ -560,6 +589,94 @@ SELECT * FROM paimon.my_db.assets;
...
+## Temporary Tables
+
+You can register in-memory temporary tables under any catalog. Temporary tables exist only for the lifetime of the `SQLContext` instance and are automatically cleaned up when the context is dropped.
+
+The table name accepts flexible references, similar to DataFusion:
+- `"my_table"` — uses the current catalog and current database
+- `"database.my_table"` — uses the current catalog with the specified database
+- `"catalog.database.my_table"` — fully qualified
+
+### register_mem_table
```
jerry-024: **P1 — Non-existent API**

`register_mem_table` is documented here but doesn't exist in the code — the PR implements `register_temp_table` instead. Either add `register_mem_table` as a convenience wrapper, or update the docs to reference `register_temp_table`.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -162,13 +190,102 @@ impl CatalogProvider for PaimonCatalogProvider {
...
+    /// the same database. Logs a warning if the name shadows a real Paimon table.
+    pub fn register_temp_table(
+        &self,
+        database: &str,
+        table_name: &str,
+        table: Arc<dyn TableProvider>,
+    ) -> DFResult<()> {
+        // Check if a temp table with this name already exists
+        {
+            let databases = self.temp_tables.read().unwrap_or_else(|e| e.into_inner());
```
jerry-024: **P1 — TOCTOU between existence check and registration**

This read-lock check and the actual registration via `get_or_create_temp_database` (line 254) are done in separate lock acquisitions. Between them, there's a blocking `block_on_with_runtime` call (shadow warning, lines 235-252) where another thread could register a table with the same name.

Since `register_temp_table` is a public API, consider holding the write lock for the entire check-then-register operation, or using `MemorySchemaProvider::register_table`'s return value to detect the conflict atomically.

jerry-024 (commented):
Thanks for the comprehensive work on catalog-scoped temp tables! The core design is solid — scoping temp tables per-catalog, the `TempTableTracker` RAII pattern, and migrating all tests from `datafusion.public` to the Paimon catalog path are all good improvements.

Below are inline comments on the issues I found, prioritized by severity.

**P2 — Suggest fixing (non-blocking):**

1. **`looks_like_create_table` also matches `CREATE TEMP VIEW`** (`sql_context.rs:1273`): The TEMPORARY/TEMP skip doesn't distinguish TABLE from VIEW. If a `CREATE TEMPORARY VIEW` ever contains Paimon-specific syntax like `PARTITIONED BY`, the pre-parser would incorrectly try to extract partition keys. Currently not triggered but is a latent bug.

2. **`bindings/python/README.md` — Usage section removed without replacement**: Users reading the README on GitHub/PyPI lose the quick-start guide. Consider adding a link to `project-description.md` or restoring a minimal example.

3. **Python `register_batch` lacks negative test**: No test for calling `register_batch` with an invalid/unknown catalog name.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -275,6 +405,130 @@ impl SQLContext {
...
+                "Failed to parse time-travel clause in SQL".to_string(),
+            ));
+        };
+
+        // Resolve the table from our catalog and create a provider with scan options
+        let table_ref: datafusion::common::TableReference = table_name.as_str().into();
+        let (catalog, _catalog_name, identifier) = self.resolve_table_name_from_ref(&table_ref)?;
+
+        let paimon_table = catalog
+            .get_table(&identifier)
```
jerry-024: **P0 — Must fix (dynamic options bypass)**

`catalog.get_table(&identifier)` here loads the table directly from the Paimon catalog, bypassing the normal `PaimonSchemaProvider` path. This means session-scoped dynamic options set via `SET 'paimon.xxx' = ...` are **not applied** to time-travel queries.

For example:
```sql
SET 'paimon.blob-as-descriptor' = 'true';
SELECT * FROM paimon.default.my_table VERSION AS OF 1;  -- blob-as-descriptor NOT applied
```

The normal table loading path in `PaimonSchemaProvider::table()` applies `dynamic_options` before creating the provider. Time-travel should do the same — either pass `self.dynamic_options` to `copy_with_options` alongside the time-travel options, or go through the schema provider.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -190,8 +291,16 @@ impl SQLContext {
         } else {
             (sql.to_string(), vec![])
         };
-        let dialect = GenericDialect {};
-        let statements = Parser::parse_sql(&dialect, &rewritten_sql)
+        let sql_lower = rewritten_sql.to_lowercase();
+        let has_time_travel =
+            sql_lower.contains("version as of") || sql_lower.contains("timestamp as of");
```
jerry-024: **P0 — Must fix (false positive on string literals)**

This `contains()` check is the entry gate that short-circuits all subsequent SQL parsing. Any SQL containing the substring `version as of` or `timestamp as of` — even inside a string literal or comment — will skip `Parser::parse_sql` entirely and enter the regex-based time-travel path.

Examples that would be misrouted:
```sql
SELECT * FROM t WHERE note = 'version as of 1'
-- comment: version as of something
INSERT INTO t VALUES ('timestamp as of now')
```

This needs at minimum a quoted-string-aware scan before deciding to treat the query as time-travel. One approach: skip over single-quoted strings (`'...'`) and SQL comments (`--` / `/* */`) before checking for the keyword.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -262,6 +375,23 @@ impl SQLContext {
                 self.ctx.sql(sql).await
             }
             Statement::Truncate(truncate) => self.handle_truncate_table(truncate).await,
+            Statement::CreateView(create_view) => self.handle_create_view(create_view).await,
```
jerry-024: **P0 — Must fix (breaks non-Paimon catalogs)**

This intercepts **all** `CREATE VIEW` statements, and the `Statement::Drop` arm below (line 385) intercepts all `DROP TABLE` / `DROP VIEW`. Before this PR, these would fall through to `self.ctx.sql(sql)` and be handled by DataFusion normally.

Now:
- `CREATE VIEW datafusion.public.my_view AS ...` → errors with "CREATE VIEW (non-temporary) is not supported" (line 1113), even though DataFusion can handle it.
- `DROP TABLE datafusion.public.x` → `resolve_catalog_and_table` looks up only Paimon catalogs and fails with "Unknown catalog 'datafusion'".

**Suggestion:** Only intercept these statements when the resolved catalog is a registered Paimon catalog. If the catalog name doesn't exist in `self.catalogs`, fall through to `self.ctx.sql(sql).await` for DataFusion's default handling.

JingsongLi:
Thanks @jerry-024 for the review, addressed comments.

jerry-024 (approved):
+1

PR #310, opened 2026-05-07
fix(rest): pass catalog options to FileIO when no data token is vended

plusplusjiajia:
### Purpose
`RESTCatalog::get_table` falls through to `FileIO::from_path(path)?.build()?` when `data_token_enabled=false` (older DLF) or `is_external=true`. With no properties passed, OSS-backed paths fail at `oss_config_parse` with `Missing required OSS config: fs.oss.endpoint`.


JingsongLi (approved):
+1

PR #312, opened 2026-05-11
fix: validate data evolution matched row ids

QuakeWang:
### Purpose

Fix `DataEvolutionWriter` validation for data-evolution MERGE/UPDATE.

### Brief change log

  - Reject null `_ROW_ID`.
  - Reject duplicate updates to the same target row.
  - Validate update columns against each matched batch schema.
  - Add core unit tests and a DataFusion MERGE integration test.


JingsongLi (approved):
+1

PR #263, opened 2026-04-18
feat: support fixed-bucket partial-update merge engine

QuakeWang:
### Purpose

Linked issue: part of #256

This PR delivers a minimum usable `merge-engine=partial-update` implementation for primary-key tables with fixed buckets.
It shifts the earlier groundwork-heavy approach toward end-to-end functionality, while keeping only the validation that is still necessary for correctness.

  This PR does not close #256 yet. Support for `bucket=-1` partial-update, including cross-partition existing-key behavior, remains follow-up work.

### Brief change log

  - support fixed-bucket `merge-engine=partial-update` end to end for primary-key tables
  - keep fail-fast for unsupported advanced partial-update options
  - remove redundant validation and generic guard rails
  - implement the read/write path changes required by basic latest-non-null partial-update semantics
  - keep `bucket=-1` and `deletion-vectors.enabled=true` explicitly unsupported in this PR
  - add fixed-bucket E2E and targeted regression tests

PR #314, opened 2026-05-12
fix: repair CI after partial-update merge

QuakeWang:
### Purpose

related to #263 

This PR fixes CI failures on `main` after the partial-update changes were merged.

The failures were caused by stale call sites from the partial-update PR after upstream APIs and test helpers had changed on `main`:

  - `TableWrite::new` no longer accepts the old boolean argument.
  - `setup_handler()` no longer exists in the DataFusion PK table tests.
  - A duplicate `file_format` local variable triggered Clippy `-D warnings`.

The individual PR checks were green before merge, but the final merged result on `main` failed because the branch was not validated against the latest base before landing. A **merge queue**, or requiring branches to be up to date before merge, would help prevent this class of CI breakage.


jerry-024 (approved):
+1

PR #316, opened 2026-05-12
Add explicit read-only permissions to CI workflow

arpitjain099:
## Summary
- Add an explicit `permissions` block to `.github/workflows/ci.yml`.
- Set default token scope to `contents: read` for CI jobs.

## Why
The CI workflow only needs repository read access. Declaring this explicitly aligns with least-privilege workflow hardening.


JingsongLi (approved):
+1

PR #317, opened 2026-05-13
feat: support dynamic-bucket partial-update

QuakeWang:
### Purpose

Linked issue: follow-up to #256

This PR adds basic `merge-engine=partial-update` support for primary-key tables using `bucket=-1` dynamic bucket mode.

It extends the existing fixed-bucket partial-update support to non-cross-partition dynamic bucket tables, including partitioned tables whose primary keys include partition fields. Cross-partition dynamic bucket partial-update remains unsupported and fails fast.

### Brief change log

  - Allow `merge-engine=partial-update` for non-cross-partition `bucket=-1` PK tables.
  - Restore existing HASH index entries before assigning buckets, so updates to existing keys stay in their original buckets.
  - Keep `merge-engine=partial-update` with cross-partition dynamic bucket explicitly unsupported.
  - Reject partition-only primary keys, which would leave only one record per partition.
  - Add DataFusion integration tests for unpartitioned, partitioned, and existing-key dynamic bucket partial-update cases.

### API and Format

This PR uses the existing dynamic bucket HASH index format and does not introduce a new storage format.

### DIFF
--- a/crates/integrations/datafusion/tests/dynamic_bucket_tables.rs
+++ b/crates/integrations/datafusion/tests/dynamic_bucket_tables.rs
@@ -22,10 +22,10 @@ mod common;
 use common::{
     collect_id_name, collect_id_value, create_sql_context, create_test_env, setup_sql_context,
 };
-use datafusion::arrow::array::{Int32Array, StringArray};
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #323, opened 2026-05-18
feat: add referenced_files_size and physical_files_size table functions

JingsongLi:
### Purpose

Introduce two DataFusion table-valued functions for storage analysis:

- `referenced_files_size('table')`: Aggregates manifest/data/index file counts and sizes across all snapshots from main branch, tags, and other branches. Output rows: total, branch:main, branch:<name>. Uses a shared manifest cache to avoid redundant reads and processes snapshots concurrently.

- `physical_files_size('table')`: Scans the table directory recursively and reports actual physical file sizes categorized by file type (manifest, data, index). Concurrently lists subdirectories for high throughput on object stores.

Both functions gracefully handle NotFound errors from concurrently deleted files during cleanup.

### DIFF
--- a/crates/paimon/src/table/referenced_files.rs
+++ b/crates/paimon/src/table/referenced_files.rs
@@ -122,7 +122,12 @@ impl ExtraFileResolver {
         partition_bytes: &[u8],
         bucket: i32,
         extra_file_name: &str,
+        external_path: Option<&str>,
     ) -> Option<String> {
+        if let Some(ext_path) = external_path {
+            let dir = ext_path.trim_end_matches('/');
+            return Some(format!("{}/{}", dir, extra_file_name));
+        }
         let partition_path = if let Some(ref computer) = self.partition_computer {
             let row = BinaryRow::from_serialized_bytes(partition_bytes).ok()?;
             computer.generate_partition_path(&row).ok()?
@@ -425,8 +430,12 @@ async fn collect_snapshot_files(
                 file_entries.push((e.file().file_name.clone(), e.file().file_size));
                 for extra in &e.file().extra_files {
                     let entry_idx = file_entries.len();
-                    let full_path =
-                        extra_resolver.resolve_extra_file_path(e.partition(), e.bucket(), extra);
+                    let full_path = extra_resolver.resolve_extra_file_path(
+                        e.partition(),
+                        e.bucket(),
+                        extra,
+                        e.file().external_path.as_deref(),
+                    );
                     if let Some(path) = full_path {
                         extra_file_stat_tasks.push((manifest_idx, entry_idx, path));
                     }
@@ -490,6 +499,21 @@ async fn collect_snapshot_files(
                 .index_files
                 .entry(entry.index_file.file_name.clone())
                 .or_insert(entry.index_file.file_size as i64);
+        }
+    }
+
+    // Collect statistics file if present
+    if let Some(statistics_name) = snapshot.statistics() {
+        let statistics_path = format!(
+            "{}/statistics/{}",
+            extra_resolver.table_location, statistics_name
+        );
+        let size = try_stat_file_size(file_io, &statistics_path).await?;
+        if size > 0 {
+            file_set
+                .manifest_files
+                .entry(statistics_name.to_string())
+                .or_insert(size);
         }
     }
 
@@ -591,15 +615,17 @@ pub struct PhysicalFilesSummary {
 }
 
 /// Categorize a file name into a file type.
+/// Everything that is not a manifest/statistics or index file is classified as data.
 fn classify_file_name(file_name: &str) -> FileType {
-    if file_name.starts_with("manifest-") || file_name.starts_with("index-manifest-") {
+    if file_name.starts_with("manifest-")
+        || file_name.starts_with("index-manifest-")
+        || file_name.starts_with("statistics-")
+    {
         FileType::Manifest
-    } else if file_name.starts_with("data-") {
-        FileType::Data
     } else if file_name.starts_with("index-") {
         FileType::Index
     } else {
-        FileType::Other
+        FileType::Data
     }
 }
 
@@ -607,7 +633,6 @@ enum FileType {
     Manifest,
     Data,
     Index,
-    Other,
 }
 
 const DIR_LIST_CONCURRENCY: usize = 32;
@@ -619,10 +644,9 @@ const DIR_LIST_CONCURRENCY: usize = 32;
 /// on object stores with many partition directories.
 ///
 /// Files are classified by their file name prefix:
-/// - `manifest-*` / `manifest-list-*` / `index-manifest-*` → manifest
-/// - `data-*` → data
+/// - `manifest-*` / `index-manifest-*` → manifest
 /// - `index-*` (excluding `index-manifest-*`) → index
-/// - Other files (snapshots, schemas, etc.) are not counted.
+/// - Everything else → data
 pub async fn collect_physical_files_summary(
     file_io: &FileIO,
     table_location: &str,
@@ -693,7 +717,6 @@ fn accumulate_file(summary: &mut PhysicalFilesSummary, file_name: &str, size: u6
             summary.index_file_count += 1;
             summary.index_file_size += size as i64;
         }
-        FileType::Other => {}
     }
 }
 
--- a/crates/paimon/src/table/referenced_files.rs
+++ b/crates/paimon/src/table/referenced_files.rs
@@ -113,35 +113,42 @@ pub async fn collect_referenced_files_summary(
     let manifest_cache: ManifestCache = Mutex::new(HashMap::new());
     let manifest_cache_ref = &manifest_cache;
 
-    // 1. Main branch snapshots + tags
     let sm = SnapshotManager::new(file_io.clone(), table_location.to_string());
-    let mut main_files = collect_scope_files(file_io, &sm, manifest_cache_ref).await?;
-
     let tm = TagManager::new(file_io.clone(), table_location.to_string());
-    let tag_files = collect_tag_files(file_io, &sm, &tm, manifest_cache_ref).await?;
+
+    // 1. Main branch snapshots + tags (concurrently)
+    let (main_files, tag_files) = tokio::try_join!(
+        collect_scope_files(file_io, &sm, manifest_cache_ref),
+        collect_tag_files(file_io, &sm, &tm, manifest_cache_ref),
+    )?;
+    let mut main_files = main_files;
     main_files.merge(&tag_files);
 
-    // 2. Branch file sets (snapshots + branch-level tags)
+    // 2. Branch file sets (all branches concurrently)
     let bm = BranchManager::new(file_io.clone(), table_location.to_string());
     let branch_names = bm.list_all().await?;
-    let mut branch_file_sets = Vec::new();
-    for branch_name in &branch_names {
-        let branch_sm = sm.with_branch(branch_name);
-        let mut branch_files =
-            collect_scope_files(file_io, &branch_sm, manifest_cache_ref).await?;
 
-        let branch_tm = tm.with_branch(branch_name);
-        let branch_tag_files =
-            collect_tag_files(file_io, &branch_sm, &branch_tm, manifest_cache_ref).await?;
-        branch_files.merge(&branch_tag_files);
-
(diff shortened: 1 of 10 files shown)


crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,535 @@
...
+        })
+        .buffer_unordered(SNAPSHOT_CONCURRENCY)
+        .try_collect()
+        .await?;
+
+    let mut summary = ReferencedFilesSummary {
+        source: source.to_string(),
+        ..Default::default()
+    };
+    for s in per_snapshot.into_iter().flatten() {
```
QuakeWang: A newer Paimon snapshot carries previous base/delta manifests into its new base manifest list, so the same still-referenced manifest/data/index file can appear in many snapshot summaries. Since this function is documented as comparable with physical size for orphan-file analysis, summing snapshots can make referenced size grow with snapshot count and even exceed physical size. This should collect a de-duplicated referenced file set by file identity/path first, then aggregate sizes from that set.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,535 @@
...
+        .iter()
+        .map(|path| try_read_manifest_list(file_io, path))
+        .collect();
+    let manifest_lists = try_join_all(manifest_list_futures).await?;
+
+    // Flatten all manifest file metas from all manifest lists
+    let all_manifest_metas: Vec<&ManifestFileMeta> =
+        manifest_lists.iter().flat_map(|ml| ml.iter()).collect();
+
+    summary.manifest_file_count = all_manifest_metas.len() as i64;
```
QuakeWang: This code only counts `ManifestFileMeta` entries inside the manifest lists, but the files named by `snapshot.base_manifest_list()`, `delta_manifest_list()`, and `changelog_manifest_list()` are themselves snapshot-referenced metadata files. The `snapshot.index_manifest()` file is also read below but not counted as a referenced manifest file. Since `physical_files_size` classifies both `manifest-list-*` and `index-manifest-*` as manifest files, the two functions cannot be safely compared until these directly referenced metadata files are included and de-duplicated.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,535 @@
...
+
+        let manifest_futures: Vec<_> = uncached_paths
+            .iter()
+            .map(|path| try_read_manifest(file_io, path))
+            .collect();
+        let results = try_join_all(manifest_futures).await?;
+
+        // Store results in cache
+        let mut cache = manifest_cache.lock().unwrap();
+        for (path, entries) in uncached_paths.into_iter().zip(results) {
```
QuakeWang: Manifest entries carry `FileKind::Add/Delete`, and after overwrite/compaction the same data file can be present as an ADD from a historical base manifest and as a DELETE in the current delta manifest. Java's orphan cleaner uses a used-file-name set, while scan semantics first merge ADD/DELETE entries before reading live files; this code does neither and simply adds entry count and file size. Please define whether this function reports cleanup-protected referenced files or live readable files, then aggregate from a de-duplicated file set or from merged ADD/DELETE entries instead of raw entries.

leaves12138 (changes requested):
Thanks for adding these storage-inspection table functions. I reviewed the latest head (`d6c3ca0`) and found one correctness issue that should be fixed before merging.

**Blocking: `referenced_files_size` does not include tags under non-main branches.**

`collect_referenced_files_summary` merges tags only for the main table path (`collect_tag_files(file_io, &sm, &tm, ...)`), but for each branch it only calls `collect_scope_files` on the branch `SnapshotManager`. A branch can contain tag files under `branch/branch-<name>/tag/`, and those tag snapshots still protect their referenced manifest/data/index files from cleanup. Java `OrphanFilesClean.safelyGetAllSnapshots(branch)` handles this by switching to each branch and collecting snapshots, tagged snapshots, and changelogs for that branch.

I reproduced this locally with a temporary memory-table test that creates branch `b1` with only `branch/branch-b1/tag/tag-v1` referencing a manifest list/manifest, and no branch snapshot file. `collect_referenced_files_summary` returned `branch:b1` with all zero counts, so branch-tag-only referenced files were missed.

Please merge branch tag files into each branch scope, e.g. by using `TagManager::with_branch(branch_name)` together with `sm.with_branch(branch_name)`, and add a regression test for a branch whose referenced files are only reachable through a branch tag.

Validation I ran locally:
- `cargo clippy -p paimon -p paimon-datafusion --all-targets -- -D warnings`: passed.
- `cargo test -p paimon referenced_files -- --nocapture`: passed.
- A targeted temporary branch-tag repro failed as described above.
- Full workspace clippy is blocked in my local environment by Python 3.6 vs `pyo3` `abi3-py310`; `cargo test -p paimon-datafusion` also has unrelated fixture failures for missing `partitioned_log_table` tables.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,592 @@
...
+            .iter()
+            .map(|path| try_read_manifest(file_io, path))
+            .collect();
+        let results = try_join_all(manifest_futures).await?;
+
+        let mut cache = manifest_cache.lock().unwrap();
+        for (path, entries) in uncached_paths.into_iter().zip(results) {
+            let file_entries: Vec<(String, i64)> = entries
+                .iter()
+                .map(|e| (e.file().file_name.clone(), e.file().file_size))
```
QuakeWang: The ADD/DELETE semantics look clearer now, but this still only adds `file_name` to the referenced data-file set.

Should we also include `e.file().extra_files` here? Java orphan cleanup treats extra files as used files too, and otherwise file-index / blob / lookup-related files can appear as orphan delta when comparing with `physical_files_size`.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,681 @@
...
+            .map(|path| try_read_manifest(file_io, path))
+            .collect();
+        let results = try_join_all(manifest_futures).await?;
+
+        let mut cache = manifest_cache.lock().unwrap();
+        for (path, entries) in uncached_paths.into_iter().zip(results) {
+            let mut file_entries: Vec<(String, i64)> = Vec::new();
+            for e in &entries {
+                file_entries.push((e.file().file_name.clone(), e.file().file_size));
+                for extra in &e.file().extra_files {
```
QuakeWang, line 431: `extra_files` are included now, but counting them as `0` bytes still undercounts `data_file_size`.

These files are protected physical files too, and `referenced_files_size` is meant to be compared with `physical_files_size` for orphan analysis. Can we stat the aligned extra-file paths here and use their real sizes instead?

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,681 @@
...
+            .time_millis(2000)
+            .build();
+        branch_tm.create("v1", &branch_snapshot).await.unwrap();
+
+        let result = collect_referenced_files_summary(&file_io, table_path)
+            .await
+            .unwrap();
+
+        // Should have: total, branch:main, branch:b1
+        assert_eq!(result.len(), 3);
```
QuakeWang, line 895: This test does not really prove that branch-tag files are collected.

`branch:b1` is listed from the branch directory even without merging branch tags, and the tag's manifest lists do not exist, so the counts stay zero. A stronger regression test would create readable manifest-list / manifest files reachable only through the branch tag and assert a non-zero count or size for `branch:b1`.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,681 @@
...
+    let branch_names = bm.list_all().await?;
+
+    let branch_futures: Vec<_> = branch_names
+        .iter()
+        .map(|branch_name| {
+            let branch_sm = sm.with_branch(branch_name);
+            let branch_tm = tm.with_branch(branch_name);
+            async move {
+                let (mut branch_files, branch_tag_files) = tokio::try_join!(
+                    collect_scope_files(file_io, &branch_sm, manifest_cache_ref),
```
QuakeWang: Branch snapshots are collected now, but their manifest files are still resolved through the branch `SnapshotManager`.

`branch_sm` points to `table/branch/branch-<name>`, so `manifest_path()` looks under `branch/branch-<name>/manifest`. Paimon branches only keep branch-local snapshot/tag/schema files there; the manifest/data/index files remain under the table root. Because missing manifest lists are treated as empty, this can silently report zero referenced files for branch snapshots or branch tags.

We should read branch snapshot/tag files from the branch path, but resolve manifest paths from the table root.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,897 @@
...
+        let mut all_file_entries: Vec<Vec<(String, i64)>> = Vec::with_capacity(results.len());
+
+        for (manifest_idx, entries) in results.iter().enumerate() {
+            let mut file_entries: Vec<(String, i64)> = Vec::new();
+            for e in entries {
+                file_entries.push((e.file().file_name.clone(), e.file().file_size));
+                for extra in &e.file().extra_files {
+                    let entry_idx = file_entries.len();
+                    let full_path =
+                        extra_resolver.resolve_extra_file_path(e.partition(), e.bucket(), extra);
```
QuakeWang: `extra_files` are stat-ed now, but this path still only comes from the table partition/bucket directory.

For data files with `external_path`, Java aligns extra files with the main data file's actual directory (`DataFilePathFactory.toAlignedPath` uses `externalPathDir()` when present). With the current resolver, those extra files are looked up from the table root path instead, and missing files are silently counted as `0`.

This should resolve extra files from the main data file's directory.

crates/paimon/src/table/referenced_files.rs
```diff
@@ -0,0 +1,897 @@
...
+    pub data_file_size: i64,
+    pub index_file_count: i64,
+    pub index_file_size: i64,
+}
+
+/// Categorize a file name into a file type.
+fn classify_file_name(file_name: &str) -> FileType {
+    if file_name.starts_with("manifest-") || file_name.starts_with("index-manifest-") {
+        FileType::Manifest
+    } else if file_name.starts_with("data-") {
```
QuakeWang: This assumes physical data files always use the default `data-` prefix.

Paimon supports `data-file.prefix`, and Java's `DataFilePathFactory` uses that configured prefix when creating data file names. For a table with a custom prefix, `referenced_files_size` can still count the files from manifests, but `physical_files_size` will classify the same files as `Other`, so the comparison becomes misleading.

docs/src/sql.md
```diff
@@ -537,6 +537,105 @@ SELECT * FROM full_text_search('paimon.my_db.docs', 'content', 'paimon search',
 
 The function searches across all Tantivy full-text index files for the target column, merges results by relevance score, and returns the top-k matching rows. If no matching index is found, an empty result is returned.
 
+## Referenced Files Size
+
+The `referenced_files_size` table-valued function computes aggregated file size summaries for all snapshots referenced by a table, including snapshots from the main branch, tags, and other branches. This is useful for understanding storage usage and for orphan file cleanup.
```
QuakeWang: The orphan-cleanup wording is a little broader than the current implementation.

Java's orphan cleaner also protects `snapshot.statistics()`, while this PR reports only manifest/data/index files. If statistics files are intentionally out of scope here, the doc should say this is a manifest/data/index summary rather than implying full orphan-cleanup coverage.

crates/paimon/src/table/referenced_files.rs
JingsongLi: All files except Manifest and Index are classified as Data

QuakeWang (approved):
+1

leaves12138 (approved):
Thanks for the update. The branch-tag issue I raised earlier is fixed in `c2fec37`.

I rechecked `collect_referenced_files_summary`: branch scopes now merge both branch snapshots and `TagManager::with_branch(branch_name)` results, while resolving manifests from the table root. The added `test_branch_tag_referenced_files` covers the branch-tag-only case that previously returned zero counts.

Validation I ran locally:
- `cargo clippy -p paimon -p paimon-datafusion --all-targets -- -D warnings`: passed.
- `cargo test -p paimon referenced_files -- --nocapture`: passed, including the new branch-tag regression test.

GitHub checks are also green on the latest head. Looks good to me.

PR #324, opened 2026-05-18
feat(datafusion): auto-register built-in table functions on catalog registration

shyjsarah:
### Purpose

Linked issue: close #xxx

The `vector_search` and `full_text_search` table-valued functions could only be made available by manually calling their `register_*` function on a `SessionContext`. `SQLContext` users (including the Python binding / pypaimon) had no way to reach `register_udtf`, so these functions were effectively unusable through `SQLContext`.

This registers them automatically, in Rust, when a catalog is registered, so no caller-side setup is needed.

### Brief change log

- `SQLContext::register_catalog` now registers `vector_search` and `full_text_search` against the catalog being registered. Any `SQLContext` with a catalog gets them with no extra call.
- `full_text_search` registration is `#[cfg(feature = "fulltext")]`-gated, so builds without the feature are unaffected.
- The Python binding (`bindings/python/Cargo.toml`) enables the `fulltext` feature on `paimon-datafusion` so `full_text_search` is compiled into the binding.

Note: `referenced_files_size` / `physical_files_size` are intentionally not covered — they are system tables (`table$referenced_files_size`), not UDTFs, so they need no registration.

### Tests

`bindings/python/tests/test_datafusion.py` — `test_table_functions_registered_with_catalog`: after `register_catalog`, calling `vector_search` / `full_text_search` with the wrong argument count surfaces each function's own validation error, proving it is registered (an unregistered name would instead fail with "table function
not found").

### API and Format

- Behavior change: `SQLContext::register_catalog` now also registers the `vector_search` and `full_text_search` table functions. No new public API, no signature change.
- Build: the Python binding enables `paimon-datafusion/fulltext` (adds the pure-Rust `tantivy` dependency).
- No storage format change.

### Documentation

`docs/src/sql.md` is updated: the Vector Search / Full-Text Search registration sections now state that a `SQLContext` registers these functions automatically when a catalog is registered, and that the explicit `register_*` call is only needed with a raw `SessionContext`.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -136,6 +136,25 @@ impl SQLContext {
                 self.dynamic_options.clone(),
             )),
         );
+        // Register the built-in table-valued functions against this catalog so
+        // they are usable in SQL without any extra registration call.
+        crate::vector_search::register_vector_search(&self.ctx, catalog.clone(), default_db);
+        #[cfg(feature = "fulltext")]
+        crate::full_text_search::register_full_text_search(
+            &self.ctx,
+            catalog.clone(),
+            default_db,
+        );
+        crate::referenced_files_size::register_referenced_files_size(
+            &self.ctx,
+            catalog.clone(),
+            default_db,
+        );
+        crate::physical_files_size::register_physical_files_size(
+            &self.ctx,
+            catalog.clone(),
+            default_db,
+        );
         self.catalogs.insert(catalog_name.clone(), catalog);
         if is_first {
             self.set_current_catalog(catalog_name).await?;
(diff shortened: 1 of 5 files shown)


JingsongLi (changes requested):
We should register it in Catalog by default in Rust. This is a legacy work from before. Can you modify it?

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -136,6 +136,11 @@ impl SQLContext {
                 self.dynamic_options.clone(),
             )),
         );
+        // Register the built-in table-valued functions against this catalog so
+        // they are usable in SQL without any extra registration call.
+        crate::vector_search::register_vector_search(&self.ctx, catalog.clone(), default_db);
```
JingsongLi: create a func for table function registry.

shyjsarah: updated

JingsongLi (approved):
+1

PR #325, opened 2026-05-19
refactor(datafusion): convert files_size from table functions to system tables

JingsongLi:
### Purpose

These are table metadata and fit better as system tables (table$physical_files_size) rather than UDTFs (physical_files_size('table')), consistent with $partitions, $snapshots, etc.

### DIFF
--- a/crates/integrations/datafusion/src/lib.rs
+++ b/crates/integrations/datafusion/src/lib.rs
@@ -43,10 +43,8 @@ mod filter_pushdown;
 #[cfg(feature = "fulltext")]
 mod full_text_search;
 mod merge_into;
-mod physical_files_size;
 mod physical_plan;
 mod procedures;
-mod referenced_files_size;
 mod relation_planner;
 pub mod runtime;
 mod sql_context;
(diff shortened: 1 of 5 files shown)


leaves12138 (approved):
Looks good to me.

This refactor fits the existing DataFusion system table model: the two file-size providers now hang off `<table>$physical_files_size` and `<table>$referenced_files_size`, are registered through the shared `system_tables` registry, and the old table-function argument/catalog lookup layer is removed cleanly. The schemas and collection paths remain equivalent to the previous UDTF providers.

I checked the diff, verified the GitHub CI checks are green for this head commit, and locally ran:

- `cargo test -p paimon-datafusion system_tables`
- `cargo test -p paimon-datafusion referenced_files_size`
- `cargo test -p paimon-datafusion physical_files_size`
- `cargo test -p paimon-datafusion --test system_tables -- --nocapture`

All passed. A future follow-up could add dedicated integration assertions for the two new `$...files_size` system tables, but I do not think that needs to block this refactor.

PR #334, opened 2026-05-19
fix: validate filesystem catalog identifiers

### LINKED ISSUE #333: Filesystem catalog should reject path-control characters in identifiers
## Bug

`FileSystemCatalog` builds database and table paths from raw database and object names. Names containing path-control segments such as `..`, `/`, or `\` can make catalog operations derive paths outside the intended warehouse namespace on local filesystem storage.

## Impact

Catalog operations such as create/drop/rename database or table can run filesystem operations on paths derived from unsafe identifier input.

## Reproduction

Using a local filesystem warehouse:

1. Create a filesystem catalog with a warehouse under a temporary directory.
2. Call `create_database("../escaped", false, HashMap::new())`.
3. Observe that `escaped.db` is created next to the warehouse instead of under it.

A table object name such as `../../table_escape` can similarly escape the database directory.

## Expected behavior

Filesystem catalog identifiers should reject path-control names before path construction. At minimum, empty/whitespace-only names, `.`, `..`, path separators, and control characters should be rejected with `IdentifierInvalid`.

fallintoplace:
## What changed

This adds path-safe identifier validation helpers and enforces them in `FileSystemCatalog` before database or table paths are constructed.

The existing `Identifier::new` API remains infallible and compatible. `FileSystemCatalog` now rejects database and object names that are empty or whitespace-only, `.`, `..`, contain `/` or `\`, or contain control characters before using them in `database_path` or `table_path`.

C/Go binding identifier constructors remain compatible and do not reject path-control names at identifier construction time. Unsafe names are rejected when they are used with `FileSystemCatalog` operations.

## Why

Filesystem catalog paths are derived from database and table names. Without validation, path-control input can escape the intended warehouse namespace on local filesystem storage.

Fixes #333.

## Validation

- `cargo test -p paimon catalog::`
- `git diff --check`

### DIFF
--- a/crates/paimon/src/catalog/mod.rs
+++ b/crates/paimon/src/catalog/mod.rs
@@ -69,29 +69,27 @@ pub struct Identifier {
 }
 
 impl Identifier {
-    /// Create an identifier from database and object name, validating both names.
-    pub fn new(database: impl Into<String>, object: impl Into<String>) -> Result<Self> {
-        let identifier = Self {
+    /// Create an identifier from database and object name.
+    pub fn new(database: impl Into<String>, object: impl Into<String>) -> Self {
+        Self {
             database: database.into(),
             object: object.into(),
-        };
-        identifier.validate()?;
-        Ok(identifier)
+        }
     }
 
     /// Validate this identifier's database and object names.
-    pub fn validate(&self) -> Result<()> {
+    pub(crate) fn validate(&self) -> Result<()> {
         Self::validate_database_name(&self.database)?;
         Self::validate_object_name(&self.object)
     }
 
     /// Validate a database name for path-safe catalog use.
-    pub fn validate_database_name(name: &str) -> Result<()> {
+    pub(crate) fn validate_database_name(name: &str) -> Result<()> {
         validate_identifier_name("database", name)
     }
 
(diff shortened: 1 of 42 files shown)


crates/paimon/src/catalog/mod.rs
```diff
@@ -76,6 +77,29 @@ impl Identifier {
...
 
+    /// Create an identifier from database and object name, validating both names.
+    pub fn try_new(database: impl Into<String>, object: impl Into<String>) -> Result<Self> {
+        let identifier = Self::new(database, object);
+        identifier.validate()?;
+        Ok(identifier)
+    }
+
+    /// Validate this identifier's database and object names.
+    pub fn validate(&self) -> Result<()> {
```
JingsongLi: Why not just in `Identifier.new`?

fallintoplace: Good point. I will implement this.

JingsongLi:
After the signature modification, this is indeed a major change. Perhaps we should still keep Identifier:: new() invincible (without changing the signature) and only perform validation in the entry method of the FileSystemCatalog. (RESTCatalog will have validation in server side.)

JingsongLi (approved):
+1

PR #340, opened 2026-05-21
feat: support merge-engine=aggregation (basic mode)

TheR1sing3un:
### Purpose

Add `merge-engine=aggregation` (basic mode) — Rust port of Java `AggregateMergeFunction`.

### Brief change log

- `MergeEngine::Aggregation` + `AggregationConfig` (basic-mode option whitelist; rejects `ignore-retract`, `distinct`, `sequence-group`, `nested-key`, `count-limit`, `aggregation.remove-record-on-delete`, `ignore-delete`).
- 12 aggregators: `sum` / `product` / `min` / `max` / `count` / `last_value` / `first_value` / `last_non_null_value` / `first_non_null_value` / `bool_and` / `bool_or` / `listagg`.
- `AggregateMergeFunction` wired through `KeyValueFileReader`. Sequence fields forced to `last_value`; PK columns copied through.
- `DELETE` / `UPDATE_BEFORE` rejected at runtime. `DECIMAL` `product` deferred (BigDecimal-style scale rebasing TBD).

### Tests

- 53 unit tests under `aggregator/*` and `sort_merge`.
- 9 DataFusion e2e tests in `pk_tables.rs` covering sum+listagg, default function, mixed aggregators, sequence-field forced `last_value`, DELETE rejection, missing-config error, unsupported-options rejection, all-NULL sum, and read-path routing.
- `cargo fmt`, `cargo clippy --features fulltext,vortex -- -D warnings`, `cargo test -p paimon --features fulltext,vortex --lib`, `cargo test -p paimon-datafusion --test pk_tables` all green.

### API and Format

No on-disk format change. All new types are `pub(crate)`.

### Documentation

Inline module docs reference the Java counterparts. No user-facing doc update yet.

### DIFF
--- a/crates/paimon/src/spec/aggregation.rs
+++ b/crates/paimon/src/spec/aggregation.rs
@@ -27,6 +27,7 @@ const AGGREGATION_REMOVE_RECORD_ON_DELETE_OPTION: &str = "aggregation.remove-rec
 const FIELDS_DEFAULT_AGG_FUNCTION_OPTION: &str = "fields.default-aggregate-function";
 const FIELDS_PREFIX: &str = "fields.";
 const AGG_FUNCTION_SUFFIX: &str = ".aggregate-function";
+const LIST_AGG_DELIMITER_SUFFIX: &str = ".list-agg-delimiter";
 const IGNORE_RETRACT_SUFFIX: &str = ".ignore-retract";
 const DISTINCT_SUFFIX: &str = ".distinct";
 const SEQUENCE_GROUP_SUFFIX: &str = ".sequence-group";
@@ -93,7 +94,7 @@ impl<'a> AggregationConfig<'a> {
             }
         };
         if mode.is_some() {
-            self.validate_field_aggregators(fields)?;
+            self.validate_field_scoped_options(fields)?;
         }
         Ok(mode)
     }
@@ -156,17 +157,22 @@ impl<'a> AggregationConfig<'a> {
     }
 
     /// Schema-aware checks run by [`validate_create_mode`] once the engine is
-    /// confirmed active.  For every `fields.<col>.aggregate-function` key:
-    /// * the `<col>` segment must name an existing schema field
+    /// confirmed active.  For every `fields.<col>.<known-suffix>` key
+    /// (currently `aggregate-function` and `list-agg-delimiter`):
+    /// * the `<col>` segment must name an existing schema field; this catches
+    ///   typo'd column names that would otherwise silently fall back to the
+    ///   default function / default delimiter at read time.
+    ///
+    /// For `aggregate-function` keys additionally:
     /// * the function name must be one of the supported aggregators
     /// * the function must accept the field's declared data type
     ///
     /// `fields.default-aggregate-function` only has its name validated;
     /// per-column type compatibility for the default is deferred to runtime
     /// because the default applies broadly across columns.
-    fn validate_field_aggregators(&self, fields: &[DataField]) -> crate::Result<()> {
+    fn validate_field_scoped_options(&self, fields: &[DataField]) -> crate::Result<()> {
         for (key, value) in self.options {
-            let Some(col) = parse_field_aggregator_key(key) else {
+            let Some((col, kind)) = parse_field_scoped_option_key(key) else {
                 continue;
             };
             let Some(field) = fields.iter().find(|f| f.name() == col) else {
@@ -180,7 +186,9 @@ impl<'a> AggregationConfig<'a> {
                     ),
                 });
             };
-            validate_aggregator_for_type(value, col, field.data_type())?;
+            if matches!(kind, FieldScopedOptionKind::AggregateFunction) {
+                validate_aggregator_for_type(value, col, field.data_type())?;
+            }
         }
 
         if let Some(default) = self
@@ -202,17 +210,39 @@ impl<'a> AggregationConfig<'a> {
     }
 }
 
-/// Parse the `<col>` segment out of `fields.<col>.aggregate-function`, or
-/// return `None` if `key` is some other option string.
-fn parse_field_aggregator_key(key: &str) -> Option<&str> {
+/// Field-scoped option suffixes that schema-aware validation recognizes.
+/// Each variant maps to a single `fields.<col>.<suffix>` key shape.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FieldScopedOptionKind {
+    AggregateFunction,
+    ListAggDelimiter,
+}
+
+/// Parse the `<col>` segment and option kind out of a
+/// `fields.<col>.<known-suffix>` key, or return `None` if `key` doesn't
+/// match any known field-scoped option suffix.
+fn parse_field_scoped_option_key(key: &str) -> Option<(&str, FieldScopedOptionKind)> {
     let inner = key.strip_prefix(FIELDS_PREFIX)?;
-    let col = inner.strip_suffix(AGG_FUNCTION_SUFFIX)?;
-    if col.is_empty() {
-        // `fields..aggregate-function` is malformed; reject by treating as
-        // "no match" so the caller surfaces a typo-style error elsewhere.
-        return None;
+    for (suffix, kind) in [
+        (
+            AGG_FUNCTION_SUFFIX,
+            FieldScopedOptionKind::AggregateFunction,
+        ),
+        (
+            LIST_AGG_DELIMITER_SUFFIX,
+            FieldScopedOptionKind::ListAggDelimiter,
+        ),
+    ] {
+        if let Some(col) = inner.strip_suffix(suffix) {
+            if col.is_empty() {
+                // `fields..<suffix>` is malformed; treat as "no match" so the
+                // caller surfaces a typo-style error elsewhere.
+                continue;
+            }
+            return Some((col, kind));
+        }
     }
-    Some(col)
+    None
 }
 
 const SUPPORTED_AGGREGATOR_NAMES_HINT: &str = "supported: sum, product, min, max, last_value, \
@@ -452,6 +482,26 @@ mod tests {
                     && message.contains("fields.amout.aggregate-function")
                     && message.contains("amount")),
             "expected unknown-field error to surface the typo + available columns, got {err:?}"
+        );
+    }
+
+    #[test]
+    fn test_validate_create_mode_rejects_unknown_field_for_list_agg_delimiter() {
+        // typo: `tga` instead of `tag`; without this check `listagg` on `tag`
+        // would silently fall back to the default delimiter at read time.
+        let options = aggregation_options(&[
+            ("fields.tag.aggregate-function", "listagg"),
+            ("fields.tga.list-agg-delimiter", "|"),
+        ]);
+        let err = AggregationConfig::new(&options)
+            .validate_create_mode(true, &sample_fields())
+            .unwrap_err();
+        assert!(
+            matches!(err, crate::Error::ConfigInvalid { ref message }
+                if message.contains("tga")
+                    && message.contains("fields.tga.list-agg-delimiter")
+                    && message.contains("tag")),
+            "expected unknown-field error for list-agg-delimiter typo, got {err:?}"
         );
     }
 
--- a/crates/paimon/src/spec/aggregation.rs
+++ b/crates/paimon/src/spec/aggregation.rs
@@ -17,7 +17,7 @@
 
 use std::collections::HashMap;
 
-use crate::spec::{CoreOptions, DataField, DataType};
+use crate::spec::{CoreOptions, DataField, DataType, VarCharType};
 
 const MERGE_ENGINE_OPTION: &str = "merge-engine";
 const AGGREGATION_ENGINE: &str = "aggregation";
@@ -339,7 +339,9 @@ pub(crate) fn validate_aggregator_for_type(
                 | DataType::VarChar(_)
         ),
         "bool_and" | "bool_or" => matches!(dt, DataType::Boolean(_)),
-        "listagg" => matches!(dt, DataType::Char(_) | DataType::VarChar(_)),
+        // Java `FieldListaggAggFactory` only accepts unbounded VARCHAR (STRING);
+        // CHAR and bounded VARCHAR(n) are rejected.
+        "listagg" => matches!(dt, DataType::VarChar(v) if v.length() == VarCharType::MAX_LENGTH),
         "last_value" | "first_value" | "last_non_null_value" | "first_non_null_value" => true,
         _ => {
             return Err(crate::Error::ConfigInvalid {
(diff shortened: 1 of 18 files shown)


JingsongLi (approved):
This is a substantial and well-structured PR. The aggregation engine implementation is clean and well-tested.

Highlights:

1. **Architecture**: The `FieldAggregator` trait with per-field accumulator state is the right abstraction. The `Mutex`-wrapped aggregator vector in `AggregateMergeFunction` is a pragmatic choice given the existing `MergeFunction` trait signature (shared `&self`), and uncontended in practice.

2. **Correctness safeguards**:
   - Overflow detection via `checked_add`/`checked_mul` for integer types rather than silent wrapping.
   - NaN handling in min/max mirrors Java behavior.
   - Non-nullable field NULL check post-merge catches configuration errors early.
   - Retract row rejection up-front prevents partial accumulation leaking into output.

3. **Option whitelist approach** (basic mode): Explicitly rejecting unsupported options (`ignore-retract`, `distinct`, `sequence-group`, etc.) at both create-time and runtime is the right strategy — fail fast rather than silently ignore advanced features.

4. **Test coverage**: 53 unit tests + 9 e2e tests covering sum, listagg, mixed aggregators, default function, sequence-field forcing, DELETE rejection, missing config errors, all-NULL groups, and read-path routing verification. Very thorough.

5. The `pick_agg!` macro for `last_value`/`first_value`/`last_non_null_value`/`first_non_null_value` is elegant — stores a 1-row Arrow slice rather than typed scalars, so it works for any Paimon type without per-type dispatching.

Minor observations (non-blocking):
- The `CountAgg` semantics: it counts non-NULL inputs but the column holds `i64` placeholders. The Java reference also does this, so it's correct by design, but the behavior may surprise users who write `(1, 1)` expecting the count column to be the literal value rather than "one row was observed." A doc note in the system table docs could help.
- `DECIMAL` `product` rejection is well-documented with a clear error message pointing the user to alternatives.

Well done!

crates/paimon/src/spec/schema.rs
```diff
@@ -293,6 +294,7 @@ impl Schema {
         let fields = Self::normalize_fields(&fields, &partition_keys, &primary_keys)?;
         Self::validate_blob_fields(&fields, &partition_keys, &options)?;
         PartialUpdateConfig::new(&options).validate_create_mode(!primary_keys.is_empty())?;
+        AggregationConfig::new(&options).validate_create_mode(!primary_keys.is_empty())?;
```
JingsongLi: This validation only sees the option map plus whether the table has PKs, so it lets schema-dependent aggregation mistakes through CREATE TABLE. For example, a typo like `fields.amout.aggregate-function = 'sum'` is accepted and then the real `amount` column silently falls back to the default/`last_non_null_value`; an unknown function name or an incompatible function/type pair is also accepted until the first read/write constructs `AggregateMergeFunction`. Java Paimon validates field aggregation options during schema validation (including factory lookup), so can we make this create-time check schema-aware and reject unknown fields, unknown functions, and incompatible function/type pairs before persisting the table schema?

JingsongLi (changes requested):
I found one create-time validation gap that can let invalid aggregation table metadata be committed. The implementation and e2e behavior otherwise look solid; targeted tests passed locally.

crates/paimon/src/spec/schema.rs
TheR1sing3un: done

TheR1sing3un:
@JingsongLi Hi, any issue to fix? Lets land it! Thanks!

crates/paimon/src/spec/aggregation.rs
```diff
@@ -0,0 +1,553 @@
...
+    /// * the `<col>` segment must name an existing schema field
+    /// * the function name must be one of the supported aggregators
+    /// * the function must accept the field's declared data type
+    ///
+    /// `fields.default-aggregate-function` only has its name validated;
+    /// per-column type compatibility for the default is deferred to runtime
+    /// because the default applies broadly across columns.
+    fn validate_field_aggregators(&self, fields: &[DataField]) -> crate::Result<()> {
+        for (key, value) in self.options {
+            let Some(col) = parse_field_aggregator_key(key) else {
```
JingsongLi: This validation only parses `fields.<col>.aggregate-function`, so typoed `fields.<col>.list-agg-delimiter` options still get persisted silently. For example, `fields.tga.list-agg-delimiter = "|"` on a table with column `tag` is accepted; later `listagg` looks up `fields.tag.list-agg-delimiter`, misses the typo, and falls back to `,`, producing the wrong aggregation result without any error. Since `list-agg-delimiter` is an allowed aggregation field option, it should go through the same field-name validation as aggregate-function keys.

crates/paimon/src/table/kv_file_reader.rs
```diff
@@ -328,7 +341,14 @@ impl KeyValueFileReader {
                     user_sequence_indices.clone(),
                     value_indices.clone(),
                     merge_output_schema.clone(),
-                    Self::new_merge_function(merge_engine, &table_options, &table_name)?,
+                    Self::new_merge_function(
```
JingsongLi, line 344: This creates a sort-merge reader inside the per-`DataSplit` loop, so aggregation is only applied within each split. `TableScan` can bin-pack files from the same partition/bucket into multiple splits, and the same primary key can appear in more than one split. In that case this emits one partial aggregate per split instead of one globally aggregated row. I reproduced it with an aggregation table using `source.split.target-size = 1b` and `source.split.open-file-cost = 1b`, then inserting `(1, 10)`, `(1, 20)`, `(1, 30)` in three commits: `SELECT` returns 3 rows instead of one row with `amount = 60`. Aggregation/partial-update reads need either split planning that keeps all files for a partition/bucket together, or a second merge stage across split outputs.

crates/paimon/src/spec/aggregation.rs
TheR1sing3un: Good catch — confirmed and fixed

crates/paimon/src/table/kv_file_reader.rs
TheR1sing3un: Thanks for the careful repro — confirmed the behavior. I'd like to handle this in a separate follow-up PR rather than in this one, for two reasons.

  1. It's a pre-existing bug, not something this PR introduces. Both Deduplicate and PartialUpdate already go through the same for split in &splits { SortMergeReaderBuilder::new(...) } pattern in kv_file_reader.rs, and the same per-split bin-packing in table_scan.rs (split_for_batch(data_files, target_split_size, open_file_cost) on the non-data-evolution
  path). With source.split.target-size=1b, dedup/partial-update will also miss cross-split merges whenever file key ranges overlap — aggregation just makes the symptom more visible.
  2. The fix is scoped to scan planning, not to the aggregation merge function. The minimal correctness fix is to keep all files of a (partition, bucket) in a single split for PK tables (skip split_for_batch on that path). The cleaner fix mirrors Java's MergeTreeSplitGenerator, grouping files into sections by key-range overlap. Either way, the change lives in
  table_scan.rs and affects all three PK merge engines, so it really wants its own PR + test surface.

  Plan: get this PR merged for the aggregation engine itself (which is correct under the default source.split.target-size), then open the follow-up PR immediately after. Happy to link it back to this thread. Does that work for you?

crates/paimon/src/spec/aggregation.rs
```diff
@@ -0,0 +1,603 @@
...
+                return Err(crate::Error::ConfigInvalid {
+                    message: format!(
+                        "Aggregation field '{col}' referenced by '{key}' is not declared in \
+                         the table schema; available columns: [{}]",
+                        available.join(", ")
+                    ),
+                });
+            };
+            if matches!(kind, FieldScopedOptionKind::AggregateFunction) {
+                validate_aggregator_for_type(value, col, field.data_type())?;
```
JingsongLi: [P2] Mirror runtime field priority during CREATE validation

This type-compatibility check runs for every `fields.<col>.aggregate-function`, but `AggregateMergeFunction::new` later ignores that configured function for `sequence.field` columns (forcing `last_value`) and for primary-key columns (no aggregator). That makes otherwise usable table definitions fail at CREATE TABLE time when the ignored function is incompatible with the column type, e.g. `sequence.field=amount` plus `fields.amount.aggregate-function=listagg` on an INT column, or the same option on an INT primary key. Could validation skip type compatibility for sequence/PK fields, or otherwise resolve the effective runtime aggregator before validating the pair?

TheR1sing3un: fixed

TheR1sing3un: fixed

crates/paimon/src/table/aggregator/numeric.rs
```diff
@@ -0,0 +1,960 @@
...
+}
+
+fn decimal_array(
+    precision: u8,
+    scale: i8,
+    value: Option<i128>,
+    agg_name: &str,
+    field_name: &str,
+) -> crate::Result<ArrayRef> {
+    let mut builder = Decimal128Builder::with_capacity(1)
```
QuakeWang, line 683: `with_precision_and_scale` only sets DECIMAL metadata; it does not validate that the accumulated raw value fits the declared precision. For example, `DECIMAL(3,2)` with `9.99 + 0.01` can produce raw `1000` under `Decimal128(3,2)`. Please add precision validation or match Java `DecimalUtils.add/fromBigDecimal` semantics, with a boundary test.

crates/paimon/src/spec/aggregation.rs
```diff
@@ -0,0 +1,686 @@
...
+                | DataType::Double(_)
+                | DataType::Decimal(_)
+                | DataType::Date(_)
+                | DataType::Time(_)
+                | DataType::Timestamp(_)
+                | DataType::Char(_)
+                | DataType::VarChar(_)
+        ),
+        "bool_and" | "bool_or" => matches!(dt, DataType::Boolean(_)),
+        "listagg" => matches!(dt, DataType::Char(_) | DataType::VarChar(_)),
```
QuakeWang: This accepts `CHAR` and bounded `VARCHAR(n)` for `listagg`, but Java `FieldListaggAggFactory` only accepts unbounded `VarCharType` / `STRING`. This can create Rust metadata that Java rejects. Please restrict create-time validation and the runtime constructor, and add rejection tests for `CHAR` / bounded `VARCHAR`.

JingsongLi:
I found one issue that is easy to miss when this PR is reviewed independently from #374.

In `crates/paimon/src/table/kv_file_reader.rs`, the merged batch is reordered with `RecordBatch::try_new(output_schema.clone(), columns)`. When the projection is empty, e.g. `COUNT(*)`, `columns` is empty and Arrow cannot infer the row count from the arrays, so the reordered batch can lose the row count even though the merge stream produced rows. This affects aggregation/PK KV reads because this PR routes aggregation tables through the KV merge path.

Please either include the `RecordBatchOptions::new().with_row_count(Some(batch.num_rows()))` fix here, or make this PR explicitly depend on the PR that adds it, so `COUNT(*)` / empty projection reads do not regress when this lands.

TheR1sing3un:
Thanks @JingsongLi, good catch — confirmed.

I've rebased this PR onto current `main`, which now includes #374 (the PR that adds `RecordBatchOptions::new().with_row_count(Some(batch.num_rows()))` to the reorder step in `kv_file_reader.rs`). So the empty-projection / `COUNT(*)` path no longer loses the row count when this lands — the fix flows in cleanly since this PR doesn't touch that reorder line.

To lock it in, I also added an e2e regression test `test_pk_aggregation_count_star_empty_projection` in `pk_tables.rs`: it writes overlapping primary keys across two commits (5 raw rows, 3 distinct keys) and asserts `SELECT COUNT(*)` returns 3. Without the `with_row_count` fix the empty-projection reorder would report 0 rows and the count would collapse to 0.

All green locally: `cargo fmt`, `cargo clippy -p paimon -p paimon-datafusion --features fulltext,vortex --tests -- -D warnings`, `cargo test -p paimon --features fulltext,vortex --lib` (828 passed), and `cargo test -p paimon-datafusion --test pk_tables` (49 passed).

crates/paimon/src/table/aggregator/numeric.rs
TheR1sing3un: Fixed in 21889e3. A Decimal `sum` that exceeds the declared precision now yields NULL instead of persisting an out-of-range `Decimal128` value, matching Java `DecimalUtils.add` / `Decimal.fromBigDecimal` (which return null on precision overflow rather than throwing). Added boundary tests in `numeric.rs`: `DECIMAL(3,2)` with `9.99 + 0.01` (raw 1000) -> NULL, plus an in-range `1.23 + 4.56 = 5.79` case so the precision guard doesn't false-positive on valid sums.

crates/paimon/src/spec/aggregation.rs
TheR1sing3un: Fixed in 21889e3. `listagg` now accepts only unbounded VARCHAR (STRING): both create-time validation (`validate_aggregator_for_type`) and the runtime constructor (`ListaggAgg::new`) reject `CHAR` and bounded `VARCHAR(n)`, matching Java `FieldListaggAggFactory` (`fieldType instanceof VarCharType` && `getLength() == VarCharType.MAX_LENGTH`). So we no longer create metadata Java would reject. Added rejection tests for `CHAR` and `VARCHAR(255)` plus an acceptance test for STRING, and extended the validator/constructor consistency table with an unbounded STRING type.

crates/paimon/src/table/aggregator/numeric.rs
TheR1sing3un: Nice catch, thank you, fixed

crates/paimon/src/spec/aggregation.rs
TheR1sing3un: Nice catch, thank you, fixed

crates/paimon/src/spec/schema.rs
```diff
@@ -305,6 +306,7 @@ impl Schema {
         let fields = Self::normalize_fields(&fields, &partition_keys, &primary_keys)?;
         Self::validate_blob_fields(&fields, &partition_keys, &options)?;
         PartialUpdateConfig::new(&options).validate_create_mode(!primary_keys.is_empty())?;
+        AggregationConfig::new(&options).validate_create_mode(&primary_keys, &fields)?;
```
JingsongLi, line 313: Could we run this same aggregation validation from `TableSchema::apply_changes` after applying `SetOption` / `RemoveOption`? `CREATE TABLE` now rejects typoed `fields.<col>.aggregate-function` keys, unknown function names, and function/type mismatches, but `ALTER TABLE ... SET TBLPROPERTIES` goes through `apply_changes` and currently only revalidates the first-row changelog constraint. That can still persist invalid aggregation metadata after table creation; for example `fields.amout.aggregate-function = sum` would be accepted by ALTER and then silently fall back to `last_non_null_value` at read time. Reusing `AggregationConfig::validate_create_mode(&new_schema.primary_keys, &new_schema.fields)` before saving the altered schema would close that gap.

TheR1sing3un: fixed

JingsongLi (approved):
+1

PR #342, opened 2026-05-22
fix: validate changelog producer for first-row merge engine

QuakeWang:
### Purpose

Prevent invalid `merge-engine=first-row` tables from using unsupported `changelog-producer` values. `first-row` only supports `none` and `lookup`.

### Brief change log

 - Add shared validation for `first-row` changelog producer compatibility.
 - Reject invalid options during schema creation and schema changes.
 - Add a runtime guard in table write setup for loaded or legacy schemas.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -106,6 +106,13 @@ impl ChangelogProducer {
     }
 }
 
+pub(crate) fn first_row_supports_changelog_producer(producer: ChangelogProducer) -> bool {
+    matches!(
+        producer,
+        ChangelogProducer::None | ChangelogProducer::Lookup
+    )
+}
+
 /// Format the bucket directory name for a given bucket number.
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
LGTM! Clean implementation with good test coverage.

A few minor observations:

1. The validation correctly covers both the `Schema::build()` path (create-time) and the `TableSchema::apply_changes()` path (alter-time), as well as a runtime guard in `TableWrite::new()` for legacy schemas loaded from disk — nice defense in depth.

2. The `first_row_supports_changelog_producer` helper is `pub(crate)` and shared between `core_options.rs` and `table_write.rs`, keeping the logic in one place.

3. Tests cover all the key scenarios: compatible producers (none, lookup), incompatible producers (input, full-compaction), alter-time rejection, and final-options validation when both merge-engine and changelog-producer change together.

One optional thought for a follow-up: if the `MERGE_ENGINE_OPTION` constant in `schema.rs` could be unified with the one already defined in `core_options.rs`, it would avoid having two identical string literals — but not a blocker for this PR.

PR #344, opened 2026-05-26
feat: support canonical Lumina index type

QuakeWang:
### Purpose

Rust vector search previously only recognized Lumina index manifest entries with `index_type = "lumina-vector-ann"`. Java now uses `index_type = "lumina"` as the canonical Lumina identifier while keeping `lumina-vector-ann` only for legacy compatibility.

This PR adds the canonical Lumina identifier and a shared helper for Lumina index type detection. `VectorSearchBuilder` now accepts both `lumina` and `lumina-vector-ann` manifest entries, while preserving existing behavior for non-Lumina indexes, delete entries, and field mismatches.

### DIFF
--- a/crates/paimon/src/lumina/mod.rs
+++ b/crates/paimon/src/lumina/mod.rs
@@ -20,7 +20,16 @@ pub mod reader;
 
 use std::collections::HashMap;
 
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #346, opened 2026-05-29
feature: add more object storage backends

liujiwen-up:
# Purpose
Linked issue: close #xxx

Add OpenDAL-backed support for more object storage backends in Paimon Rust, covering COS, Azure Data Lake Storage Gen2, OBS, and GCS.

# Brief change log
Added storage feature flags for storage-cos, storage-azdls, storage-obs, and storage-gcs.
Added storage-all coverage for the new backends.
Implemented backend config parsing and OpenDAL operator construction for:
- Tencent Cloud COS
- Azure Data Lake Storage Gen2
- Huawei Cloud OBS
- Google Cloud Storage
Added scheme aliases such as cosn, gs, abfs, and abfss.
Added object storage FileIO path tests to verify scheme detection and relative path handling.
Extended GCS config alias support for Hadoop/OpenDAL-style option keys.
Updated docs to list and show usage examples for the new storage backends.
# Tests
```
cargo check -p paimon
cargo check -p paimon --features storage-all
cargo test -p paimon --features storage-all io::object_storage_path_test --no-fail-fast
cargo test -p paimon --features storage-all io::storage_gcs --no-fail-fast
cargo test -p paimon --features storage-all io:: --no-fail-fast
cargo test -p paimon --features storage-all --no-fail-fast
```
# API and Format
This change adds optional Cargo feature flags and storage backend support. It does not change the Paimon table format or existing storage format.

# Documentation
Updated documentation to include the newly supported storage backends and basic configuration examples.

PR #353, opened 2026-06-02
build: pin unicode-segmentation for vortex tests

JingsongLi:
## Summary

Main CI is still hanging after reverting #346, so the object-storage change does not appear to be the root cause.

Comparing the successful #350 Ubuntu unit job with the hanging main Ubuntu unit job shows the same compiled crate set except for one resolved patch dependency:

- successful #350 PR merge-ref: `unicode-segmentation v1.13.2`
- hanging main run: `unicode-segmentation v1.13.3`

This PR pins `unicode-segmentation` to `=1.13.2` through the `paimon` crate so CI resolves back to the dependency set that passed before the Vortex tests started hanging.

## Changes

- Add an exact `unicode-segmentation = "=1.13.2"` dependency in `crates/paimon/Cargo.toml`.
- Leave the #346 object-storage changes intact.

## Testing

- [x] `cargo tree -p paimon --features fulltext,vortex -i unicode-segmentation` resolves `unicode-segmentation v1.13.2`
- [x] `cargo test -p paimon arrow::format::vortex::tests::test_vortex_read_with --all-targets --features fulltext,vortex -- --nocapture`
- [x] `cargo test -p paimon-datafusion --features vortex --test vortex_tables -- --nocapture`

## Notes

This is intended as a targeted unblock while the interaction between the latest `unicode-segmentation` patch and the Linux Vortex tests is investigated further.

### DIFF
--- a/crates/paimon/Cargo.toml
+++ b/crates/paimon/Cargo.toml
@@ -104,6 +104,9 @@ tempfile = { version = "3", optional = true }
 vortex = { version = "0.68", features = ["tokio"], optional = true }
 kanal = { version = "0.1.1", optional = true }
 libloading = "0.8"
+# Keep CI on the dependency set that passed before unicode-segmentation 1.13.3.
+# The 1.13.3 resolver update correlates with Linux Vortex tests hanging.
+unicode-segmentation = "=1.13.2"
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


leaves12138 (approved):
LGTM. This is a small dependency pin, and all CI jobs are green. Pinning `unicode-segmentation` to the last known-good version for the Vortex test hang is reasonable.

PR #354, opened 2026-06-02
test: add mixed-format schema evolution coverage

QuakeWang:
### Purpose

Linked issue: #258

Rust reader already had coverage for Java/Spark-generated full type boundary values, but schema evolution compatibility across mixed data file formats was not explicitly covered. In particular, existing fixtures did not verify that Parquet/ORC/Avro files written under different schema versions are read consistently by field id, with null-fill for added columns and casts for promoted types.

### Brief change log

  - Add `format_schema_evolution_add_column` Spark fixture with Parquet old-schema files and ORC/Avro new-schema files.
  - Add `format_schema_evolution_type_promotion` Spark fixture with Parquet INT files and ORC/Avro BIGINT files.
  - Add Rust integration tests that assert:
    - planned files include Parquet, ORC, and Avro
    - planned files include multiple schema ids
    - reader output preserves null-fill and INT to BIGINT cast semantics

### Tests

- `cargo test -p paimon-integration-tests test_read_format_schema_evolution -- --nocapture`

Issue #258, opened 2026-04-17
Add more consistency file tests from Java data files


JingsongLi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Description

more consistency file tests

### Willingness to contribute

- [ ] I'm willing to submit a PR!

liujiwen-up:
I’d like to work on this.

PR #366, opened 2026-06-05
test: cover mixed-format reorder column reads

liujiwen-up:
### Purpose

Linked issue: #258

Add focused Spark-provisioned integration coverage for mixed-format schema evolution reads after `ALTER COLUMN ... FIRST/AFTER`.

This PR intentionally covers only the reorder/move-column case to avoid overlapping with #354, which covers mixed-format add-column and type-promotion reads.

### Brief change log

- Add a Spark fixture with Parquet, ORC, and Avro files written across column reorder/move schema versions.
- Add an integration test verifying full reads expose the current table schema order.
- Verify mixed-format reads map values by field id rather than physical column position.
- Verify projected reads preserve caller-specified projection order after column reorder/move.

### Tests

- `cargo fmt`: passed
- `git diff --check`: passed
- `cargo test -p paimon-integration-tests --test read_tables --no-run`: passed

### API and Format

No API or storage format changes. This PR only adds Spark provisioning and Rust integration test coverage.

### Documentation

### DIFF
--- a/crates/integration_tests/tests/read_tables.rs
+++ b/crates/integration_tests/tests/read_tables.rs
@@ -1508,16 +1508,10 @@ async fn test_read_schema_evolution_rename_column() {
     let (plan, batches) =
         scan_and_read_with_fs_catalog("schema_evolution_rename_column", None).await;
 
-    let formats: HashSet<&str> = plan
-        .splits()
-        .iter()
-        .flat_map(|split| split.data_files())
-        .filter_map(|file| file.file_name.rsplit_once('.').map(|(_, ext)| ext))
-        .collect();
-    assert_eq!(
-        formats,
-        HashSet::from(["avro", "orc", "parquet"]),
-        "schema_evolution_rename_column should scan all provisioned file formats"
+    assert_plan_file_formats(
+        &plan,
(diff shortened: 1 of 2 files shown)


JingsongLi:
Please rebase main.

crates/integration_tests/tests/read_tables.rs
```diff
@@ -1675,6 +1669,109 @@ async fn test_read_mixed_format_schema_evolution_drop_column() {
...
+async fn test_read_mixed_format_schema_evolution_reorder_move_column() {
+    let (plan, batches) =
+        scan_and_read_with_fs_catalog("mixed_format_schema_evolution_reorder_move_column", None)
+            .await;
+
+    assert_plan_file_formats(
+        &plan,
+        &["avro", "orc", "parquet"],
+        "mixed_format_schema_evolution_reorder_move_column",
+    );
```
QuakeWang, line 1684: Non-blocking: can we also assert multiple schema ids here, like the adjacent mixed-format schema evolution tests? This would make the test prove it is exercising the historical-schema path, not only mixed file formats.

JingsongLi (approved):
+1

PR #370, opened 2026-06-10
feat(catalog): support column-level alter table

### LINKED ISSUE #368: Support column-level alter table in FileSystemCatalog and RESTCatalog
### Problem

`TableSchema::apply_changes` currently only handles `SetOption` / `RemoveOption`; all column-level `SchemaChange`s (add/rename/drop column, update column type/nullability/comment/position, update comment) return `Unsupported`. As a result `FileSystemCatalog::alter_table` cannot evolve a table's columns, and `RESTCatalog::alter_table` is entirely unimplemented.

In addition, the `SchemaChange` JSON does not match Java Paimon's wire format (Java uses an `@JsonTypeInfo` discriminator `action` with `fieldNames` arrays, `newDataType`/`keepNullability`, `referenceFieldName`, and FIRST/AFTER/BEFORE/LAST moves), so a REST `alter_table` could never interoperate with a Java server.

### Proposal

- Implement all column-level changes in `apply_changes` for top-level columns, mirroring Java `SchemaManager.generateTableSchema` (field-id allocation, move semantics, primary/partition-key guards).
- Align the `SchemaChange` JSON wire format with Java Paimon.
- Add `AlterTableRequest` + `RESTApi::alter_table` and implement `RESTCatalog::alter_table`.

### Out of scope (follow-ups)

- Nested struct field paths (only top-level columns for now).
- `UpdateColumnType` cast-compatibility validation (lenient replace for now).
- Java's `dropPrimaryKey` / `updateColumnDefaultValue` variants.

TheR1sing3un:
### Purpose

Linked issue: close #368

Implement column-level `alter table` on both `FileSystemCatalog` and `RESTCatalog`, and align the `SchemaChange` JSON wire format with Java Paimon.

### Brief change log

- `TableSchema::apply_changes`: implement add / rename / drop column, update column type / nullability / comment / position, and update table comment, mirroring Java `SchemaManager.generateTableSchema` (field-id allocation, FIRST/AFTER/BEFORE/LAST moves, primary/partition-key guards). Operates on top-level columns; reuses `ColumnAlreadyExist` / `ColumnNotExist`. The method keeps its single-argument signature — the catalog fills in the table name on column errors.
- Align `SchemaChange` JSON with Java Paimon: internally tagged by `action`, with `fieldNames` arrays, `comment` / `newDataType` / `keepNullability` / `newNullability` / `newComment` fields, `referenceFieldName` move anchors, and FIRST/AFTER/BEFORE/LAST move types.
- Add `AlterTableRequest` + `RESTApi::alter_table`; implement `RESTCatalog::alter_table`.

### Tests

- New unit tests in `spec::schema_change` (Java-format (de)serialization), `spec::schema`, and `catalog::filesystem` (add+move, rename syncing PK refs, drop rejecting PK columns, type/nullability/comment updates, reposition, and the `ColumnAlreadyExist` / `ColumnNotExist` / `TableNotExist` error paths).
- Existing `paimon-datafusion` alter-table tests pass.

### API and Format

Changes the `SchemaChange` **JSON wire format**. This is effectively a fix: the previous format was incompatible with Java Paimon, and REST `alter_table` was never implemented, so no working interaction is broken. No public Rust API **signature** change (`apply_changes` keeps its arity).

### Documentation

None required.

#### Out of scope (follow-ups)

Nested struct field paths; `UpdateColumnType` cast-compatibility validation; Java's `dropPrimaryKey` / `updateColumnDefaultValue`.


TheR1sing3un:
Pushed a fix for the `integration` job: the existing `test_alter_table_add_column` (in `crates/integrations/datafusion/tests/sql_context_tests.rs`) asserted that `ALTER TABLE ... ADD COLUMN` *fails* because AddColumn was unsupported. Since this PR implements it, the test now asserts the statement succeeds and the new column is appended to the schema. Verified locally: `cargo test -p paimon-datafusion --test sql_context_tests test_alter_table` passes.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +127,153 @@ impl TableSchema {
...
+                } => {
+                    let name = top_level_field(&field_names)?;
+                    if field_index(&fields, name).is_some() {
+                        return Err(crate::Error::ColumnAlreadyExist {
+                            full_name: full_name.to_string(),
+                            column: name.to_string(),
+                        });
+                    }
+                    highest_field_id += 1;
+                    let field = DataField::new(highest_field_id, name.to_string(), data_type)
```
JingsongLi: This adds the requested type directly to the table schema. That misses two invariants that Java applies when adding a column: NOT NULL columns are rejected, and nested field IDs are reassigned from the table-wide highestFieldId. As written, ALTER TABLE can add `INT NOT NULL` columns whose existing rows have no value, and can add ROW/ARRAY/MAP-with-ROW columns whose nested DataField IDs collide with existing field IDs because the SchemaBuilder assignment path is bypassed. Please reject non-nullable add-column types and assign/recompute nested field IDs before storing the field.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +127,153 @@ impl TableSchema {
...
+                    if new_schema.partition_keys.iter().any(|k| k == name)
+                        || new_schema.primary_keys.iter().any(|k| k == name)
+                    {
+                        return Err(crate::Error::Unsupported {
+                            message: format!(
+                                "Cannot drop partition or primary key column '{name}' of table {full_name}"
+                            ),
+                        });
+                    }
+                    fields.remove(idx);
```
JingsongLi, line 313: This allows dropping the last non-key column and persisting a schema with `fields: []`. Java rejects this (`Cannot drop all fields in table`), and several Rust paths assume a table schema still has data fields for bucket keys, scans, writes, and partition/key resolution. Please reject the drop when it would leave the table with no fields.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +127,153 @@ impl TableSchema {
...
+                            column: name.to_string(),
+                        })?;
+                    let old = &fields[idx];
+                    // Lenient: replace the type without cast-compatibility checks.
+                    let target = if keep_nullability {
+                        new_data_type.copy_with_nullable(old.data_type().is_nullable())?
+                    } else {
+                        new_data_type
+                    };
+                    fields[idx] = DataField::new(old.id(), old.name().to_string(), target)
```
JingsongLi, line 382: Type changes on key columns should be rejected before persisting the new schema. Java rejects updating partition keys and primary keys here because existing partitions, bucket assignment, key encoding, and manifests were written with the old type. This branch currently permits `updateColumnType` on those columns, so a filesystem catalog can write an incompatible schema version that later readers/writers interpret with the wrong key type. Please add the same partition-key/primary-key guard for type updates.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +127,153 @@ impl TableSchema {
...
+                    new_nullability,
+                } => {
+                    let name = top_level_field(&field_names)?;
+                    let idx =
+                        field_index(&fields, name).ok_or_else(|| crate::Error::ColumnNotExist {
+                            full_name: full_name.to_string(),
+                            column: name.to_string(),
+                        })?;
+                    let old = &fields[idx];
+                    let nt = old.data_type().copy_with_nullable(new_nullability)?;
```
JingsongLi, line 409: This can make a primary-key column nullable (`newNullability = true`). Primary keys are normalized to NOT NULL at create time, and Java rejects changing primary-key nullability to nullable; otherwise subsequent writes can accept NULLs in key columns and break key/bucket semantics. Please reject `UpdateColumnNullability` when the target column is a primary key and `new_nullability` is true.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +127,153 @@ impl TableSchema {
...
+                            column: name.to_string(),
+                        })?;
+                    if new_name != name && field_index(&fields, &new_name).is_some() {
+                        return Err(crate::Error::ColumnAlreadyExist {
+                            full_name: full_name.to_string(),
+                            column: new_name,
+                        });
+                    }
+                    fields[idx] = fields[idx].clone().with_name(new_name.clone());
+                    rename_in_keys(&mut new_schema.partition_keys, name, &new_name);
```
JingsongLi: Renaming a partition column should be rejected instead of updating `partition_keys`. Existing partition data is laid out using the old key names in paths and metadata, while future writes after this rename would resolve/generate partitions with the new key name; Java Paimon rejects partition-key renames for this reason. Please add a guard here for partition keys and only allow non-partition columns to be renamed.

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +134,204 @@ impl TableSchema {
...
+                            message: "Cannot update primary key".to_string(),
+                        });
+                    }
+                    let idx =
+                        field_index(&fields, name).ok_or_else(|| crate::Error::ColumnNotExist {
+                            full_name: full_name.to_string(),
+                            column: name.to_string(),
+                        })?;
+                    let old = &fields[idx];
+                    // Lenient: replace the type without cast-compatibility checks.
```
JingsongLi: This lets `UpdateColumnType` commit any replacement type without checking whether existing data files can be read as the new type. Existing files keep the old schema, and `DataFileReader` later tries to cast old Arrow columns to the new target type; unsupported changes can therefore succeed at ALTER TABLE time and fail when users read existing rows. Java `SchemaManager` rejects unsupported casts before committing, so this should either validate supported type evolution here or reject unsupported changes.

crates/paimon/src/spec/schema.rs
```diff
@@ -127,27 +134,204 @@ impl TableSchema {
...
+                    apply_move(&mut fields, &column_move, full_name)?;
                 }
             }
         }
 
+        new_schema.fields = fields;
+        new_schema.highest_field_id =
+            highest_field_id.max(Self::current_highest_field_id(&new_schema.fields));
+
         Schema::validate_first_row_changelog_producer(&new_schema.options)?;
```
JingsongLi, line 446: After applying column and option changes, this only re-runs the first-row changelog validation. Create-time schema validation also enforces invariants such as BLOB columns requiring `data-evolution.enabled` and partial-update options being limited to the supported basic mode, but alter can now add/update a column to BLOB or set unsupported partial-update options and still persist the schema. Please re-run the relevant final-schema validations before saving the new schema.

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed in 6973290. Added `spec::data_type_casts::supports_cast` ported from Java `DataTypeCasts` (implicit/explicit rules, honoring `disable-explicit-type-casting`), and additionally required `arrow_cast::can_cast_types` on the converted arrow types — the Rust analog of Java's `CastExecutors.resolve(...) != null` — so a committed change is guaranteed executable by the read path. Also mirrored the adjacent Java guards in the same code block: `assertNotChangingBlobColumnType` (type changes involving BLOB are rejected) and `assertNullabilityChange` (nullable -> NOT NULL rejected by default per `alter-column-null-to-not-null.disabled`, applied to both UpdateColumnType and UpdateColumnNullability).

crates/paimon/src/spec/schema.rs
TheR1sing3un: Fixed in 6973290. `apply_changes` now re-runs the create-time final-schema validations before returning: `validate_blob_fields` (BLOB columns require `data-evolution.enabled`, must coexist with normal columns, not in partition keys) and `PartialUpdateConfig::validate_create_mode` (partial-update limited to the supported basic mode), in addition to the existing first-row changelog check — mirroring Java's `SchemaValidation.validateTableSchema(newSchema)` after applying changes.

JingsongLi:
Cross-PR note with #340: if `merge-engine=aggregation` lands before or together with this alter-table work, `RenameColumn` needs to update aggregation field-scoped options as well.

Right now the rename path rewrites `primary_keys`, `bucket-key`, and `sequence.field`, but it does not rename keys such as `fields.<col>.aggregate-function` or `fields.<col>.list-agg-delimiter`. After renaming an aggregation value column, those options would still point at the old column name. Since the final validation here reruns blob / partial-update / first-row checks but not aggregation validation, the stale option can be persisted and the renamed column may silently fall back to the default aggregator/delimiter at read time.

This is not visible against current `main` by itself, but it should be handled if this PR is intended to compose with #340.

TheR1sing3un:
Thanks — confirmed. The rename path here only ports case 1 of Java's applyRenameColumnsToOptions (primary_keys / bucket-key / sequence.field); it doesn't rewrite the field-scoped fields.<col>.* keys (Java case 2: aggregate-function / ignore-retract / distinct / list-agg-delimiter, and case 3: sequence-group / nested-key). Worth noting this is already latent for partial-update tables on main, not only when composing with https://github.com/apache/paimon-rust/pull/340. To keep this PR scoped to the alter-table feature as reviewed, I'll track the field-scoped option rewriting as a follow-up: https://github.com/apache/paimon-rust/issues/383.
@JingsongLi

crates/paimon/src/spec/schema.rs
```diff
@@ -140,32 +146,268 @@ impl TableSchema {
...
+                                "Cannot drop partition or primary key column '{name}' of table {full_name}"
+                            ),
+                        });
+                    }
+                    if fields.len() == 1 {
+                        return Err(crate::Error::Unsupported {
+                            message: "Cannot drop all fields in table".to_string(),
+                        });
+                    }
+                    fields.remove(idx);
```
JingsongLi, line 313: Dropping a column that is referenced by table options leaves those options pointing at a field that no longer exists. For example, a fixed-bucket append table with `bucket-key=name` can drop `name`; `TableWrite` then still sees `bucket-key` configured, but `bucket_key_indices` becomes empty and writes fall back to `ConstantBucketAssigner(0)`. Similarly, dropping a `sequence.field` column silently disables sequence ordering. Please reject drops of columns referenced by `bucket-key`/`sequence.field` or update/remove those options as part of the alter.

crates/paimon/src/spec/schema.rs
```diff
@@ -140,32 +146,268 @@ impl TableSchema {
...
+                    }
+                    fields[idx] = fields[idx].clone().with_name(new_name.clone());
+                    rename_in_keys(&mut new_schema.primary_keys, name, &new_name);
+                    rename_in_option_list(
+                        &mut new_schema.options,
+                        BUCKET_KEY_OPTION,
+                        name,
+                        &new_name,
+                    );
+                    rename_in_option_list(
```
JingsongLi, line 256: This only rewrites `bucket-key` and `sequence.field`, but aggregation tables also have field-scoped options such as `fields.<col>.aggregate-function` and `fields.<col>.list-agg-delimiter`. After renaming `tag` to `label`, `fields.tag.list-agg-delimiter` remains under the old key; `ListaggAgg::new("label", ...)` will not find it and will silently switch back to the default delimiter. Java rewrites these field-scoped option keys in `applyRenameColumnsToOptions`, so we should update them here as well to preserve aggregation semantics across renames.

crates/paimon/src/spec/schema.rs
TheR1sing3un: fixed

crates/paimon/src/spec/schema.rs
TheR1sing3un: fixed

JingsongLi (approved):
+1

PR #371, opened 2026-06-10
feat(rest-server): add FileSystemCatalog-backed REST catalog server

### LINKED ISSUE #369: Add a FileSystemCatalog-backed REST catalog server for end-to-end testing
### Problem

The only REST test fixture today is an in-memory mock. It does not persist data, does not return real table schemas/paths, and does not support commit, so the `RESTCatalog` client cannot be exercised end to end (metadata + append write + commit + read).

### Proposal

Add a small dev crate (`paimon-rest-server`, `publish = false`) that serves the Paimon REST protocol over a real `FileSystemCatalog`:

- config + database/table metadata CRUD;
- append write + commit (the commit endpoint persists the posted snapshot via `SnapshotManager`) + read back;
- column-level `alter table`.

Because the server and client share the same local warehouse, and the wire format mirrors Java Paimon, the same warehouse can be round-tripped with a Java reader/writer. A CI job runs the crate's e2e tests.

This depends on the column-level alter table work tracked separately.

TheR1sing3un:
### Purpose

Linked issue: close #369

> [!NOTE]
> **Stacked on #370** (column-level alter table). This branch currently also
> contains #370's commit, so the diff includes it until #370 merges. Kept as a
> **draft** for now; once #370 is merged I will rebase and only the
> `paimon-rest-server` commit will remain.

Add a real REST catalog server backed by `FileSystemCatalog`, so the `RESTCatalog` client can be tested end to end without a Java server.

### Brief change log

- New `paimon-rest-server` crate (library + binary, `publish = false`) that maps the Paimon REST protocol onto a real `FileSystemCatalog`: config, database/table metadata CRUD, append write + commit (the commit endpoint persists the posted snapshot via `SnapshotManager`) + read back, and column-level alter table.
- The wire format mirrors Java Paimon, so the existing `RESTCatalog` client drives it and warehouses round-trip with Java.
- New CI job running the crate's e2e tests; crate `README`.

### Tests

`tests/e2e.rs`: metadata CRUD, write/commit/read round trip, and alter-table over REST. Wired into CI via a new `cargo test -p paimon-rest-server` step.

### API and Format

No change to `paimon`'s public API or storage format. The new crate introduces `axum`/`tower` (MIT) dependencies; `cargo deny check licenses` passes without changes to `deny.toml`.

### Documentation

Crate `README.md` (purpose, usage, endpoints).

### DIFF
--- a/crates/paimon-rest-server/src/lib.rs
+++ b/crates/paimon-rest-server/src/lib.rs
@@ -52,9 +52,9 @@ use serde_json::json;
 use paimon::api::{
     AlterDatabaseRequest, AlterTableRequest, AuditRESTResponse, ConfigResponse, CreateTableRequest,
     ErrorResponse, GetDatabaseResponse, GetTableResponse, ListDatabasesResponse,
-    ListTablesResponse, RenameTableRequest, ResourcePaths,
+    ListPartitionsResponse, ListTablesResponse, RenameTableRequest, ResourcePaths,
 };
-use paimon::catalog::{Catalog, Identifier};
+use paimon::catalog::{list_partitions_from_file_system, Catalog, Identifier};
 use paimon::common::{CatalogOptions, Options};
 use paimon::spec::{Schema, Snapshot};
 use paimon::table::SnapshotManager;
@@ -168,6 +168,10 @@ fn build_router(prefix: &str, state: Arc<AppState>) -> Router {
         .route(
             &format!("{base}/databases/:db/tables/:table/commit"),
             post(commit),
+        )
+        .route(
+            &format!("{base}/databases/:db/tables/:table/partitions"),
+            get(list_partitions),
         )
         // Token endpoint is never hit when `data-token.enabled=false` (default),
         // but we serve a stub so a misconfigured client gets a clear 501.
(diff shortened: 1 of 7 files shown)


TheR1sing3un:
@JingsongLi Hello, I'm not sure if the community needs this ability, so I'd like to ask for your opinion first. Internally, we mainly use rest catalogs. However, when conducting rest catalog tests on rust, it is often necessary to start the real rest catalog before we can test the behavior of real rest clients on the rust side. Therefore, I introduced a Rust version of the Rest Catalog that is directly based on the FileSystem wrapper. There is similar code in Java as well. Do you think this contribution to the community is reasonable? Or do I need to start it directly into the Test module instead of using it as a complete rest catalog server?

crates/paimon-rest-server/src/lib.rs
```diff
@@ -0,0 +1,493 @@
...
+        )
+        .route(
+            &format!("{base}/databases/:db/tables"),
+            get(list_tables).post(create_table),
+        )
+        .route(
+            &format!("{base}/databases/:db/tables/:table"),
+            get(get_table).post(alter_table).delete(drop_table),
+        )
+        .route(&format!("{base}/tables/rename"), post(rename_table))
```
JingsongLi, line 168: The REST client already exposes `list_partitions` and calls `/v1/.../partitions`, but this router does not serve that path. A request through `RESTCatalog::list_partitions` will get Axum’s default 404, which maps to `NoSuchResource`; the catalog fallback only handles 501 NotImplemented, so partition listing fails instead of returning partitions or gracefully falling back. Please add the partitions route, or an explicit 501 stub like the token endpoint, and cover it in e2e.

TheR1sing3un: fixed

crates/paimon-rest-server/src/lib.rs
```diff
@@ -0,0 +1,523 @@
...
+        .create_database(&payload.name, false, payload.options)
+        .await
+    {
+        Ok(()) => ok_empty(),
+        Err(e) => error_response(e),
+    }
+}
+
+async fn get_database(
+    Path(db): Path<String>,
```
JingsongLi: The path params need to be decoded with the same codec the client uses. `ResourcePaths` builds paths with `RESTUtil::encode_string`, which is `application/x-www-form-urlencoded`; spaces therefore become `+`. Axum's `Path` extractor percent-decodes `%xx`, but it leaves `+` untouched, so valid catalog names containing spaces become unaddressable through `RESTCatalog`.

I reproduced this against this server: `POST /v1/databases` with `{"name":"db 1"}` succeeds, but the client-style path `GET /v1/databases/db+1` returns 404 for `db+1`; only `GET /v1/databases/db%201` works. Since `Identifier::validate_*` allows non-empty names with spaces, `RESTCatalog::create_table(&Identifier::new("db 1", ...))` will also fail because this handler family looks for database `db+1`.

Please decode all captured database/table path params with the existing REST path codec (or change the client path encoder consistently) and add an e2e case with a space in the database/table name.

TheR1sing3un: fixed

JingsongLi (approved):
+1

PR #374, opened 2026-06-10
fix: merge primary-key rows correctly when bucket files span multiple splits

### LINKED ISSUE #373: Primary-key tables return unmerged rows when one bucket's files span multiple splits
### Bug description

`TableScan::plan_snapshot` bin-packs the files of a (partition, bucket) purely
by size (`split_for_batch`), and the PK read path builds one sort-merge reader
per `DataSplit`. When files holding versions of the same primary key land in
different splits, each split emits its own version of the key — SELECT returns
multiple rows where it should return one merged row.

### Reproduce

```sql
CREATE TABLE t (id INT NOT NULL, value INT, PRIMARY KEY (id)) WITH (
  'bucket' = '1',
  'source.split.target-size' = '1b',
  'source.split.open-file-cost' = '1b'
);
INSERT INTO t VALUES (1, 10);
INSERT INTO t VALUES (1, 20);
INSERT INTO t VALUES (1, 30);
SELECT * FROM t;  -- returns 3 rows; expected 1 row (1, 30)
```

### Affected

All PK merge engines on the non-data-evolution path (`deduplicate`,
`partial-update`; `aggregation` once #340 lands). A second instance of the
same root cause: `read_pk` raw-reads splits that contain no level-0 files,
but compacted files on different levels can still overlap on key range
(e.g. written by Java/Spark compaction) and need merging.

### Fix direction

Port Java's `MergeTreeSplitGenerator` / `IntervalPartition`: group a bucket's
files into key-range-overlap sections, bin-pack whole sections into splits,
and make the read-side kv/raw dispatch overlap-aware. Deletion-vector and
first-row tables keep the raw fast path (stale rows are masked by DVs /
level-0 is skipped), mirroring Java's fast path.

First reported by @JingsongLi in
https://github.com/apache/paimon-rust/pull/340#discussion_r3384978245

TheR1sing3un:
### Purpose

Linked issue: close #373

PK tables could return unmerged (duplicate-key) rows in two ways sharing one
root cause — split planning and read dispatch were blind to key-range overlap:

1. `plan_snapshot` bin-packed a bucket's files purely by size, so files
   holding versions of the same key could land in different splits; each
   split runs its own sort-merge reader and emits its own version.
   Reproducible with `source.split.target-size=1b` and three commits of one
   key: SELECT returned 3 rows instead of 1.
2. `read_pk` sent splits without level-0 files to the raw (non-merging)
   reader, but compacted files on different levels can still overlap on key
   range (e.g. produced by Java/Spark compaction) and need merging.

Reported by @JingsongLi while reviewing #340; affects `deduplicate` and
`partial-update` on main, and `aggregation` once #340 lands.

### Changes

- **`table/merge_tree_split_generator.rs` (new)** — port of Java
  `MergeTreeSplitGenerator` / `IntervalPartition`:
  - `KeyComparator` decodes serialized BinaryRow min/max keys with the
    trimmed-PK data types and compares via `datum_cmp`. BinaryRow stores
    fields little-endian, so raw byte comparison would order int 256 before
    int 1 — decoding is mandatory for correctness.
  - `interval_partition` sorts files by decoded `(min_key, max_key)` and
    groups transitively overlapping files into sections; sections never
    overlap each other.
  - `pack_sections` bin-packs whole sections into splits (reusing
    `pack_for_ordered`); a section is atomic and never separated.
  - Fail-safe: empty/undecodable key ranges collapse the bucket into one
    section — losing parallelism, never correctness.
- **`table_scan.rs`** — PK tables route through
  `pack_sections(interval_partition(...))` on the non-data-evolution path;
  append-only tables keep the existing file-level `split_for_batch`.
- **`table_read.rs`** — `read_pk` dispatch is now overlap-aware
  (`split_requires_merge`): any level-0 file or key-overlapping compacted
  files → sort-merge reader; only disjoint compacted files keep the raw
  fast path.
- **Deletion-vector / first-row fast path** — mirroring Java
  `MergeTreeSplitGenerator`: DV tables resolve stale versions through DVs
  (and `KeyValueFileReader` rejects DV splits), and first-row tables skip
  level-0 at plan time, so both keep plain size-based packing and the plain
  level-0 read dispatch.

### Tests

- 16 unit tests: comparator ordering (little-endian regression), section
  grouping, running-bound chains, atomic packing, overlap detection,
  undecodable-key degradation.
- 3 plan-level regression tests (memory FileIO, real write→commit→plan→read):
  overlapping files share one split and read back merged; disjoint files keep
  split parallelism; append tables keep file-level bin pack.
- 2 DataFusion e2e tests reproducing the reviewer scenario for `deduplicate`
  and `partial-update`. All new tests fail without the fix.
- Verified against the Spark-provisioned warehouse
  (`make docker-up` + `cargo test -p paimon-datafusion --all-targets`),
  which caught and now guards the deletion-vector interaction.

### Out of scope

- `first-row` reads never go through `read_pk` (pre-existing routing); the
  same overlap consideration for its raw path can be tracked separately.
- Java's `rawConvertible` flag plumbing through `DataSplit`; this PR derives
  the same decision read-side from file metadata instead.

### DIFF
--- a/crates/paimon/src/table/merge_tree_split_generator.rs
+++ b/crates/paimon/src/table/merge_tree_split_generator.rs
@@ -1,0 +1,353 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Split generation for primary-key (merge-tree) tables.
+//!
+//! Files whose primary-key ranges overlap must be read by the same
+//! sort-merge reader, so they have to stay in the same split. This module
+//! groups a bucket's files into key-range "sections" and then bin-packs
+//! whole sections into splits, mirroring the Java implementation.
+//!
+//! References:
+//! [MergeTreeSplitGenerator](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/table/source/MergeTreeSplitGenerator.java),
+//! [IntervalPartition](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/mergetree/compact/IntervalPartition.java)
+
+use super::bin_pack::pack_for_ordered;
+use crate::spec::{datum_cmp, BinaryRow, DataFileMeta, DataType, Datum};
+use std::cmp::{self, Ordering};
+
+/// Compares serialized `BinaryRow` keys field-by-field using the trimmed
+/// primary-key data types.
+///
+/// BinaryRow stores fields little-endian, so raw byte comparison would order
+/// e.g. int 256 (`[00 01 00 00]`) before int 1 (`[01 00 00 00]`); keys must
+/// be decoded before comparing.
+pub(crate) struct KeyComparator {
+    key_types: Vec<DataType>,
+}
+
+/// A decoded key: one `Option<Datum>` per trimmed primary-key field
+/// (`None` = SQL NULL).
+type DecodedKey = Vec<Option<Datum>>;
+
+impl KeyComparator {
+    pub(crate) fn new(key_types: Vec<DataType>) -> Self {
+        Self { key_types }
+    }
+
+    /// Decode a serialized min/max key. Returns `None` when the key is empty
+    /// or malformed, letting callers degrade to the safe "treat everything as
+    /// overlapping" path instead of failing the scan.
+    fn decode(&self, key: &[u8]) -> Option<DecodedKey> {
+        if key.is_empty() {
+            return None;
+        }
+        let row = BinaryRow::from_serialized_bytes(key).ok()?;
+        if (row.arity() as usize) != self.key_types.len() {
+            return None;
+        }
+        self.key_types
+            .iter()
+            .enumerate()
+            .map(|(pos, dt)| row.get_datum(pos, dt).ok())
+            .collect()
+    }
+}
+
+/// Compare decoded keys field-by-field. NULL sorts first; fields that
+/// `datum_cmp` cannot order (e.g. float NaN) compare as equal, which forces
+/// the files into the same section — conservative but never incorrect.
+fn compare_decoded(a: &DecodedKey, b: &DecodedKey) -> Ordering {
+    for (fa, fb) in a.iter().zip(b.iter()) {
+        let ord = match (fa, fb) {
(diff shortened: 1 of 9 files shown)


crates/paimon/src/table/table_scan.rs
```diff
@@ -721,6 +740,14 @@ impl<'a> TableScan<'a> {
                 }
 
                 result
+            } else if let Some(ref comparator) = pk_comparator {
+                // Merge-tree path: section files by key-range overlap first, then
+                // bin-pack whole sections. Overlapping files always share a split.
+                pack_sections(
```
JingsongLi: This branch can make `merged_row_count()` overestimate rows for PK tables: once overlapping files are packed into the same split, `DataSplit::merged_row_count()` still falls back to summing the physical `row_count`s because these files have no `first_row_id`. If a LIMIT hint is pushed down, `apply_limit_pushdown` can stop after this split even though the sort-merge reader later collapses those versions to fewer logical rows, so queries can return fewer rows than the requested LIMIT. Could we disable limit split pruning for merge-tree packed PK scans (or teach `merged_row_count` to return unknown for these splits)?

TheR1sing3un: fixed

crates/paimon/src/table/merge_tree_split_generator.rs
```diff
@@ -0,0 +1,551 @@
...
+    }
+
+    pack_sections(
+        interval_partition(files, comparator),
+        target_split_size,
+        open_file_cost,
+    )
+    .into_iter()
+    .map(|files| {
+        let raw_convertible = files.len() == 1 && without_delete_row(&files[0]);
```
QuakeWang: This is not safe for partial-update tables. The Rust partial-update writer can keep multiple physical rows for the same primary key inside a single data file, and `TableRead` always routes partial-update reads through the KV merge reader for that reason.

If this single-file group is marked `raw_convertible=true`, `DataSplit::merged_row_count()` reports the physical row count as exact. That can make `COUNT(*)` return too many rows and can let LIMIT pushdown prune later splits too early.

A minimal case is one INSERT writing three partial updates for the same key: normal SELECT merges to one row, but COUNT(*) sees 3. Please keep partial-update splits non-raw-convertible unless file-internal key uniqueness is proven, and add COUNT/LIMIT regression coverage.

TheR1sing3un: Confirmed — thanks for the careful analysis. The root cause is a writer-side gap rather than the split generator itself: Java's `MergeTreeWriter#flushWriteBuffer` runs the engine's merge function over the write buffer before flushing, so every data file has unique keys regardless of merge engine — that invariant is what lets `MergeTreeSplitGenerator` mark single-file groups raw convertible. The Rust writer currently only deduplicates at flush for deduplicate/first-row (`select_flush_indices`), while partial-update intentionally keeps all rows for read-side field-wise merge, so the uniqueness premise doesn't hold for it.

Fixed in this PR as you suggested: split groups are now raw convertible only when the merge engine guarantees file-internal key uniqueness (currently deduplicate only), so partial-update splits always report an unknown merged row count — consistent with the read path, which always routes them through the merge reader. Added COUNT(*) and LIMIT regression tests using your minimal case.

As a follow-up, #380 ports Java's write-time merge (`MergeTreeWriter#flushWriteBuffer`) for partial-update, restoring the file-internal uniqueness invariant; once that lands the conservative gating here can be relaxed.

JingsongLi (approved):
+1

PR #375, opened 2026-06-10
fix: keep Vortex runtime alive during async IO

QuakeWang:
### Purpose

Fixes the Ubuntu unit test hang in Vortex format tests.

The previous code used `VortexSession::default()` directly. In Vortex 0.68, runtime handles are weak references to the underlying executor, so async read/write work can outlive the temporary runtime handle owner and stall when Vortex later spawns background IO tasks.

### Brief change log

- Create Vortex sessions from the active Tokio runtime.
- Keep the Vortex runtime owner alive for the full reader stream lifetime.
- Keep the Vortex runtime owner alive while the background writer task is running.

### Tests

- `git diff --check`

### API and Format

### Documentation


JingsongLi (approved):
+1

PR #377, opened 2026-06-11
fix(scan): net manifest add/delete by full file identity including level

### LINKED ISSUE #376: Scan drops in-place upgraded data file, silently losing rows on read
### Search before asking

### Please describe the bug 🐞

When planning a scan, `merge_manifest_entries` in `crates/paimon/src/table/table_scan.rs` nets manifest `ADD`/`DELETE` entries using the identity tuple `(partition, bucket, file_name)` — it **omits `level`**:

```rust
let deleted_keys: HashSet<(&[u8], i32, &str)> = delete_entries
    .iter()
    .map(|e| (e.partition(), e.bucket(), e.file().file_name.as_str()))
    .collect();
// keep adds whose (partition, bucket, file_name) is not in deleted_keys
```

A single-sorted-run compaction upgrades a file **in place**: Java `PojoDataFileMeta.upgrade(newLevel)` reuses the same `fileName` and only bumps `level`, so the commit emits `DELETE f@oldLevel` + `ADD f@newLevel` with the **same file name**. Because the dedup key ignores `level`, the `DELETE@oldLevel` cancels **both** the old add and the upgraded `ADD@newLevel`, so the live (upgraded) file is dropped from the plan and **its rows are silently lost on read**.

This diverges from Java, where `FileEntry.Identifier` includes `level` in `equals`/`hashCode`, and `AbstractFileStoreScan.readAndMergeFileEntries` nets add/delete by that full identifier — so Java is unaffected.

**Manifest entries for an upgraded file (what the scan sees):**

```
ADD    f.parquet  level=0   (initial write)
DELETE f.parquet  level=0   (compaction)
ADD    f.parquet  level=5   (same file upgraded in place)
ADD    g.parquet  level=0   (later write)
```
Expected live set: `{f@L5, g@L0}`. Actual (buggy): `{g@L0}` — `f` lost entirely.

**Minimal reproduction (real):** a primary-key table with a single `INSERT` (one sorted run), then `CALL sys.compact` (in-place upgrade), then read — the table reads back as if the compacted file does not exist (rows missing / empty).

**Reproduction (hermetic unit test), reverting only the fix shows:**
```
test table::table_scan::tests::test_merge_manifest_entries_keeps_in_place_upgraded_file ... FAILED
assertion `left == right` failed: upgraded file (f@L5) must survive; only f@L0 is cancelled by the DELETE
  left: [("g.parquet", 0)]
 right: [("f.parquet", 5), ("g.parquet", 0)]
```

Environment: apache/paimon-rust `main` (also present on current `main`). Affects any primary-key table that has undergone a single-run compaction (level upgrade).

### Solution

Net add/delete by the **full file identity including `level`** (and the other fields of `FileEntry.Identifier`), mirroring Java `AbstractFileStoreScan.readAndMergeFileEntries`: collect every `DELETE` entry's `Identifier`, then keep `ADD`s whose `identifier()` is not deleted. The codebase already has a correct `ManifestEntry::identifier()` (incl. `level`) — the scan dedup just doesn't use it. Also align `Identifier`'s `Hash` impl with its `Eq` (the current `Hash` only hashes `partition/bucket/file_name`, so upgraded files all collide) to match Java `FileEntry.Identifier.hashCode`.

TheR1sing3un:
### Purpose

Linked issue: close #376

When planning a scan, `merge_manifest_entries` (`crates/paimon/src/table/table_scan.rs`) nets manifest `ADD`/`DELETE` entries by `(partition, bucket, file_name)`, **omitting `level`**. A single-sorted-run compaction upgrades a file *in place* (Java `PojoDataFileMeta.upgrade` reuses the file name and only bumps `level`), emitting `DELETE f@oldLevel` + `ADD f@newLevel` with the same name. The level-less key lets the `DELETE` cancel the upgraded `ADD` too, so the live file is dropped from the plan and **its rows are silently lost on read**.

This mirrors the fix to Java semantics: `FileEntry.Identifier` includes `level`, and `AbstractFileStoreScan.readAndMergeFileEntries` nets add/delete by that full identifier.

### Brief change log

- Rewrite `merge_manifest_entries` to collect every `DELETE`'s full `Identifier` (incl. `level`) and keep `ADD`s whose `identifier()` is not deleted — reusing the existing `ManifestEntry::identifier()`. This matches Java `AbstractFileStoreScan.readAndMergeFileEntries` and is independent of add/delete ordering.
- Align `Identifier`'s `Hash` impl with its `Eq` (and Java `FileEntry.Identifier.hashCode`): hash all identity fields (`partition, bucket, level, file_name, extra_files, embedded_index, external_path`). The previous `Hash` only hashed `partition/bucket/file_name`, so in-place upgraded files all collided in the dedup set.

### Tests

Two hermetic unit tests (pure logic, no external deps):

- `table_scan::tests::test_merge_manifest_entries_keeps_in_place_upgraded_file`: feeds `ADD f@L0, DELETE f@L0, ADD f@L5, ADD g@L0` and asserts the live set is `{f@L5, g@L0}`.
- `manifest_entry::tests::test_identifier_distinguishes_in_place_level_upgrade`: `f@L0` and `f@L5` are unequal and do not alias in a `HashSet<Identifier>`.

Before/after (revert the fix → the first test fails, restore → passes):
```
# before fix
test ... test_merge_manifest_entries_keeps_in_place_upgraded_file ... FAILED
  left: [("g.parquet", 0)]
 right: [("f.parquet", 5), ("g.parquet", 0)]
# after fix
test result: ok. 684 passed; 0 failed
```
`cargo fmt --all -- --check` and `cargo clippy -p paimon --all-targets` are clean.

### API and Format

### Documentation

Not needed.

### DIFF
--- a/crates/paimon/src/spec/manifest_entry.rs
+++ b/crates/paimon/src/spec/manifest_entry.rs
@@ -35,10 +35,19 @@ pub struct Identifier {
 }
 
 impl Hash for Identifier {
+    // Hash every field that participates in `Eq`, matching Java
+    // `FileEntry.Identifier.hashCode`. `level` in particular must be included:
+    // a single-run compaction upgrades a file in place (same name, new level),
+    // producing distinct identifiers that would otherwise all collide and
+    // degrade the hash sets used to net add/delete entries during scan.
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.partition.hash(state);
         self.bucket.hash(state);
+        self.level.hash(state);
         self.file_name.hash(state);
+        self.extra_files.hash(state);
+        self.embedded_index.hash(state);
+        self.external_path.hash(state);
     }
 }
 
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #379, opened 2026-06-11
fix(table): use the snapshot's schema version for time travel reads

TheR1sing3un:
### Purpose

Time travel (`scan.version` / `scan.timestamp-millis` / SQL `VERSION AS OF` / `TIMESTAMP AS OF`) previously only switched which snapshot was scanned, while the table schema, scan pruning, the read evolution target, and the DataFusion provider all kept using the **latest** schema — `Snapshot.schemaId` was never consumed on the read path. Reading an old snapshot therefore lost its historical shape: columns dropped later were invisible, and type updates were applied retroactively to historical data.

Java switches the table to the snapshot's schema in `AbstractFileStoreTable.copy(dynamicOptions)` → `tryTimeTravel` → `schemaManager.schema(snapshot.schemaId()).copy(mergedOptions)`. This PR mirrors that behavior.

### Brief change log

- Add async `Table::copy_with_time_travel(extra)`, mirroring Java `copy(dynamicOptions)`: merge options, resolve the time-travel selector, and when the resolved snapshot has a different schema id, replace the table schema with the snapshot's schema while keeping the merged options. Resolution failures fall back silently (Java `tryTimeTravel` catch-all); invalid selectors still fail at scan planning, so existing error behavior is unchanged. The existing `Table::copy_with_options` keeps its non-traveling semantics (Java `copyWithoutTimeTravel`).
- Add `TableSchema::copy_with_replaced_options`, matching Java `TableSchema.copy(Map)` (options are replaced, not merged; id/fields/keys/comment/timeMillis preserved).
- Extract snapshot resolution from `TableScan::resolve_snapshot` into `table::time_travel::travel_to_snapshot` (Java `TimeTravelUtil`) and share it.
- Wire up the DataFusion entry points: the `SQLContext` time-travel rewrite path, the catalog provider's dynamic-options path (`SET 'paimon.scan.version'`), and `PaimonRelationPlanner` (bridged via the existing `block_on_with_runtime`, since the planner hook is synchronous).
- Reject `new_write()` on a time-travelled table copy. Java has no runtime guard but avoids the situation structurally — write paths always use `copyWithoutTimeTravel` — whereas the shared DataFusion provider here can serve both reads and INSERT, so an explicit error is safer than silently writing data shaped like the old schema. This is isolated and easy to drop if undesired.
- For the same consistency reason, the SQL layer rejects UPDATE / DELETE / MERGE INTO / INSERT OVERWRITE / TRUNCATE while a session-level time-travel selector (`SET 'paimon.scan.version'` / `'paimon.scan.timestamp-millis'`) is active, instead of silently ignoring the selector and writing to the latest state while reads in the same session resolve through the snapshot schema.
- The snapshot resolved by `copy_with_time_travel` is cached on the table copy and reused by `TableScan::resolve_snapshot`, so each scan of a time-travelled table doesn't re-read tag/snapshot files to resolve the same selector (invalidated whenever options change through `copy_with_options`).

Because the schema switch happens at the `Table` level, scan stats pruning, the per-file evolution target, and the provider's Arrow schema stay consistent automatically; the existing field-id based stats devolution is unaffected (files in an old snapshot always have `schema_id <=` the snapshot's schema id).

### Tests

- New unit tests in `table::time_travel` (multi-schema fixture built by persisting `schema-0`/`schema-1` and committing one snapshot per version): schema switch by version/tag/timestamp, no-selector no-op, silent fallback for invalid/conflicting selectors, merged-options replacement semantics, write rejection, and an end-to-end read asserting the old snapshot returns only the old columns.
- New `time_travel_schema_tests` in paimon-datafusion: `VERSION AS OF` (SQLContext path and the relation-planner path on a raw `SessionContext`), `TIMESTAMP AS OF`, `SET 'paimon.scan.version'` + SELECT/INSERT, and that selecting a later-added column at an old snapshot fails at planning.
- Existing time-travel tests (conflicting/invalid selector behavior) pass unchanged.

### API and Format

New public APIs: `Table::copy_with_time_travel`, `Table::is_time_traveled`, `TableSchema::copy_with_replaced_options`. No storage format change.

Scope notes (deliberate, follow-ups welcome):
- Selector coverage stays the existing Rust subset (`scan.version`, `scan.timestamp-millis`); Java additionally supports `scan.snapshot-id` / `scan.tag-name` / `scan.watermark` / `scan.timestamp`.
- `FileSystemCatalog::get_table` does not auto-travel selectors persisted in table options (Java `FileStoreTableFactory.create` does); only dynamic entry points are covered here.

### Documentation

None required.

### DIFF
--- a/crates/paimon/src/table/mod.rs
+++ b/crates/paimon/src/table/mod.rs
@@ -55,6 +55,7 @@ mod table_read;
 mod table_scan;
 pub(crate) mod table_write;
 mod tag_manager;
+pub(crate) mod time_travel;
 mod vector_search_builder;
 mod write_builder;
 
@@ -97,6 +98,9 @@ pub struct Table {
     schema: TableSchema,
     schema_manager: SchemaManager,
     rest_env: Option<RESTEnv>,
+    /// True when this table copy was switched to a historical schema by
+    /// [`Table::copy_with_time_travel`]. Such a copy is read-only.
+    time_traveled: bool,
 }
 
 impl Table {
@@ -116,6 +120,7 @@ impl Table {
             schema,
             schema_manager,
             rest_env,
+            time_traveled: false,
         }
     }
 
@@ -176,6 +181,11 @@ impl Table {
     }
 
     /// Create a copy of this table with extra options merged into the schema.
+    ///
+    /// This never switches the schema version; it corresponds to Java
+    /// `FileStoreTable.copyWithoutTimeTravel`. Use
+    /// [`Table::copy_with_time_travel`] when the options may select a
+    /// historical snapshot whose schema should be used for reading.
     pub fn copy_with_options(&self, extra: HashMap<String, String>) -> Self {
         Self {
             file_io: self.file_io.clone(),
@@ -184,7 +194,47 @@ impl Table {
             schema: self.schema.copy_with_options(extra),
             schema_manager: self.schema_manager.clone(),
             rest_env: self.rest_env.clone(),
+            time_traveled: self.time_traveled,
         }
+    }
+
+    /// Create a copy of this table with extra options merged in, switching to
+    /// the schema of the time-travelled snapshot when the merged options
+    /// select one.
+    ///
+    /// Mirrors Java `AbstractFileStoreTable.copy(dynamicOptions)` →
+    /// `tryTimeTravel`: if the merged options contain a time-travel selector
+    /// (`scan.version` / `scan.timestamp-millis`) that resolves to a snapshot,
+    /// the table's fields and keys come from that snapshot's schema while the
+    /// options stay the merged ones (Java `TableSchema.copy(newOptions)`).
+    /// Like Java, resolution failures fall back silently to the current
+    /// schema; an invalid selector still fails later at scan planning.
+    pub async fn copy_with_time_travel(&self, extra: HashMap<String, String>) -> Result<Self> {
+        let mut table = self.copy_with_options(extra);
+        let core_options = crate::spec::CoreOptions::new(table.schema.options());
+        // No selector configured: nothing to resolve, no IO.
+        match core_options.try_time_travel_selector() {
+            Ok(Some(_)) => {}
+            _ => return Ok(table),
+        }
+        if let Ok(Some(snapshot)) =
+            time_travel::travel_to_snapshot(&table.file_io, &table.location, table.schema.options())
+                .await
+        {
+            if snapshot.schema_id() != table.schema.id() {
+                let snapshot_schema = table.schema_manager.schema(snapshot.schema_id()).await?;
+                table.schema =
+                    snapshot_schema.copy_with_replaced_options(table.schema.options().clone());
+                table.time_traveled = true;
+            }
+        }
+        Ok(table)
+    }
+
+    /// Whether this table copy reads a historical snapshot with its
+    /// historical schema (see [`Table::copy_with_time_travel`]).
+    pub fn is_time_traveled(&self) -> bool {
+        self.time_traveled
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/paimon/src/table/write_builder.rs
```diff
@@ -80,6 +80,16 @@ impl<'a> WriteBuilder<'a> {
     /// For primary-key tables, sequence numbers are lazily scanned per partition
     /// when the first writer for that partition is created.
     pub fn new_write(&self) -> crate::Result<TableWrite> {
+        // A time-travelled table copy carries a historical schema; writing
+        // through it would silently produce data shaped like the old schema.
+        // Java avoids this structurally (write paths use copyWithoutTimeTravel);
+        // here the same table copy can serve both reads and writes, so reject
+        // explicitly. Commit-only flows (new_commit) stay untouched.
+        if self.table.is_time_traveled() {
```
QuakeWang: This guard is too narrow. `is_time_traveled()` is only true when the resolved snapshot uses a different schema id, so writes are still allowed when `scan.version` / `scan.timestamp-millis` resolves to a snapshot with the same schema id.

That leaves ordinary `INSERT` able to write the latest table while the session is reading a historical snapshot. A minimal case is: create two snapshots without schema evolution, `SET 'paimon.scan.version' = '1'`, then `INSERT INTO ...` succeeds.

Please reject writes whenever the table options contain a time-travel selector, not only when the table switched to a historical schema.

crates/paimon/src/table/mod.rs
```diff
@@ -184,7 +199,54 @@ impl Table {
             schema: self.schema.copy_with_options(extra),
             schema_manager: self.schema_manager.clone(),
             rest_env: self.rest_env.clone(),
+            time_traveled: self.time_traveled,
```
QuakeWang, line 209: `copy_with_options` keeps `time_traveled` and the current schema, but clears `travel_snapshot`. If this is called on a table already returned by `copy_with_time_travel`, the result can keep the old snapshot schema while resolving a different snapshot later during scan.

That can produce a mismatched table state: historical schema from one snapshot, scan snapshot from another selector/latest. Please either make this transition impossible, reset back to the base/latest schema, or re-resolve time travel when options change.

crates/paimon/src/table/write_builder.rs
TheR1sing3un: Fixed in 05bdc13. `new_write` now rejects whenever the table options contain a time-travel selector (`CoreOptions::try_time_travel_selector()` returning `Some` or a conflicting-selector error), independent of whether the schema was switched — matching the SQL-layer guard semantics. Covered by `test_copy_with_time_travel_same_schema_still_rejects_write` (your exact scenario: two snapshots without schema evolution) and a datafusion case asserting `SET 'paimon.scan.version'='2'` + INSERT fails.

crates/paimon/src/table/mod.rs
TheR1sing3un: Fixed in 05bdc13 by making the mismatched state fail loudly (your first option). `copy_with_options` now only invalidates the resolved snapshot when the merged options actually change the selector (`scan.version` / `scan.timestamp-millis`); unrelated option merges keep the snapshot/schema pair intact. If the selector is changed on a time-travelled copy, `TableScan::resolve_snapshot` returns an error directing to `copy_with_time_travel`, which re-resolves both the snapshot and the schema (re-traveling an already-travelled copy is supported and tested). Re-resolving inside `copy_with_options` itself wasn't viable since it is synchronous and resolution needs IO. Covered by `test_changing_selector_after_travel_fails_scan`.

JingsongLi (approved):
+1

PR #381, opened 2026-06-13
feat(scan): add pruning trace baselines

hhhizzz:
Part of https://github.com/apache/paimon-rust/issues/385

## What changed

Adds scan planning trace counters to Paimon table scans and exposes them through DataFusion physical plan display.

The trace records manifest list counts, partition/bucket/data-stats pruning counters, cross-schema stats pruning, data-evolution pruning counters, LIMIT split reduction, and final split/file counts.

## Why

This gives a self-contained way to verify metadata pruning behavior for partition pruning, bucket-key pruning, LIMIT, COUNT(*) statistics rewrite, SQL BETWEEN partition pruning, and time-travel snapshot selection.

## Validation

- `cargo test -p paimon test_plan_with_trace_records_between_data_stats_pruning --lib -- --nocapture`
- `cargo test -p paimon-datafusion --test scan_pruning_trace -- --nocapture`
- `git diff --check`


JingsongLi:
This branch no longer merges cleanly into the current `origin/main`. I checked locally with `git merge-tree --write-tree origin/main codex/review-pr-381`, and it reports a content conflict in `crates/paimon/src/table/table_scan.rs`. The trace tests themselves pass on the PR branch, but the PR needs a rebase/merge conflict resolution before it can be reviewed against current main.

crates/paimon/src/table/table_scan.rs
```diff
@@ -670,8 +780,11 @@ impl<'a> TableScan<'a> {
             };
 
         for ((partition, bucket), (total_buckets, data_files)) in groups {
-            let partition_row = BinaryRow::from_serialized_bytes(&partition)?;
-
+            let partition_row = if partition_computer.is_some() {
+                BinaryRow::from_serialized_bytes(&partition)?
+            } else {
+                BinaryRow::new(0)
```
JingsongLi: This changes the non-partitioned path from deserializing the manifest partition bytes to constructing a fresh `BinaryRow::new(0)`. That drops the backing data carried by the manifest entry. The existing integration test now fails with `Non-partitioned split should have backing data from manifest deserialization` (`paimon-integration-tests --test read_tables`, `test_read_log_table`) because the planned split no longer contains the manifest-deserialized empty partition row. Please keep using `BinaryRow::from_serialized_bytes(&partition)?` for all entries and only use `PartitionComputer` to build the bucket path when partition keys exist.

hhhizzz: Thanks for the review.

I updated the PR by merging the latest main and fixed the conflict in table scan planning. The non-partitioned path now keeps the manifest-deserialized partition row via BinaryRow::from_serialized_bytes(&partition)?, and PartitionComputer is only used for bucket pruning when partition keys exist.

CI is green now, including the integration job.

For scope: this PR (#381) covers the scan trace counters, DataFusion physical plan display, and the pruning baseline tests described in #385. The non-partition IN stats pruning part is handled separately by #382, so #385 should stay open until that PR is merged as well.

Could you please take another look when you have time?

JingsongLi:
Looks good to me!

JingsongLi (approved):

PR #382, opened 2026-06-13
fix(scan): prune IN predicates using file stats

### LINKED ISSUE #385: Improve scan pruning observability and IN predicate stats pruning
### Search before asking

## Motivation

Paimon Rust scan planning has several metadata pruning paths, including partition pruning, bucket pruning, file min/max stats pruning, LIMIT split reduction, COUNT(*) statistics rewrite, and time-travel snapshot selection.

Today these pruning decisions are hard to inspect from tests or physical plan output. This makes it difficult to identify cases that still do full scans or produce too many splits.

There is also a specific gap for non-partition `IN` predicates: file min/max stats can prove that some files cannot match, but `IN` currently fails open and keeps those files.

## Proposal

Add lightweight scan planning trace counters so tests and DataFusion physical plan display can show how many manifests, manifest entries, splits, and files survive each pruning stage.

Use the trace to add self-contained pruning baselines for:

- partition pruning
- bucket-key pruning
- SQL BETWEEN partition pruning
- LIMIT split reduction
- COUNT(*) statistics rewrite
- time-travel snapshot selection

Separately, improve non-partition `IN` stats pruning by checking whether any `IN` literal overlaps the file min/max range. Keep conservative behavior for `NOT IN`, missing stats, corrupt stats, and unsupported comparisons.

### Solution

- #381 adds scan pruning trace counters and baseline tests.
- #382 fixes non-partition `IN` stats pruning with file min/max stats.

### Anything else?

_No response_

### Willingness to contribute

hhhizzz:
## What changed

This adds file-stats pruning for non-partition `IN` predicates.

When a data file has value min/max stats, an `IN` predicate now keeps the file if at least one literal can fall inside the file range, and prunes the file only when every literal is outside `[min, max]`.

The pruning is deliberately scoped to data-file stats. Other stats consumers, such as partition manifest stats and reader-side row-group/page stats, continue to fail open for `IN` unless they explicitly opt in.

Conservative behavior is preserved:

- `NOT IN` still fails open.
- Missing, undecodable, inverted, or incomparable stats fail open.
- All-null data files are pruned for `IN`, matching normal comparison semantics.

## Why

Before this change, non-partition `IN` predicates skipped stats pruning completely. That could keep files whose min/max metadata already proves they cannot match, leading to unnecessary scan planning work and extra files/splits.

## Related issue

Fixes #385.

#381 already added the scan trace counters and pruning baselines from that issue; this PR completes the remaining non-partition `IN` stats-pruning part.

## Validation

- `cargo test -p paimon test_data_file_matches_in --lib`
- `cargo test -p paimon test_data_file_matches_not_in_fails_open --lib`
- `cargo test -p paimon predicate_stats::tests --lib`
- `cargo clippy -p paimon --lib --all-targets -- -D warnings`
- `cargo test -p paimon-datafusion --test scan_pruning_trace -- --nocapture`

Manual trace check for `id IN (15)` over three files with stats ranges `[1, 2]`, `[10, 20]`, and `[100, 101]`:

- `manifest_entries_read = 3`
- `manifest_entries_pruned_by_data_stats = 2`
- `manifest_entries_after_manifest_filters = 1`
- `final_files = 1`

### DIFF
--- a/crates/paimon/src/predicate_stats.rs
+++ b/crates/paimon/src/predicate_stats.rs
@@ -59,7 +59,8 @@ pub(crate) fn data_leaf_may_match<T: StatsAccessor>(
         PredicateOperator::IsNotNull => {
             return all_null != Some(true);
         }
-        PredicateOperator::In | PredicateOperator::NotIn => {
+        PredicateOperator::In => {}
+        PredicateOperator::NotIn => {
             return true;
         }
         PredicateOperator::EndsWith | PredicateOperator::Contains => {
@@ -113,6 +114,10 @@ pub(crate) fn data_leaf_may_match<T: StatsAccessor>(
     };
 
     match op {
+        PredicateOperator::In => literals.iter().any(|literal| {
+            !matches!(literal.partial_cmp(&min_value), Some(Ordering::Less))
+                && !matches!(literal.partial_cmp(&max_value), Some(Ordering::Greater))
+        }),
         PredicateOperator::Eq => {
             !matches!(literal.partial_cmp(&min_value), Some(Ordering::Less))
                 && !matches!(literal.partial_cmp(&max_value), Some(Ordering::Greater))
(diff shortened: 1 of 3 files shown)


QuakeWang (approved):
+1

Issue #378, opened 2026-06-11
feat: Support Mosaic file format in paimon-rust


QuakeWang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Motivation

Apache Paimon already supports the Mosaic file format as a wide-table optimized data format. Mosaic is a columnar-bucket hybrid format that groups columns into deterministic buckets and compresses each bucket independently, reducing read amplification for very wide tables when queries project only a small subset of columns.

Currently, `paimon-rust` dispatches data file readers by file extension and supports formats such as Parquet, ORC, Avro, Blob, and feature-gated Vortex. However, `.mosaic` files are not recognized. As a result, Rust clients cannot read Paimon tables whose `file.format` is set to `mosaic`, even if those tables were written by Java Paimon or PyPaimon.

This matters for compatibility with the broader Paimon ecosystem, especially for wide-table workloads where Mosaic is the recommended format.

### Solution

Add feature-gated support for reading Mosaic data files in `paimon-rust`, with the implementation split into small reviewable steps.

Proposed phases:

1. **Reader foundation**

   Add an optional `mosaic` feature, depend on `paimon-mosaic-core` behind that feature, and introduce a `MosaicFormatReader` wired into the existing file-format dispatch for `.mosaic` files.

   This phase should provide basic Arrow `RecordBatch` reading and projection support.

2. **Paimon read-path correctness**

   Integrate the Mosaic reader with the existing table read semantics, including schema evolution, missing-column null filling, projection order, deletion vectors, and row-range selection.

   This phase should add table-level and mixed-format tests to ensure Mosaic files behave consistently with other supported formats.

3. **Documentation and compatibility polish**

   Document the feature flag, supported scope, and current limitations. The initial scope should be read-only and focused on compatibility with Mosaic files written by Java Paimon, PyPaimon, or `paimon-mosaic-core`.

Out of scope for the first phase:

- Writing `.mosaic` data files from `paimon-rust`.
- Emitting Mosaic row-group statistics into `DataFileMeta.value_stats`.
- Making Mosaic the default file format.
- Implementing Mosaic bloom filter support.
- Changing the Mosaic storage format.

Follow-up work can add writer support, stats integration, and performance benchmarks once read compatibility is stable.

### Anything else?

Relevant context:

- `paimon-rust` already has a format abstraction through `FormatFileReader` and `FormatFileWriter`.
[...]

PR #389, opened 2026-06-16
fix: support missing columns in Mosaic reads

QuakeWang:
### Purpose

Linked issue: #378

Mosaic reads used to validate the full requested schema before checking the physical Mosaic file schema. When a requested column was missing from an older Mosaic file, an unsupported type in that missing column could fail the read before `DataFileReader` had a chance to fill it with nulls.

### Brief change log

- Filter Mosaic projections by the physical file schema before validating supported Mosaic types.
- Return empty batches with the correct row count when all requested Mosaic columns are physically missing.
- Keep missing column null filling in `DataFileReader`.
- Add Mosaic reader tests for missing columns, missing unsupported columns, all-missing projections, and row selection.
- Add a `DataFileReader` test covering null fill for a physically missing Mosaic column.

### Tests

- `cargo test -p paimon --features mosaic arrow::format::mosaic -- --nocapture`
- `cargo test -p paimon --features mosaic table::data_file_reader -- --nocapture`
- `cargo check -p paimon-datafusion --features mosaic`
- `cargo clippy -p paimon --features mosaic --tests -- -D warnings`

### DIFF
--- a/crates/paimon/src/arrow/format/mosaic.rs
+++ b/crates/paimon/src/arrow/format/mosaic.rs
@@ -28,6 +28,7 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use futures::StreamExt;
 use paimon_mosaic_core::reader::{InputFile, MosaicReader, ReaderAccess};
+use std::collections::HashSet;
 use std::io;
 
 pub(crate) struct MosaicFormatReader;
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #391, opened 2026-06-16
test(datafusion): cover SQL schema evolution reads

liujiwen-up:
### Purpose

Linked issue:  #258

Add DataFusion SQL coverage for schema evolution tables generated from Java/Spark consistency fixtures, focusing on mixed-format reads through SQL. The rename coverage uses `schema_evolution_rename_column`, whose fixture contains parquet/orc/avro files.

### Brief change log

- Add a small `assert_sql` helper for table-formatted SQL result assertions.
- Cover SQL reads for mixed-format add column and type promotion tables.
- Cover SQL reads for rename, drop, and reorder/move schema evolution cases with `SELECT`, `WHERE`, and `ORDER BY id`.

### Tests

- `cargo fmt --check`: passed
- `git diff --check`: passed
- `cargo test -p paimon-datafusion --test read_tables schema_evolution --no-run`: passed
- `cargo test -p paimon-datafusion --test read_tables schema_evolution -- --nocapture`: not fully runnable locally because `/tmp/paimon-warehouse` has not been provisioned and Docker daemon is unavailable

### API and Format

### Documentation

### DIFF
--- a/crates/integrations/datafusion/tests/read_tables.rs
+++ b/crates/integrations/datafusion/tests/read_tables.rs
@@ -19,7 +19,8 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use datafusion::arrow::array::{Array, Int32Array, StringArray};
-use datafusion::arrow::util::pretty::pretty_format_batches;
+use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::arrow::util::display::array_value_to_string;
 use datafusion::catalog::CatalogProvider;
 use datafusion::datasource::TableProvider;
 use datafusion::logical_expr::{col, lit, TableProviderFilterPushDown};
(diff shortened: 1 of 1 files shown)


crates/integrations/datafusion/tests/read_tables.rs
```diff
@@ -98,6 +99,15 @@ async fn collect_query(
     ctx.sql(sql).await?.collect().await
 }
 
+async fn assert_sql(sql: &str, expected: &[&str]) {
```
QuakeWang: Nit: this helper compares the full `pretty_format_batches` output verbatim. It keeps the expected SQL result readable, but also couples the tests to Arrow/DataFusion pretty-printer details such as padding, borders, and NULL rendering. Fine for these small result sets; if we want less brittle tests across dependency upgrades, comparing normalized rows/columns would be more stable.

liujiwen-up: thanks,i done

QuakeWang (approved):
+1

PR #393, opened 2026-06-16
feat: add Mosaic row-group predicate pruning

QuakeWang:
### Purpose

Linked issue: Related to #378

Mosaic reads previously treated data predicates as residual only, so row groups were read even when Mosaic row-group statistics could prove they could not match.

This PR adds conservative row-group predicate pruning for Mosaic files while preserving residual filtering semantics.

### Brief change log

  - Add a Mosaic `StatsAccessor` over row-group `ColumnStats`.
  - Map file-level Paimon fields to Mosaic column indices by name.
  - Convert supported Mosaic stats values into Paimon `Datum` for shared stats pruning.
  - Fail open when stats are missing or cannot be interpreted safely.
  - Update `ReadBuilder` docs to describe generic format-level reader pruning.

### Tests

  - `cargo test -p paimon --features mosaic arrow::format::mosaic::tests -- --nocapture`
  - `cargo fmt --check`

### DIFF
--- a/crates/paimon/src/arrow/format/mosaic.rs
+++ b/crates/paimon/src/arrow/format/mosaic.rs
@@ -17,8 +17,9 @@
 
 use super::{FilePredicates, FormatFileReader};
 use crate::arrow::build_target_arrow_schema;
+use crate::arrow::filtering::{predicates_may_match_with_schema, StatsAccessor};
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #394, opened 2026-06-18
feat(spec): implement Display for DataFileMeta

Aitozi:
## What changed

Implemented `Display` for `DataFileMeta` and added a focused unit test covering all fields currently present on the Rust struct.

## Why

`DataFileMeta` previously used `todo!()` in its `Display` implementation, so formatting it would panic at runtime.

## CI fix

The first CI run failed because the test asserted the complete `BinaryTableStats::empty()` debug string. Under CI full-feature builds, empty row bytes are represented with a different length than the local default build. The test now checks the displayed field fragments instead of internal serialized byte details.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo test -p paimon display_includes_all_data_file_meta_fields --features fulltext,vortex,mosaic`

### DIFF
--- a/crates/paimon/src/spec/data_file.rs
+++ b/crates/paimon/src/spec/data_file.rs
@@ -109,8 +109,35 @@ pub struct DataFileMeta {
 }
 
 impl Display for DataFileMeta {
-    fn fmt(&self, _: &mut Formatter<'_>) -> std::fmt::Result {
-        todo!()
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
LGTM

JingsongLi (approved):
+1

PR #395, opened 2026-06-18
fix: preserve snapshot properties

Aitozi:
## Summary
- Preserve the Java snapshot `properties` map in Rust snapshot metadata.
- Add fixture coverage for deserializing and round-tripping non-empty properties.

## Tests
- `cargo fmt --check`
- `cargo test -p paimon spec::snapshot::tests::test_snapshot_serialization_deserialization`
- `git diff --check`


QuakeWang (approved):
+1

JingsongLi (approved):
+1

PR #396, opened 2026-06-18
feat(table): support mod and hive bucket functions

Aitozi:
### Purpose

Support the remaining fixed-bucket function types in Rust table writes and scan pruning:

- `bucket-function.type=mod` for single-column INT/BIGINT bucket keys using floor-mod semantics.
- `bucket-function.type=hive` using Hive-compatible hash semantics for fixed bucket keys.
- Reuse one bucket calculation path for write routing and bucket predicate pruning so both sides stay aligned.

### Tests

- `git diff --check`
- `cargo test -p paimon bucket_function --lib`
- `cargo test -p paimon test_compute_target_buckets_ --lib`
- `cargo test -p paimon-integration-tests test_fixed_bucket_scan_filters_by_bucket --test append_tables`
- `cargo check -p paimon -p paimon-integration-tests -p paimon-datafusion`

Note: `cargo check --workspace` is blocked locally by the Python binding environment because the active interpreter is Python 3.9 while `pyo3` is configured with `abi3-py310`.

### DIFF
--- a/crates/paimon/src/spec/binary_row.rs
+++ b/crates/paimon/src/spec/binary_row.rs
@@ -1131,7 +1131,7 @@ fn write_typed_value(
 
 /// Build BinaryRows for all rows in the batch for the given field indices.
 /// Downcasts columns once, then iterates rows — O(F) downcasts instead of O(N*F).
-fn batch_build_binary_rows(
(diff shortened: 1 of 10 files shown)


JingsongLi (approved):
+1

PR #399, opened 2026-06-19
feat: support vindex vector search

JingsongLi:
## Summary

Add read-side support for Java `paimon-vector` vindex files through `paimon-vindex-core` 0.1.0. The vector search path now treats vindex as a peer backend to Lumina instead of routing it through the Lumina module.

## Changes

- Add a top-level `vindex` module with `ivf-flat`, `ivf-pq`, `ivf-hnsw-flat`, and `ivf-hnsw-sq` index type detection plus a `paimon-vindex-core` backed global index reader.
- Move shared vector search request/result IO types into a neutral `vector_search` module so Lumina and vindex readers can share them without coupling vindex to Lumina.
- Update `VectorSearchBuilder` to dispatch matching global index files to either the Lumina reader or the vindex reader based on the index type.
- Add a DataFusion fixture produced by Java vindex and verify Rust vector search can read and search it.

## Testing

- [x] `cargo fmt`
- [x] `cargo check -p paimon`
- [x] `cargo test -p paimon vector_search --lib`
- [x] `cargo test -p paimon vindex --lib`
- [x] `cargo test -p paimon lumina --lib`
- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_java_vindex_table`
- [x] `git diff --check`


leaves12138 (changes requested):
Thanks for adding vindex vector search support. I reviewed the vindex reader, backend dispatch, Java-generated vindex table coverage, and verified the focused vector/vindex tests locally. CI is also green now.

I found one correctness issue that should be fixed before merging: `paimon-vindex-core` represents `InnerProduct` search distance as the negative inner product (`fvec_distance` returns `-fvec_inner_product`) and sorts lower distances as better. This PR currently converts `MetricType::InnerProduct` by returning the raw distance as the score. Because `SearchResult::top_k` later keeps larger scores when merging multiple index shards, this reverses the ordering for InnerProduct across shards (and also in `collect_results` if it ever needs to trim padded/filtered results). Please convert InnerProduct distance back to similarity, e.g. `-distance`, and add a small test that `[-0.9, -0.5, -0.1]` keeps the `0.9`/`0.5` inner-product matches rather than the smaller similarities.

Non-blocking nit: the new test tarball contains macOS AppleDouble `._*` entries; it would be cleaner to regenerate it without those entries if convenient.

leaves12138 (approved):
Thanks for the update. I re-reviewed the latest head and the previously requested InnerProduct issue is fixed now: vindex distances are converted back to similarity with `-distance`, and the new unit test covers the negative-distance ordering case.

I also checked that the regenerated vindex test tarball no longer contains macOS AppleDouble entries. Focused local validation passed for `paimon vindex`, `paimon vector_search`, and the vindex DataFusion table test. The remaining CI jobs are still running at the time of this review, but I do not see any remaining code-level blocker.

PR #402, opened 2026-06-22
fix(vortex): avoid footer read deadlock

QuakeWang:
### Purpose

Vortex 0.68 can hang while decoding file footers because `session.layouts().registry()` keeps a session borrow alive while `session.allows_unknown()` may lazily insert default session state. That can require a write lock on the same session map and deadlock.

### Brief change log

  - Upgrade Vortex from 0.68 to 0.75.0, where `allows_unknown()` no longer lazily writes session state.
  - Adapt Vortex-to-Arrow conversion to the Vortex session execution API.
  - Keep `constant_time_eq` below 0.5.0 so fresh dependency resolution remains compatible with Rust 1.91.

### Tests

  - `RUSTUP_TOOLCHAIN=1.91.0 cargo fmt --all -- --check`
  - `git diff --check`
  - `RUSTUP_TOOLCHAIN=1.91.0 cargo clippy --all-targets --workspace --features fulltext,vortex,mosaic -- -D warnings`
  - `RUSTUP_TOOLCHAIN=1.91.0 cargo build --features fulltext,vortex,mosaic`
  - `RUSTUP_TOOLCHAIN=1.91.0 cargo test -p paimon --all-targets --features fulltext,vortex,mosaic`
  - `timeout 120s taskset -c 0 env RUSTUP_TOOLCHAIN=1.91.0 cargo test -p paimon-datafusion --features vortex --test vortex_tables -- --nocapture`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ edition = "2021"
 homepage = "https://paimon.apache.org/docs/rust/"
 repository = "https://github.com/apache/paimon-rust"
 license = "Apache-2.0"
-rust-version = "1.86.0"
+rust-version = "1.91.0"
 
 [workspace.dependencies]
 arrow = "58.0"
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #408, opened 2026-06-25
[btree] Add support for prefix range scan on btree global index

ArnavBalyan:
### Purpose
 - Prefix lookup currently has no direct path and returns every non null row, every prefix query reads the whole segment.
 - Add `query_prefix` on the btree reader, which resolves a prefix on range [prefix, prefix_successor(prefix)).

### Tests
 - UT added that covers multi match, disjoint and empty prefix.


JingsongLi (approved):
+1

Issue #257, opened 2026-04-17
Support agg merge engine


JingsongLi:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Description

For pk table with fixed bucket and -1 bucket.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #409, opened 2026-06-25
fix: reject aggregation on sequence fields

QuakeWang:
### Purpose

Linked issue: #257 

Rust aggregation schema validation allowed `fields.<sequence-field>.aggregate-function`, but the runtime ignores that option because sequence fields are forced to `last_value`. Java rejects this metadata during schema validation, so Rust could create table options that are not accepted by Java validation.

### Brief change log

  - Reject per-field aggregate functions on `sequence.field` columns during aggregation schema validation.
  - Keep runtime sequence-field merging unchanged: sequence fields still use `last_value` for existing or external metadata.
  - Add unit and DataFusion SQL coverage for create/alter validation and ignored primary-key functions.
  - Document the sequence-field aggregation restriction.

### Tests

  - `cargo test -p paimon --lib aggregation`
  - `cargo test -p paimon --lib schema`
  - `cargo test -p paimon-datafusion --test pk_tables aggregation`

### API and Format

No storage format change. This tightens validation for invalid aggregation table options.

### Documentation

Updated SQL docs for sequence-field aggregation behavior.

### DIFF
--- a/crates/integrations/datafusion/tests/pk_tables.rs
+++ b/crates/integrations/datafusion/tests/pk_tables.rs
@@ -2514,8 +2514,8 @@ async fn test_pk_aggregation_mixed_aggregators() {
     assert_eq!(first_seen.value(0), "a"); // first non-null wins
 }
 
-/// `sequence.field` forces the named column to `last_value`, even when the
-/// user explicitly configures another aggregator for it.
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

Issue #410, opened 2026-06-25
Support reading Paimon VECTOR type in paimon-rust


JunRuiLee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.
## Motivation

Apache Paimon has a first-class `VECTOR` type (`VECTOR<element, length>`), and
tables written by Java/Python Paimon can contain `VECTOR` columns. This issue
tracks giving paimon-rust the ability to **read** `VECTOR` columns and use them
as input to **vector (ANN) search**, matching Java/Python Paimon on the read and
query side.

In Paimon, `VectorType` is a general-purpose column type (it can be read,
projected, and written like any other column); it also happens to be the natural
input to vector search. This issue is scoped to the **read + search** path.
**Writing** `VECTOR` columns (Arrow write, per-format writers, dedicated
`.vector.` files, Parquet stats) is a separate, larger effort tracked in its own
issue: #<WRITE_ISSUE>.

## Scope (incremental PRs)

Read foundation:

- [x] **PR 1** — Add native `VectorType` to the type system (variant,
  validation, JSON serde, `Display`). (#411)
- [x] **PR 2** — Arrow conversion (`VectorType` ↔ `FixedSizeList`) + read vector
  columns inlined in ordinary data files. (#412)
- [x] **PR 3** — Read dedicated `.vector.` files (parquet/vortex). (#423)

Vector (ANN) search:

- [x] **PR 4** — Accept `VECTOR<FLOAT>` in the Lumina index build path (derive
  the dimension from `VectorType::length` instead of an explicit `dimension`
  option), matching Java where both `ARRAY<FLOAT>` and `VECTOR<FLOAT>` are valid
  vector-index columns. (#440)

Once PR 4 merges, reading and vector-searching `VECTOR` columns is complete and
this issue can be closed.

## Out of scope (tracked separately)

- **Writing `VECTOR` columns** — inline + dedicated `.vector.` files + Parquet
  value-stats. 
- **vindex index build (writer)** — Rust `src/vindex/` currently has only a
  reader; porting the native vindex writer (which will accept both `ARRAY<FLOAT>`
  and `VECTOR<FLOAT>` internally) is separate follow-up work.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #411, opened 2026-06-25
[type] Add native VectorType to the type system (#410 PR 1)

JunRuiLee:
**PR 1** of reading Paimon `VECTOR` columns in paimon-rust, tracked in #410. Adds the type-system foundation; reading vector *data* is deferred to later PRs.

Roadmap (#410): **PR 1 (this)** type system → PR 2 Arrow conversion + inlined reads → PR 3 dedicated `.vector.` files.

## Changes

- New `VectorType` mirroring upstream `org.apache.paimon.types.VectorType`: `u32` length in `[1, i32::MAX]`, element type restricted to `BOOLEAN/TINYINT/SMALLINT/INT/BIGINT/FLOAT/DOUBLE`; JSON serde matching the Java wire shape; `Display` as `VECTOR<FLOAT, 128>`.
- New `DataType::Vector` variant threaded through all exhaustive matches: type-system semantics (`is_nullable`, `copy_with_nullable`, `contains_row_type`) implemented; IO/integration sites return explicit `Unsupported`; BTree key comparator rejects `VECTOR` rather than falling back to byte comparison.

After this PR a `VECTOR` column deserializes/validates/serializes/renders correctly; vector data IO returns `Unsupported` — the intended PR 1 boundary.

## Testing

`cargo test -p paimon --lib` 895 passed; `cargo clippy -p paimon --all-targets` clean; `cargo check` passes for `paimon-datafusion` and `paimon-c`.

### DIFF
--- a/crates/paimon/src/btree/key_serde.rs
+++ b/crates/paimon/src/btree/key_serde.rs
@@ -109,12 +109,6 @@ pub fn make_key_comparator(data_type: &DataType) -> KeyComparator {
             let bv = i64::from_le_bytes(b[..8].try_into().unwrap());
             av.cmp(&bv)
         }),
-        // VECTOR is not a valid key type. Fail loudly rather than letting it fall
-        // into the lexicographic byte comparator below, which would silently sort
-        // vectors incorrectly.
-        DataType::Vector(_) => {
-            panic!("VECTOR columns are not supported as BTree keys")
-        }
         // String, VarChar, Char, Bytes — lexicographic
         _ => Box::new(|a: &[u8], b: &[u8]| a.cmp(b)),
     }
--- a/crates/paimon/src/btree/key_serde.rs
+++ b/crates/paimon/src/btree/key_serde.rs
@@ -109,6 +109,12 @@ pub fn make_key_comparator(data_type: &DataType) -> KeyComparator {
             let bv = i64::from_le_bytes(b[..8].try_into().unwrap());
             av.cmp(&bv)
         }),
+        // VECTOR is not a valid key type. Fail loudly rather than letting it fall
+        // into the lexicographic byte comparator below, which would silently sort
+        // vectors incorrectly.
+        DataType::Vector(_) => {
+            panic!("VECTOR columns are not supported as BTree keys")
+        }
         // String, VarChar, Char, Bytes — lexicographic
         _ => Box::new(|a: &[u8], b: &[u8]| a.cmp(b)),
     }
--- a/crates/paimon/src/spec/types.rs
+++ b/crates/paimon/src/spec/types.rs
@@ -103,6 +103,8 @@ pub enum DataType {
     /// Data type of a sequence of fields. A field consists of a field name, field type, and an optional
     /// description.
     Row(RowType),
+    /// Data type of a fixed-size dense vector `VECTOR<element, length>`.
+    Vector(VectorType),
 }
 
 impl DataType {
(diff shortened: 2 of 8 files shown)


crates/paimon/src/spec/types.rs
```diff
@@ -241,6 +250,196 @@ impl ArrayType {
...
+            other => {
+                unreachable!("vector element type validated at construction: {other:?}")
+            }
+        }
+    }
+}
+
+impl Display for VectorType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "VECTOR<{}, {}>", self.element_sql_name(), self.length)?;
```
QuakeWang: `Display` drops the element type nullability. For example, a `VECTOR<FLOAT NOT NULL, 4>`-like type would render as `VECTOR<FLOAT, 4>`, while serde preserves the element nullability. Please either render `element_type` with its SQL string or reject/normalize non-null vector element types.

JunRuiLee: Fixed — Display now renders the element's nullability too (e.g. VECTOR<FLOAT NOT NULL, 4>), matching elementType.asSQLString() and the serde form.

crates/paimon/src/btree/key_serde.rs
```diff
@@ -109,6 +109,12 @@ pub fn make_key_comparator(data_type: &DataType) -> KeyComparator {
             let bv = i64::from_le_bytes(b[..8].try_into().unwrap());
             av.cmp(&bv)
         }),
+        // VECTOR is not a valid key type. Fail loudly rather than letting it fall
+        // into the lexicographic byte comparator below, which would silently sort
+        // vectors incorrectly.
+        DataType::Vector(_) => {
+            panic!("VECTOR columns are not supported as BTree keys")
```
QuakeWang: `VECTOR` can still pass schema key validation and reach this comparator as a primary/key-index type, so this turns an unsupported schema into a runtime panic. Please reject `VECTOR` earlier with a normal `Unsupported`/config error, or change comparator creation to return `Result` instead of panicking.

crates/paimon/src/btree/key_serde.rs
```diff
@@ -109,6 +109,12 @@ pub fn make_key_comparator(data_type: &DataType) -> KeyComparator {
             let bv = i64::from_le_bytes(b[..8].try_into().unwrap());
             av.cmp(&bv)
         }),
+        // VECTOR is not a valid key type. Fail loudly rather than letting it fall
+        // into the lexicographic byte comparator below, which would silently sort
+        // vectors incorrectly.
+        DataType::Vector(_) => {
```
JingsongLi: Schema validation still allows a VECTOR column to be used as a primary key, but this path later panics when the global index/BTree code asks for a key comparator. For example, `Schema::builder().column("embedding", DataType::Vector(...)).primary_key(["embedding"]).build()` succeeds and normalizes the field to NOT NULL, then `make_key_comparator` reaches this `panic!`. Could we reject VECTOR primary keys during schema validation, or change the comparator construction to return `Unsupported` instead of panicking?

JunRuiLee: Fixed in [bb36f89](https://github.com/apache/paimon-rust/pull/411/commits/bb36f895e7433b36b6b12154df31ce02c2f086d6) — reject VECTOR as primary key / explicit bucket-key at schema validation (create + alter), and removed the comparator panic.

crates/paimon/src/btree/key_serde.rs
JunRuiLee: Same as above, fixed in [bb36f89](https://github.com/apache/paimon-rust/pull/411/commits/bb36f895e7433b36b6b12154df31ce02c2f086d6)

JingsongLi (approved):
+1

PR #412, opened 2026-06-25
[arrow] Read inline VECTOR columns via Arrow FixedSizeList

JunRuiLee:
**PR 2** of reading Paimon `VECTOR` columns in paimon-rust, tracked in #410. Enables reading vector columns stored inline in ordinary parquet data files.

> Stacked on **#411 (PR 1)** — depends on the `VectorType` / `DataType::Vector` introduced there. Review/merge after #411. The diff shown here includes #411's commits until it merges.

Roadmap (#410): PR 1 type system (#411) → **PR 2 (this)** Arrow conversion + parquet inline read → PR 3 dedicated `.vector.` files.

## Changes

- `VectorType` ↔ Arrow `FixedSizeList` conversion in `arrow/mod.rs` (both directions), replacing PR 1's `Unsupported`: forward `Vector` → `FixedSizeList(Field("element", elem, element_nullable), length)`; reverse guards non-positive sizes with `u32::try_from` and routes through `VectorType::try_new` (rejecting length 0 and invalid element types).
- The read path builds the target Arrow schema from `paimon_type_to_arrow`, so an inline `FixedSizeList` vector column now materializes directly.

> **Mapping note:** VECTOR is represented as an Arrow `FixedSizeList<element, length>`, matching Java `ArrowFieldTypeConversion` (`VectorType → ArrowType.FixedSizeList(length)`). The reverse direction maps any `FixedSizeList` to a Paimon VECTOR — this is the only valid correspondence, since Paimon has no separate fixed-size-array type, and non-vector element types are rejected by `VectorType::try_new`.

Unchanged boundaries (vectors still not filter-pushable / not cast / not partition or BTree keys / no `binary_row` datum); writing and dedicated `.vector.` files remain out of scope.

## Testing

`cargo test -p paimon --lib` 901 passed; `cargo fmt --all -- --check` clean; `cargo clippy -p paimon --all-targets --features fulltext,vortex,mosaic -- -D warnings` clean. Adds 5 Arrow conversion unit tests + an end-to-end parquet test reading a `FixedSizeList<Float32>` column with a null row (downcast asserts `value_length`, child values, null bitmap).

### DIFF
--- a/crates/paimon/src/arrow/format/mod.rs
+++ b/crates/paimon/src/arrow/format/mod.rs
@@ -25,6 +25,9 @@ mod row;
 #[cfg(feature = "vortex")]
 mod vortex;
 
+#[cfg(test)]
+pub(crate) use parquet::ParquetFormatWriter;
+
 use crate::io::{FileRead, OutputFile};
 use crate::spec::{DataField, Predicate};
 use crate::table::{ArrowRecordBatchStream, RowRange};
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -1410,4 +1410,65 @@ mod tests {
         let total_rows: usize = reader.into_iter().map(|r| r.unwrap().num_rows()).sum();
         assert_eq!(total_rows, 5);
     }
+
+    #[tokio::test]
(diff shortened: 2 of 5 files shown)


JingsongLi (approved):
+1

PR #416, opened 2026-06-27
Add Java-compatible row format and sidecar path support

JingsongLi:
## Summary

Adds Java-compatible `.row` file format support to the Rust reader/writer path and aligns data-file sidecar path resolution with Java.

## Changes

- Add a `.row` format reader/writer with Java-compatible block layout, footer/index encoding, zstd block compression, primitive/nested type encoding, row selection, and bounded concurrent block prefetching.
- Register `.row` in the format reader/writer factory and make table writes produce/read `.row` files when `file.format=row` is configured.
- Preserve `.row` projection correctness by decoding with the full file schema before applying the existing table-level projection.
- Align data-file sidecar path handling with Java `DataFilePathFactory` / `DataFileMeta`: prefer `_EXTERNAL_PATH` for data files, align `extra_files` beside the data file, expose the `.index` sidecar naming helper, and collect data file plus sidecars consistently.
- Add Java fixture compatibility tests, Rust writer payload alignment checks, projection tests, external-path sidecar tests, and table-write roundtrip coverage.

## Testing

- [x] `cargo test -p paimon arrow::format::row --lib`
- [x] `cargo check -p paimon`
- [x] `cargo test -p paimon data_file_paths --lib`
- [x] `cargo test -p paimon test_extra_file_resolver_uses_external_path_parent --lib`
- [x] `cargo test -p paimon test_referenced_files_stats_external_sidecar_from_parent --lib`

## Notes

The `.row` reader uses bounded async range reads for blocks, similar to existing parquet/blob I/O patterns, while preserving output order.

Java writes independent `.index` sidecars only when file-index writer options are enabled. Rust currently has the file-index format codec but not the Java `FileIndexOptions` / `DataFileIndexWriter` layer, so this PR aligns the sidecar metadata/path mechanism without creating empty index sidecars.

### DIFF
--- a/crates/paimon/src/arrow/format/row.rs
+++ b/crates/paimon/src/arrow/format/row.rs
@@ -1,0 +1,2827 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Java-compatible `.row` file format.
+//!
+//! Layout reference:
+//! `org.apache.paimon.format.row.RowFormatWriter` in paimon-java.
+
+use super::{FilePredicates, FormatFileReader, FormatFileWriter};
+use crate::arrow::{arrow_to_paimon_type, build_target_arrow_schema, paimon_type_to_arrow};
+use crate::io::{FileRead, FileWrite, OutputFile};
+use crate::spec::{DataField, DataType, IntType};
+use crate::table::{ArrowRecordBatchStream, RowRange};
+use crate::Error;
+use arrow_array::builder::{
+    BinaryBuilder, BooleanBuilder, Date32Builder, Decimal128Builder, Float32Builder,
+    Float64Builder, Int16Builder, Int32Builder, Int64Builder, Int8Builder, StringBuilder,
+    Time32MillisecondBuilder, TimestampMicrosecondBuilder, TimestampMillisecondBuilder,
+    TimestampNanosecondBuilder,
+};
+use arrow_array::{
+    Array, ArrayRef, BinaryArray, BooleanArray, Date32Array, Decimal128Array, Float32Array,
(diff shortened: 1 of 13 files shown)


crates/paimon/src/arrow/format/row.rs
```diff
@@ -0,0 +1,2827 @@
...
+        output: &OutputFile,
+        schema: SchemaRef,
+        zstd_level: i32,
+    ) -> crate::Result<Self> {
+        let row_type = schema
+            .fields()
+            .iter()
+            .enumerate()
+            .map(|(idx, field)| {
+                let data_type = arrow_to_paimon_type(field.data_type(), field.is_nullable())?;
```
QuakeWang: This derives the .row physical layout from the incoming Arrow batch schema, but the file is committed with the table schema id. If the batch Arrow schema differs semantically from the table schema, e.g. table DECIMAL(20,2) but batch Decimal128(18,2), the writer encodes compact i64 while the reader decodes using the table schema as varlen decimal. Please derive the row type from the table/file DataField schema, or validate/cast the batch schema before opening the writer.

crates/paimon/src/arrow/format/row.rs
```diff
@@ -0,0 +1,2827 @@
...
+        let block_compressed_sizes = read_index_array(bytes, &mut pos)?;
+        let block_uncompressed_sizes = read_index_array(bytes, &mut pos)?;
+        let block_row_starts = read_index_array(bytes, &mut pos)?;
+        if pos != bytes.len() {
+            return Err(Error::DataInvalid {
+                message: format!(".row block index has {} trailing bytes", bytes.len() - pos),
+                source: None,
+            });
+        }
+        if block_compressed_sizes.len() != block_uncompressed_sizes.len()
```
QuakeWang, line 1738: After decoding the block index, we should validate sizes and row starts before using them for reads/decompression. Negative compressed/uncompressed sizes or non-monotonic offsets can currently be cast to huge u64/usize ranges in read_row_block. Please reject invalid index values with DataInvalid and use checked arithmetic for block ranges.

leaves12138 (approved):
Reviewed the row-format implementation and sidecar path alignment. The Java compatibility coverage, full-schema decode before projection, and external sidecar path handling look good to me. I also ran focused local tests for the row format, row projection, table-write roundtrip, and external sidecar path cases; they passed.

crates/paimon/src/arrow/format/row.rs
JingsongLi: Fixed in `28d0849`: the `.row` writer now receives the table/file `DataField` schema through `DataFileWriter` / `create_format_writer` and validates the Arrow batch schema against it before writing. This rejects semantic mismatches such as table `DECIMAL(20,2)` vs batch `Decimal128(18,2)` instead of deriving the physical layout from Arrow alone.

crates/paimon/src/arrow/format/row.rs
JingsongLi: Fixed in `28d0849`: `RowBlockIndex::from_bytes` now rejects negative sizes and non-monotonic row starts, offset computation uses checked arithmetic, reader validation ensures block ranges stay before the index section and row starts are valid for the footer row count, and `read_row_block` uses checked conversions/range arithmetic.

Issue #414, opened 2026-06-26
Expose DataFrame-style write API (WriteBuilder / TableWrite / TableCommit) to Python


JunRuiLee:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Motivation
  Companion to the read-path effort (#413).

  PyPaimon's write path is still pure Python, while the Rust core already
  implements the full append-write pipeline:

  `WriteBuilder → TableWrite.write_arrow() → prepare_commit() → TableCommit.commit()`

  This pipeline is already used by the SQL `INSERT INTO` path. It is simply not
  exposed through `bindings/python`.

  Goal: expose the existing Rust write API to Python so PyPaimon's batch write
  path can run on Rust, with the public Python API unchanged.

  Benefits:

  - **Faster batch write** — use the Rust write pipeline instead of the pure
    Python implementation.
  - **Consistent semantics** — SQL insert and DataFrame batch write can share the
    same Rust core path.
  - **Distributed write foundation** — once commit messages are serializable,
    workers can write data in parallel and a driver can commit once.

  Out of scope: streaming write and merge/update/delete DML (already reachable via SQL).

  ## Scope

  This can be implemented incrementally:

  - **PR 1** — Expose local append write + commit:
    `new_write_builder()`, `new_write()`, `write_arrow(batch)`,
    `prepare_commit()`, `new_commit().commit(messages)`.

    This gives a minimal end-to-end verifiable loop:
    write PyArrow data → prepare commit messages → commit snapshot → read data back.

  - **PR 2** — Make commit messages serializable:
    opaque / pickle-friendly `PyCommitMessage`, so commit messages can cross
    process boundaries for distributed writes such as Ray.

  - **PR 3** — Add dynamic overwrite support:
    `with_overwrite()` and `TableCommit.overwrite(messages)` for dynamic partition
    overwrite, where the partitions touched by the new commit messages are
    replaced.

  - **PR 4** — Add static partition overwrite support:
    extend `TableCommit.overwrite(...)` with a `static_partitions` argument, e.g.
    `overwrite(messages, {"dt": "2026-06-26"})`.

  - **PR 5** (in `apache/paimon`, `[python]`) — Switch PyPaimon's batch write path
    to delegate to the Rust pipeline internally.

  PR 1–4 land here; PR 5 lands in the main repo once the bindings are released.

  ### Notes

  `with_overwrite()` is intentionally split out from the first write PR because
[...]

PR #420, opened 2026-06-29
[python] Expose DataFrame-style batch write + commit loop

JunRuiLee:
### Purpose

First PR of exposing PyPaimon's DataFrame write path to Rust. Refs #414.

Exposes the Rust core's batch write loop through `bindings/python`:

```python
wb = table.new_write_builder()
write = wb.new_write()
write.write_arrow(batch)            # PyArrow RecordBatch, may be called repeatedly
messages = write.prepare_commit()
wb.new_commit().commit(messages)    # persists a snapshot; readable via SQL
```

`CommitMessage` (from `prepare_commit()`) is an opaque object passed back to `commit()` — same-process only for now; cross-process (pickle, for Ray worker→driver) is a later PR.

Out of scope (later PRs): serializable `CommitMessage`, overwrite, pypaimon-side wiring.

### Notes

- `write_arrow` validates the incoming batch schema against the table schema and raises `ValueError` on mismatch (names + types; binary/fixed_size_binary tolerated), mirroring pypaimon's `_validate_pyarrow_schema`. No implicit casting — callers supply correctly-typed batches (e.g. `INT` → Arrow `int32`).
- A single `commit_user` is fixed per `WriteBuilder` and shared by `new_write()` and `new_commit()`, so the writer and committer agree (Paimon uses it for duplicate-commit detection).
- `commit([])` (empty messages) is a no-op.

### Tests

`bindings/python/tests/test_write.py` — 6 tests: write→commit→SQL read-back roundtrip, multi-batch, empty-commit no-op, schema-mismatch ValueError, non-message TypeError.


bindings/python/src/write.rs
```diff
@@ -0,0 +1,184 @@
...
+    };
+    if input.fields().len() != target.fields().len() {
+        return Err(mismatch());
+    }
+    for (i, t) in input.fields().iter().zip(target.fields().iter()) {
+        if i.name() != t.name() {
+            return Err(mismatch());
+        }
+        if i.data_type() != t.data_type()
+            && !(is_binary_family(i.data_type()) && is_binary_family(t.data_type()))
```
JingsongLi: This accepts LargeBinary/FixedSizeBinary as equivalent to Binary, but the batch is then passed through unchanged. The lower write path still uses the original Arrow arrays/schema: for example `extract_datum_from_arrow` only downcasts binary Paimon fields to `arrow_array::BinaryArray`, and append writes would also see the incoming schema rather than the table schema. So a batch that this check accepts can later fail with a type-mismatch (or write files whose Arrow schema differs from the table schema). Please either reject non-exact binary-family types here or normalize/cast the batch to the target Arrow schema before calling `write_arrow_batch`.

JunRuiLee: Good catch — confirmed the lower write path downcasts binary fields to arrow_array::BinaryArray only, so tolerating LargeBinary/FixedSizeBinary here was a false positive. Made type matching strict (reject rather than cast, consistent with write_arrow not casting) and added a regression test. Fixed in 2fe6b39.

bindings/python/src/write.rs
```diff
@@ -0,0 +1,180 @@
...
+        let mut inner_messages = Vec::new();
+        let iter = messages.try_iter().map_err(|_| {
+            PyTypeError::new_err("commit() expects a sequence of CommitMessage objects")
+        })?;
+        for item in iter {
+            let item = item?;
+            let msg: PyRef<PyCommitMessage> = item.extract().map_err(|_| {
+                PyTypeError::new_err("commit() expects a sequence of CommitMessage objects")
+            })?;
+            inner_messages.push(msg.inner.clone());
```
JingsongLi, line 206: `CommitMessage` should carry the table identity (or at least table location) and `commit()` should reject messages prepared by a different table/builder before cloning the inner message. Right now any `CommitMessage` object is accepted here, so a Python caller can prepare messages from `wdb.t1` and pass them to `wdb.t2`'s committer. I reproduced that flow: the cross-table commit succeeds, but reading `t2` then fails because the snapshot references data files under `t2` that were only written under `t1`. Since these messages are opaque Python objects, this is worth guarding at the wrapper boundary.

JunRuiLee: Confirmed and reproduced — commit() accepted any CommitMessage with no table check, so t1's messages committed via t2 persist a snapshot referencing t1's files and t2 then fails to read.

Guarded at the wrapper boundary: PyCommitMessage now carries its source table location (stamped in prepare_commit), and commit() rejects any message whose location differs from the committer's table before cloning the inner message. Added a regression test for the cross-table flow. Pushed in ea4cbf0.

Note: once CommitMessage pickling lands (a follow-up PR for Ray distributed commit), the serialized payload will include table_location so this guard survives cross-process round-trips.

bindings/python/src/write.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Close writers and return the commit messages (opaque; pass to commit()).
+    fn prepare_commit(&mut self, py: Python<'_>) -> PyResult<Vec<PyCommitMessage>> {
+        let rt = runtime();
+        let messages = py
+            .detach(|| rt.block_on(async { self.inner.prepare_commit().await }))
+            .map_err(to_py_err)?;
+        Ok(messages
+            .into_iter()
+            .map(|inner| PyCommitMessage {
```
JingsongLi, line 155: This still lets commit messages from a different `WriteBuilder` for the same table pass the guard. Each `PyWriteBuilder` generates its own `commit_user`, and the Rust builder contract is that writers and committers created from that builder share it; it is used for snapshot duplicate detection and, for postpone-bucket tables, embedded in data file names. A natural Python flow like `table.new_write_builder().new_write()` followed by `table.new_write_builder().new_commit().commit(messages)` now succeeds but commits the files with a different committer `commit_user` than the writer used. Please stamp the originating `commit_user` onto `PyCommitMessage` as well and reject messages whose `commit_user` does not match the committer.

bindings/python/tests/test_write.py
```diff
@@ -0,0 +1,165 @@
...
+        batch = pa.record_batch(
+            [pa.array([1], pa.int32()), pa.array(["a"], pa.string())],
+            names=["id", "name"],
+        )
+        w1 = t1.new_write_builder().new_write()
+        w1.write_arrow(batch)
+        messages = w1.prepare_commit()
+        with pytest.raises(ValueError):
+            t2.new_write_builder().new_commit().commit(messages)
+
```
JingsongLi, line 165: `git diff --check origin/main...HEAD` reports this as a new blank line at EOF. Please remove the extra trailing blank line so the diff-check stays clean.

JunRuiLee: Removed, thanks.

bindings/python/src/write.rs
JunRuiLee: You're right — the location check alone still let a different WriteBuilder (same table) through, and since each builder mints its own commit_user, the writer and committer would disagree on it. Now PyCommitMessage also carries the originating builder's commit_user, and commit() rejects a mismatch in addition to the table-location check. Added a same-table-different-builder regression test (the legit table.new_write_builder() write+commit flow still passes since it shares one builder). Fixed in [6c1337a](https://github.com/apache/paimon-rust/pull/420/commits/6c1337ac51e245839bfdab234bd1809d480b14b0).

JingsongLi (approved):
+1

PR #421, opened 2026-06-29
perf: avoid expanding Mosaic row selections

QuakeWang:
### Purpose

Mosaic row selection previously expanded each selected row range into a `UInt64Array` of per-row indices before applying `take`. For large contiguous ranges, this makes memory and CPU cost proportional to the selected row count even though the input is already range-based.

### Brief change log

  - Keep Mosaic row selections as row-group-local `(offset, len)` slices.
  - Apply a single selected range with `RecordBatch::slice`.
  - Apply multiple selected ranges with `concat_batches`.
  - Preserve row counts for empty/all-missing projections.
  - Add coverage to ensure large selections stay range-based.

### Tests

  - `cargo test -p paimon --all-targets --features fulltext,vortex,mosaic test_mosaic`
  - `cargo test -p paimon --all-targets --features fulltext,vortex,mosaic arrow::format::mosaic::tests`


JingsongLi (approved):
+1

PR #422, opened 2026-06-29
Align TableCommit with FileStoreCommit commit semantics

JingsongLi:
## Summary

Align Rust `TableCommit` more closely with Python/Java `FileStoreCommit` commit behavior where the underlying Rust writer/commit contracts match. This adds manifest rolling and minor compaction, overwrite retry caching, richer latest-snapshot conflict checks, row-tracking parity for dedicated files, and column-aware global index update handling.

## Changes

- Add manifest options and write-path support for rolling manifest files by target size, plus minor compaction of small existing manifest files.
- Reuse overwrite base entries across retries when intervening snapshots are safe, while rebuilding on target-partition changes, non-APPEND commits, or whole-table overwrite.
- Extend commit conflict handling with delete existence checks and row-id existence/range checks against the current resolved base entries.
- Fill manifest `min_row_id`/`max_row_id` metadata when all entries carry row ids.
- Align row-tracking assignment with Python/Java behavior for missing `file_source`, blob files, and vector dedicated files.
- Make global index updates column-aware with default rejection and `DROP_PARTITION_INDEX` support, including `extra_field_ids`.
- Keep overwrite changelog handling Java-aligned: overwrite commits ignore changelog-only input.

## Testing

- [x] `cargo fmt --all`
- [x] `cargo test -p paimon table::table_commit --no-fail-fast -- --nocapture`
- [x] `cargo test -p paimon --lib --no-fail-fast`
- [x] `git diff --check`

## Notes

- `table_commit` coverage was expanded with Python-inspired regression tests for overwrite cache reuse/rebuild, manifest rolling/minor compaction, global index extra-field updates, and default-partition null overwrite behavior.
- Python's `check_from_snapshot` path is intentionally not mirrored here because Rust currently does not carry the same read-snapshot contract into `CommitMessage`; keeping a public field without a reliable producer would imply stronger conflict detection than Rust actually has.

### DIFF
--- a/crates/paimon/src/table/table_commit.rs
+++ b/crates/paimon/src/table/table_commit.rs
@@ -41,6 +41,8 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 /// Batch commit identifier (i64::MAX), same as Python's BATCH_COMMIT_IDENTIFIER.
 const BATCH_COMMIT_IDENTIFIER: i64 = i64::MAX;
+/// Java RollingFileWriter.CHECK_ROLLING_RECORD_CNT.
+const CHECK_ROLLING_RECORD_COUNT: usize = 1000;
 
 type PartitionBucketKey = (Vec<u8>, i32);
 type RowIdRange = (i64, i64);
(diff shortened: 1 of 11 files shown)


crates/paimon/src/table/table_commit.rs
```diff
@@ -655,16 +1044,29 @@ impl TableCommit {
                 } else {
                     CommitKind::APPEND
                 };
+                let detect_conflicts = has_delete;
```
QuakeWang: This only enables conflict detection for direct commits when the commit contains deletes. Data-evolution partial-column updates usually add files with preassigned `first_row_id`/`write_cols` but no deletes, so they skip the historical row-id/column conflict check and only validate that the original row-id range still exists.

That misses concurrent partial updates touching the same row-id and column range, regressing the Python `check_from_snapshot` path. Please keep an equivalent source-snapshot marker in the commit plan/message and run row-id/column conflict detection for these data-evolution commits, with a concurrent same-row-id/same-column update test.

JingsongLi: Addressed in e7d879c: restored a CommitMessage/Direct plan source snapshot marker, enabled row-id + write-column conflict detection for data-evolution partial updates, and added concurrent same-row-id/same-column reject plus different-column allow tests. DataEvolutionWriter now propagates the scan snapshot id into commit messages.

leaves12138 (approved):
LGTM. Reviewed the latest head, including the manifest rolling and compression follow-up commits. I also ran `cargo fmt --all -- --check` and `cargo test -p paimon table::table_commit --no-fail-fast` locally on `6f41f30`.

QuakeWang (approved):
+1

PR #423, opened 2026-06-30
[table] Support reading dedicated and rolling (multi-segment) vector files

JunRuiLee:
## Purpose

Add read-path support for dedicated `.vector.<format>` files in data-evolution splits, including **rolling (multi-segment)** vector columns. Upstream Paimon's `DedicatedFormatRollingFileWriter` rolls a vector column into multiple `.vector.` segments (using an independent `VECTOR_TARGET_FILE_SIZE`), so the Rust reader needs to consume such tables by reassembling the segments. This mirrors upstream's `VectorFileBunch` non-pushdown semantics in `DataEvolutionSplitRead.java`.

## Changes

All changes are confined to the read path (`crates/paimon/src/table/data_evolution_reader.rs`).

- **Classify & route vector files**: `is_vector_store_file_name` detector; `normalize_merge_group` classifies `.vector.` files into their own group, excludes them from the merge anchor, and forces the column-merge path.
- **`VectorBunch`** (modeled on the existing `BlobBunch`): aggregates rolled segments belonging to one logical vector source, keyed by `(schema_id, format_suffix, normalized_write_cols)`. `add` enforces continuity/dedup (same-`first_row_id` lower-seq -> ignore, overlap lower-seq -> ignore, gap -> error, row-count overflow -> error, key-identity mismatch -> error).
- **Segment ordering**: `normalize_merge_group` sorts vector segments by `first_row_id` asc, then `max_sequence_number` desc, and no longer validates rolled segments (which are slices) against the normal-file row range.
- **`normalized_write_cols`**: computed per vector file in `load_file_infos` (write columns sorted by field position; missing `write_cols` -> `DataInvalid`), so differently-ordered raw write columns that normalize equal aggregate into one bunch.
- **`build_source_plan`**: aggregates same-key segments into a single `FieldSource::VectorBunch`, with a bunch-granularity duplicate-field-id guard and a bunch row-count check symmetric to the blob check. A single vector file is simply a bunch of length 1 -- one code path, no special-casing.
- **Read**: `VectorBunch` is read via the same sequential-concat path as `BlobBunch`; each segment carries its real `first_row_id`/`row_count`, so the existing `to_local_row_ranges` mechanism clips absolute row ranges per segment correctly.

## Out of scope

Write-side changes; blob fallback-placeholder optimization; predicate/vector pushdown into vector files; upstream's `rowIdPushDown` branch (non-pushdown semantics only).

## Tests

- Unit: `VectorBunch::add` guards (gap / overlap-lower-seq / same-first_row_id dedup / row-count overflow / key-identity mismatch); `normalize_merge_group` multi-segment sort and removal of the vector range check; `normalize_vector_write_cols` sort-by-position and missing/unknown-column errors; `build_source_plan` aggregation (incl. differently-ordered write cols collapsing to one bunch) and cross-bunch duplicate-field-id rejection.
- Integration (end-to-end): normal `data.parquet` + 3 rolled `.vector.parquet` segments reassemble into the correct column in order (including a null row); `row_ranges` selecting rows **across a segment boundary** return the correct subset.

All 41 `data_evolution_reader` tests pass; clippy clean.

> Note: this builds on the earlier dedicated-vector classification/routing commits; the diff against `main` includes those as they have not yet landed upstream.

### DIFF
--- a/crates/paimon/src/table/data_evolution_reader.rs
+++ b/crates/paimon/src/table/data_evolution_reader.rs
@@ -33,8 +33,23 @@ use futures::StreamExt;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
+/// Whether a file name denotes a dedicated vector-store file (`*.vector.<format>`).
+/// Mirrors upstream `VectorType.isVectorStoreFile`: the name contains `.vector.`.
+fn is_vector_store_file_name(file_name: &str) -> bool {
+    file_name.to_ascii_lowercase().contains(".vector.")
+}
+
 /// Whether the files in a split can be read independently (no column-wise merge needed).
 fn is_raw_convertible(files: &[DataFileMeta]) -> bool {
+    // A split containing a dedicated vector file must go through the column-merge
+    // path so vector columns are routed to their VectorFile source. Check this
+    // BEFORE the single-file early-return.
+    if files
+        .iter()
+        .any(|f| is_vector_store_file_name(&f.file_name))
+    {
+        return false;
+    }
     if files.len() <= 1 {
         return true;
     }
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #424, opened 2026-06-30
fix: avoid whole-file reads for Mosaic

QuakeWang:
### Purpose

The Rust Mosaic reader currently reads the entire `.mosaic` file before constructing the Mosaic core reader. This prevents projection and row-group pruning from reducing storage I/O and also forces file-sized memory usage for large Mosaic files.

### Brief change log

  - Replace the eager whole-file `MemoryInputFile` path with a `FileRead`-backed Mosaic `InputFile`.
  - Preserve Mosaic core's random-access range reads for footer metadata, row-group buckets, and projected column slots.
  - Add a large-file regression test verifying projected Mosaic reads do not issue a `0..file_size` storage read.

### Tests

  - `cargo fmt --check`
  - `cargo test -p paimon --features mosaic arrow::format::mosaic::tests -- --nocapture`
  - `cargo test -p paimon --features mosaic table::data_file_reader::tests::test_mosaic -- --nocapture`

### DIFF
--- a/crates/paimon/src/arrow/format/mosaic.rs
+++ b/crates/paimon/src/arrow/format/mosaic.rs
@@ -35,6 +35,8 @@ use paimon_mosaic_core::stats::ColumnStats;
 use paimon_mosaic_core::values::Value as MosaicValue;
 use std::collections::{HashMap, HashSet};
 use std::io;
+use std::ops::Range;
+use std::sync::Arc;
 
 pub(crate) struct MosaicFormatReader;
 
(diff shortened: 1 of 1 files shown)


crates/paimon/src/arrow/format/mosaic.rs
```diff
@@ -313,39 +318,86 @@ fn micros_to_millis_nanos(micros: i64) -> (i64, i32) {
...
+                tokio::runtime::Handle::try_current().is_ok_and(|handle| {
+                    handle.runtime_flavor() == tokio::runtime::RuntimeFlavor::MultiThread
+                });
+            if in_multi_thread_runtime {
+                tokio::task::block_in_place(do_read)
+            } else {
+                do_read()
+            }
+        }
+        _ => {
```
JingsongLi: This current-thread runtime fallback can still hang for real `FileRead` implementations. `read_batch_stream` is already blocked inside `MosaicReader::new` / `read_at`, so the current-thread executor is not being driven; the spawned thread then calls `Handle::block_on` on that same current-thread runtime, which does not drive IO/timer drivers. The added tests use an immediate in-memory `FileRead`, so they do not exercise opendal/network/file futures that need the runtime. Please move the synchronous Mosaic work onto a blocking thread and dispatch async reads back to the runtime with a blocking channel/oneshot, or otherwise fail/require a multi-thread runtime instead of trying to support current-thread this way.

JingsongLi (approved):
+1

PR #425, opened 2026-06-30
feat(predicate): add string and range leaf operators (StartsWith/EndsWith/Contains/Like/Between)

JunRuiLee:
### Purpose

Java Paimon's `LeafFunction` set includes `StartsWith` / `EndsWith` / `Contains` / `Like` and `Between`, but `paimon-rust` only has the basic comparison/null/set operators. So a DataFusion filter like `col LIKE 'x%'` or `col BETWEEN a AND b` can't be translated and degrades to a full scan + residual instead of pushing down. This PR adds those operators and wires them end-to-end (builder, eval, stats prune, Parquet row filter, b-tree index, DataFusion translation).

### Brief change log

3 commits (recommend per-commit review; later build on earlier):

1. **StartsWith / EndsWith / Contains** — empty pattern folds to `IsNotNull`; Vortex/ORC/b-tree fall open. Adds `arrow-string`.
2. **Like + LikeOptimization** — rewrites `prefix%`/`%suffix`/`%mid%`/no-wildcard into the string ops or `Eq`, else a residual `Like` leaf matched with arrow's like-kernel escape semantics. Only `\` escape; `ILIKE`/`NOT LIKE` fall open.
3. **Between / NotBetween** — `GtEq+LtEq` overlap semantics; `NotBetween` pushes down safely (NULL never matches, data ranges stay `Inexact`). Also fixes the b-tree global index keeping a file when a `Between` conjunct misses but a co-conjunct matches. Adds `arrow-arith`.

### Tests

New unit/translation tests across `predicate.rs`, `predicate_stats.rs`, `parquet.rs`, `filter_pushdown.rs` (incl. `NOT BETWEEN`) and a `global_index_scanner` regression test. `cargo test -p paimon --lib --features vortex` and `-p paimon-datafusion --lib filter_pushdown` pass.

### API and Format

Adds `PredicateOperator` variants + builder methods (`starts_with`/`ends_with`/`contains`/`like`/`between`/`not_between`). No format change. New deps: `arrow-string`, `arrow-arith` (58.0).

### Documentation

Covered by rustdoc on the new builder methods.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,12 +30,14 @@ rust-version = "1.91.0"
 [workspace.dependencies]
 arrow = "58.0"
 arrow-array = { version = "58.0", features = ["ffi"] }
+arrow-arith = "58.0"
 arrow-buffer = "58.0"
 arrow-schema = "58.0"
 arrow-cast = "58.0"
 arrow-ord = "58.0"
 arrow-row = "58.0"
 arrow-select = "58.0"
+arrow-string = "58.0"
 datafusion = "53.0.0"
 datafusion-ffi = "53.0.0"
 paimon = { version = "0.3.0", path = "crates/paimon" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ rust-version = "1.91.0"
 [workspace.dependencies]
 arrow = "58.0"
 arrow-array = { version = "58.0", features = ["ffi"] }
-arrow-arith = "58.0"
 arrow-buffer = "58.0"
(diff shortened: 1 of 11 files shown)


QuakeWang (approved):
LGTM. I did not find blocking issues in the predicate translation/evaluation, stats pruning, Parquet row filtering, or global-index paths.

nit: Consider adding an end-to-end SQL test for `LIKE` on a partition column, since partition-only predicates may be classified as `Exact`. The current unit coverage looks sound, so this is not blocking.

PR #426, opened 2026-07-01
fix(scan): prune OR predicates using stats

QuakeWang:
### Purpose

Data stats pruning currently treats `Predicate::Or` as unsupported and fails open. This means predicates like `id < 5 OR id > 25` cannot prune files whose stats range is `[10, 20]`, even though no branch can match.

### Brief change log

  - Evaluate `Predicate::Or` in the shared stats may-match evaluator with `any(child may match)`.
  - Apply the same OR semantics to data-evolution group stats pruning.
  - Keep `Predicate::Not` fail-open.
  - Add regression tests for OR prune/keep behavior and NOT fail-open behavior.

### Tests


JingsongLi (approved):
+1

PR #428, opened 2026-07-01
[python] Add TableRead.read(splits) returning Arrow batches

JunRuiLee:
### Purpose

Third PR of exposing PyPaimon's DataFrame read path to Rust (follows #415 scan/plan/splits and #419 `with_filter`). Refs #413.

Adds the data-reading step both prior PRs deferred: `ReadBuilder.new_read().read(splits) → List[pyarrow.RecordBatch]`.

```python
b = table.new_read_builder().with_projection(["id"])
splits = b.new_scan().plan().splits()
batches = b.new_read().read(splits)   # -> List[pyarrow.RecordBatch]
```

Splits are picklable (from #415), so they can be planned in one process and read in another — groundwork for Ray distributed reads.

### Notes

- **Shared config**: extracts `apply_read_config` (projection/limit/filter) so `new_scan().plan()` and `new_read().read()` apply the same pushdown consistently. Pure refactor, no behavior change to scan.
- **Config snapshot**: `new_read()` snapshots the builder's projection/limit/filter, so mutating the builder afterward does not affect an already-created read.
- **Projection now applies**: projection was a planning no-op in #415; it takes effect here on the read path.
- **Eager collection**: `read()` collects the full Arrow stream into a `List[RecordBatch]`.
- **Errors**: a non-iterable argument or a non-`Split` element raises `TypeError`; empty splits return `[]` without entering the runtime. Off-GIL read via `py.detach` + runtime `block_on`, matching the existing `plan()` pattern.

Out of scope (later PRs): pypaimon-side wiring.

### Tests

`bindings/python/tests/test_read.py` — 7 new tests: expected rows, empty splits, non-`Split` `TypeError`, pickled-split round-trip, projection applied, config-snapshot isolation, and a `with_filter` read smoke test. Full file 33/33 passing; `cargo clippy` clean.


bindings/python/src/read.rs
```diff
@@ -114,6 +155,39 @@ impl PyTableScan {
...
+    limit: Option<usize>,
+    filter: Option<Predicate>,
+}
+
+#[pymethods]
+impl PyTableRead {
+    /// Read the given splits into a list of PyArrow RecordBatches.
+    fn read(&self, py: Python<'_>, splits: &Bound<'_, PyAny>) -> PyResult<Vec<Py<PyAny>>> {
+        let splits = extract_splits(splits)?;
+        if splits.is_empty() {
```
QuakeWang: this empty-splits fast path skips `ReadBuilder::new_read()` validation, so invalid projections can return `[]` only when no splits are passed. Consider constructing `read` before this return to keep validation consistent.

JunRuiLee: Good catch — fixed in 79836ad.

QuakeWang (approved):
+1

PR #429, opened 2026-07-01
feat: support orc is null predicate pushdown

liujiwen-up:
### Purpose

Linked issue: none.

ORC predicate pushdown already handles supported comparison, `IN`, and `IS NOT NULL` predicates. This patch adds the missing `IS NULL` translation so ORC row-group statistics can prune groups that cannot contain nulls.

### Brief change log

- Translate supported Paimon `IS NULL` predicates to `orc_rust::predicate::Predicate::is_null`.
- Keep unsupported ORC predicate types fail-open, matching existing `IS NOT NULL` behavior.
- Add unit coverage for supported `IS NULL` pushdown and unsupported type fallback.

### Tests

- `cargo fmt --check`: passed
- `cargo test -p paimon arrow::format::orc`: passed, 21 tests

### API and Format

### Documentation

Not required; this is an internal predicate pushdown optimization with tests.

### DIFF
--- a/crates/paimon/src/arrow/format/orc.rs
+++ b/crates/paimon/src/arrow/format/orc.rs
@@ -208,10 +208,14 @@ fn build_orc_leaf_predicate(
     let column = file_field.name();
 
     match op {
-        PredicateOperator::IsNotNull
+        PredicateOperator::IsNull | PredicateOperator::IsNotNull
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
+1

PR #430, opened 2026-07-01
feat(datafusion): add push down temporal literals

liujiwen-up:
### Purpose

DataFusion filter pushdown currently cannot translate temporal literal `ScalarValue`s into Paimon `Datum`s, so predicates on Paimon `TIME`, `TIMESTAMP`, and `TIMESTAMP WITH LOCAL TIME ZONE` columns fall open unnecessarily.

### Changes

- Translate `ScalarValue::Time32Millisecond` into `Datum::Time` for Paimon `Time` fields.
- Translate timezone-free `TimestampSecond` / `TimestampMillisecond` / `TimestampMicrosecond` / `TimestampNanosecond` into `Datum::Timestamp`.
- Translate timezone-bearing timestamp scalar values into `Datum::LocalZonedTimestamp`.
- Use `div_euclid` / `rem_euclid` for sub-millisecond timestamp parts so negative timestamps before 1970 split correctly.
- Keep timezone semantic mismatches unsupported so DataFusion residual filters continue to preserve correctness.

### Tests

- `cargo test -p paimon-datafusion filter_pushdown`
- `cargo test -p paimon-datafusion --test read_tables test_temporal_filter_pushdown_via_datafusion_scan -- --nocapture`
- `cargo test -p paimon-datafusion` was also run locally; all new `filter_pushdown` tests passed, but the full package run hit existing fixture-dependent failures for missing `default.partitioned_log_table` / `default.multi_partitioned_log_table` in `table::tests::*scan*` tests.

### DIFF
--- a/crates/integrations/datafusion/tests/read_tables.rs
+++ b/crates/integrations/datafusion/tests/read_tables.rs
@@ -433,7 +433,7 @@ async fn test_temporal_filter_pushdown_via_datafusion_scan() {
         .expect("INSERT should collect");
 
     let timestamp_sql = "SELECT id, name FROM paimon.test_db.temporal_filter_pushdown \
-        WHERE ts = TIMESTAMP '2024-01-01 00:00:00.123456' AND id + 1 > 1";
+        WHERE ts = TIMESTAMP '2024-01-01 00:00:00.123456' AND id + 1 > 2";
     let plan = sql_context
         .sql(timestamp_sql)
         .await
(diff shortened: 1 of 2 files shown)


crates/integrations/datafusion/tests/read_tables.rs
```diff
@@ -400,6 +402,93 @@ async fn test_partially_translated_filter_keeps_partition_pruning_and_correctnes
...
+                (2, 'bob', TIMESTAMP '2024-01-01 00:00:00.654321', TIMESTAMP '2024-01-01 00:00:00.654321+00:00'),
+                (3, 'carol', TIMESTAMP '2024-01-02 00:00:00.123456', TIMESTAMP '2024-01-02 00:00:00.123456+00:00')",
+        )
+        .await
+        .expect("INSERT should succeed")
+        .collect()
+        .await
+        .expect("INSERT should collect");
+
+    let timestamp_sql = "SELECT id, name FROM paimon.test_db.temporal_filter_pushdown \
```
QuakeWang, line 435: This residual predicate is true for the row matched by the temporal predicate, so the result would stay the same even if the residual filter were dropped. Could we make the residual condition actually filter out a temporal match, or assert the plan still contains a residual FilterExec?

JingsongLi (approved):
+1

QuakeWang (approved):
LGTM

JingsongLi:
I think temporal literal pushdown has a correctness risk for negative sub-millisecond timestamps.

`filter_pushdown.rs::scalar_to_timestamp_parts` converts microsecond/nanosecond timestamp literals with `div_euclid` / `rem_euclid`. But the current Arrow-to-BinaryRow write path uses truncating division and remainder, e.g. in `spec/binary_row.rs`:

- `millis = micros / 1000`
- `nanos = (micros % 1000) * 1000`

For negative timestamps this produces different `(millis, nanos)` pairs. For example, `-1us` becomes `(-1ms, 999000ns)` in the pushdown literal path, but the write path represents it as `(0ms, -1000ns)`. This can make pushed timestamp predicates false-negative for pre-epoch timestamps with fractional milliseconds.

Could we either:
1. align the BinaryRow timestamp write/extract path with the euclidean representation, or
2. make pushdown use the same representation as the existing write path until the core timestamp representation is changed?

It would also be good to add an integration test that writes a negative microsecond/nanosecond timestamp and verifies a pushed equality/range predicate through DataFusion scan.

PR #440, opened 2026-07-02
[table] Accept VECTOR<FLOAT> in Lumina index build

JunRuiLee:
## Summary

Part of #410 (PR 4). Integrates the native `VECTOR<FLOAT, N>` type with the Lumina
index build path. Previously only `ARRAY<FLOAT>` columns could be indexed; this
extends validation, dimension resolution, and Arrow vector extraction to also
accept `DataType::Vector`, deriving the dimension from the column type
(`VectorType::length`) instead of requiring an explicit `lumina.index.dimension`
option. This mirrors Java `LuminaVectorGlobalIndexWriter.validateAndResolveDimension`,
where both `ARRAY<FLOAT>` and `VECTOR<FLOAT>` are valid vector-index columns.

## Changes

- **Validation** (`validate_vector_field`): accept `VECTOR<FLOAT>` in addition to
  `ARRAY<FLOAT>`; reject non-float element types (e.g. `VECTOR<DOUBLE>`).
- **Dimension resolution** (`effective_lumina_options`, new): for a
  `VECTOR<FLOAT, N>` column, derive `lumina.index.dimension = N` from the type.
  Absent → inject `N`; present-and-equal → keep; present-and-different →
  `ConfigInvalid`. Non-vector columns (`ARRAY<FLOAT>`) are returned unchanged so
  existing option/default behavior is preserved. Matches Java's
  `testVectorTypeRejectsExplicitDimensionConflict`.
- **Extraction** (`extract_vectors_from_batches`): resolve the vector column as
  either `List<Float32>` (ARRAY) or `FixedSizeList<Float32>` (VECTOR) via a
  `VectorLayout` enum; both yield a `Float32Array` plus a per-row `[start, end)`
  slice, with identical null-vector / null-element / dimension-mismatch handling.
- Add `LUMINA_DIMENSION_OPTION` constant in `lumina/mod.rs`.

## Scope note

#410's PR 4 description names `lumina`/`vindex` index build **and**
`VectorSearchBuilder`. Only the Lumina build path is touched here, deliberately:

- **vindex index build** does not exist in Rust yet — `src/vindex/` currently has
  only a reader (from #399), no writer/builder. VECTOR support for vindex will be
  built in when the vindex build path is ported (the Java vindex writer already
  accepts both `ARRAY<FLOAT>` and `VECTOR<FLOAT>`, so it will be internal to that
  work rather than a follow-up integration).
- **`VectorSearchBuilder`** dispatches by field name and index-file type against
  already-built index files; it does not re-validate the source column as ARRAY
  vs VECTOR, so it is transparent to the column type and needs no change.

## Testing

- [x] `cargo fmt`
- [x] `cargo check -p paimon`
- [x] `cargo test -p paimon --lib lumina_index_build_builder` (28 passed, 1 ignored — the ignored test requires `LUMINA_LIB_PATH`)

New tests cover: VECTOR<FLOAT> accepted / VECTOR<DOUBLE> rejected; dimension
absent-inject / matching-ok / mismatch-errors / array-unchanged; dimension
propagation into native Lumina options; and FixedSizeList<Float32> extraction
incl. null-vector, null-element, and dimension-mismatch rejection.

### DIFF
--- a/crates/paimon/src/lumina/mod.rs
+++ b/crates/paimon/src/lumina/mod.rs
@@ -227,6 +227,9 @@ pub const KEY_DIMENSION: &str = "index.dimension";
 pub const KEY_DISTANCE_METRIC: &str = "distance.metric";
 pub const KEY_INDEX_TYPE: &str = "index.type";
 
+/// Paimon-prefixed option key for the vector index dimension.
+pub const LUMINA_DIMENSION_OPTION: &str = "lumina.index.dimension";
+
 pub struct LuminaIndexMeta {
     options: HashMap<String, String>,
 }
--- a/crates/paimon/src/table/lumina_index_build_builder.rs
+++ b/crates/paimon/src/table/lumina_index_build_builder.rs
@@ -289,13 +289,18 @@ fn find_index_field<'a>(table: &'a Table, column: &str) -> Result<&'a DataField>
 }
 
 fn validate_vector_field(field: &DataField) -> Result<()> {
-    if !matches!(
+    let is_array_float = matches!(
         field.data_type(),
         DataType::Array(array) if matches!(array.element_type(), DataType::Float(_))
-    ) {
+    );
+    let is_vector_float = matches!(
+        field.data_type(),
+        DataType::Vector(vector) if matches!(vector.element_type(), DataType::Float(_))
+    );
+    if !is_array_float && !is_vector_float {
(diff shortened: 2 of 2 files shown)


QuakeWang (approved):
+1

JingsongLi (approved):
+1

PR #441, opened 2026-07-02
[python] Wire per-read scan options (time travel) into the Rust read kernel

JunRuiLee:
### Purpose

pypaimon currently cannot pass per-read scan options into the Rust core:
`ReadBuilder` only carries projection/limit/filter, so a time-travel selector
(e.g. `scan.snapshot-id`) passed through the Python binding is silently read at
latest. This wires a scan-options channel through the Python binding and hardens
time-travel selector semantics in the core.

### Brief change log

- **core**: `scan.snapshot-id` and `scan.tag-name` become distinct time-travel
  selectors with strict semantics — `scan.snapshot-id` parses a snapshot id only
  (never falls back to a tag lookup), `scan.tag-name` resolves a tag only (never
  falls back to a snapshot id), while `scan.version` keeps the ambiguous
  tag→id compatibility of SQL `VERSION AS OF`. All four selectors are mutually
  exclusive and invalidate the cached travel snapshot on re-selection in
  `copy_with_options`.
- **core**: `CoreOptions::validate_scan_options()` rejects scan options the Rust
  reader does not implement yet (`incremental-between*`, `scan.watermark`,
  `scan.mode != default`) with `Error::Unsupported`, invoked on the merged
  options inside `copy_with_time_travel` before travel resolution.
- **python**: `new_read_builder(options)` resolves time travel before building
  the `ReadBuilder` (so filters validate against the travelled schema); a
  set-but-unresolved selector raises instead of silently reading latest;
  conflicting selectors raise naming both keys; `Error::Unsupported` maps to
  `NotImplementedError`; non-string option values raise `TypeError`. Absent/empty
  options remain a zero-cost latest read.

Builds on the selector-conflict/strict-parsing work from #218 (already merged),
extending it to first-class `scan.snapshot-id`/`scan.tag-name` selectors and the
pypaimon read path.

### Tests

- core: `cargo test -p paimon --lib` (1049 passed) — selector mapping, four-way
  conflict rejection listing original keys, strict `scan.snapshot-id` /
  `scan.tag-name` rejection (tag-vs-id crossovers fail), `validate_scan_options`
  gate, and stale-cache invalidation on selector change.
- python: `pytest tests/test_read.py` (42 passed) — snapshot-id / tag-name
  travel, unresolved and conflicting selectors, unsupported options, `TypeError`,
  the no-options latest-read regression, and travelled-schema filter validation.

### API and Format

Adds an optional `options: Optional[Dict[str, str]]` parameter to
`Table.new_read_builder` in the Python binding. No storage format change.

### Documentation

Behavior is documented inline on the new option handling; no separate docs
update required.


crates/paimon/src/spec/core_options.rs
```diff
@@ -65,6 +65,13 @@ const DEFAULT_COMMIT_MIN_RETRY_WAIT_MS: u64 = 1_000;
 const DEFAULT_COMMIT_MAX_RETRY_WAIT_MS: u64 = 10_000;
 pub const SCAN_TIMESTAMP_MILLIS_OPTION: &str = "scan.timestamp-millis";
 pub const SCAN_VERSION_OPTION: &str = "scan.version";
+pub const SCAN_SNAPSHOT_ID_OPTION: &str = "scan.snapshot-id";
```
JingsongLi, line 69: Now that `scan.snapshot-id` and `scan.tag-name` are first-class time-travel selectors, the DataFusion SQL write guard also needs to treat them as time-travel options. `SQLContext::ensure_no_time_travel_for_write` still only checks `scan.version` and `scan.timestamp-millis`, while `SET 'paimon.scan.snapshot-id' = '1'` is accepted into `dynamic_options` generically. That lets write statements get past the guard even though reads in the same statement/session can resolve the target table through a historical snapshot. For example, the data-evolution UPDATE path builds `SELECT "_ROW_ID" ... FROM target` through DataFusion (honoring dynamic options) and then writes through the latest catalog table, so this can update based on stale row ids instead of failing like the existing selectors do. Please include these two new keys in the write guard (and matching tests).

JunRuiLee: Good catch, thanks — you're right that this is the same write-guard gap as the write_builder path but at the SQL layer. Fixed in bc08f7d: `ensure_no_time_travel_for_write` now also rejects `scan.snapshot-id` and `scan.tag-name`, so UPDATE/DELETE/MERGE/INSERT OVERWRITE/TRUNCATE fail up front instead of writing against latest while reads resolve a historical snapshot. Added `test_write_guard_rejects_snapshot_id_and_tag_name_selectors` covering both keys across those write statements.

crates/paimon/src/spec/core_options.rs
```diff
@@ -176,16 +183,64 @@ pub struct CoreOptions<'a> {
...
+            INCREMENTAL_BETWEEN_SCAN_MODE_OPTION,
+            SCAN_WATERMARK_OPTION,
+        ] {
+            if self.options.contains_key(key) {
+                return Err(crate::Error::Unsupported {
+                    message: format!("Scan option '{key}' is not supported by the Rust reader yet"),
+                });
+            }
+        }
+        if let Some(mode) = self.options.get(SCAN_MODE_OPTION) {
```
QuakeWang, line 250: This rejects documented Paimon option combinations such as `scan.mode=from-snapshot` with `scan.snapshot-id` / `scan.tag-name`, and `scan.mode=from-timestamp` with `scan.timestamp-millis`.

Those modes are standard ways to express the selectors this PR now supports. Could we allow the explicit modes when they match an implemented selector, and only reject truly unsupported modes like `incremental` / `compacted-full`?

crates/paimon/src/table/table_scan.rs
```diff
@@ -364,8 +364,10 @@ impl<'a> TableScan<'a> {
     /// Plan the full scan: resolve snapshot (via options or latest), then read manifests and build DataSplits.
     ///
     /// Time travel is resolved from table options:
-    /// - only one of `scan.version`, `scan.timestamp-millis` may be set
-    /// - `scan.version` → tag name (if exists) → snapshot id (if parseable) → error
+    /// - only one of `scan.version`, `scan.timestamp-millis`,
+    ///   `scan.snapshot-id`, `scan.tag-name` may be set
+    /// - `scan.version` / `scan.snapshot-id` / `scan.tag-name` → tag name (if
```
QuakeWang: The comment is now stale: `scan.snapshot-id` and `scan.tag-name` are strict selectors in the implementation. Only `scan.version` does tag-first then snapshot-id fallback.

JingsongLi:
Looks good to me!

crates/paimon/src/spec/core_options.rs
JunRuiLee: Fixed in 490f61b. `validate_scan_options` now accepts explicit `scan.mode=from-snapshot` when `scan.snapshot-id` / `scan.tag-name` / `scan.version` is present, and `scan.mode=from-timestamp` with `scan.timestamp-millis` — Java's `CoreOptions.setDefaultValues()` writes these modes next to their selectors, so configs from the Java toolchain carry both keys. Two details beyond the suggestion: an explicit mode *without* its selector fails as invalid input (mirroring Java `SchemaValidation`) rather than silently reading latest, and truly unimplemented modes (`incremental`, `compacted-full`, `latest`, ...) still raise `Unsupported` → `NotImplementedError`. Covered by new tests in `core_options.rs` and `test_read.py`.

crates/paimon/src/table/table_scan.rs
JunRuiLee: Fixed in 490f61b — the `plan()` doc now describes each selector separately: `scan.version` keeps the ambiguous tag → snapshot-id resolution, while `scan.snapshot-id` and `scan.tag-name` are documented as strict. Also scoped the `TimeTravelUtil` reference to `scan.version` only, since the strict selectors don't follow it.

QuakeWang (approved):
LGTM

PR #442, opened 2026-07-03
fix(scan): prune data-evolution files by projection

QuakeWang:
### Purpose

DataFusion projections were only applied at read time. For data-evolution tables, scan planning still kept every physical file in a row-id group even when the projected read type did not need the columns written by some files.

This made projection pruning ineffective for data-evolution planning and could keep unnecessary sidecar/partial-column files in planned splits.

### Brief change log

  - Propagate DataFusion projection columns into `ReadBuilder::new_scan()`.
  - Validate scan projection names consistently with reads.
  - Prune data-evolution row-id groups by projected stable field ids, including renamed-column cases.
  - Keep a normal representative file when needed for row count / merge reader constraints.
  - Keep the normal anchor when deletion vectors are enabled.
  - Cache `(schema_id, write_cols)` field-id resolution during scan planning.

### Tests

  - `cargo test -p paimon --all-targets --features fulltext,vortex,mosaic table::table_scan::tests::test_data_evolution_pruning`
  - `cargo test -p paimon --all-targets --features fulltext,vortex,mosaic table::read_builder::tests::test_new_scan_validates_unknown_projection`
  - `cargo test -p paimon-datafusion --all-targets test_scan_applies_projection_to_data_evolution_planning`

### DIFF
--- a/crates/integrations/datafusion/src/table/mod.rs
+++ b/crates/integrations/datafusion/src/table/mod.rs
@@ -177,6 +177,16 @@ impl TableProvider for PaimonTableProvider {
         // Plan splits eagerly so we know partition count upfront.
         let filter_analysis = analyze_filters(filters, self.table.schema().fields());
         let mut read_builder = self.table.new_read_builder();
+        let projected_columns = projection.map(|indices| {
+            indices
+                .iter()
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #443, opened 2026-07-03
[python] Support Date/Time/Timestamp/Decimal filter literals in py_to_datum

JunRuiLee:
### Purpose

`py_to_datum` in the Python binding raises `NotImplementedError` for any non-primitive literal, so a filter on a temporal or decimal column can never be pushed down through the DataFrame-style read path — every such query fails even though the core predicate machinery fully supports these `Datum` variants. This adds DataType-driven literal conversion for Date / Time / Timestamp / LocalZonedTimestamp / Decimal.

Complements #430, which adds the equivalent temporal conversions on the DataFusion (SQL) path; this PR covers the Python binding path.

Linked issue: none.

### Brief change log

- **Date**: `datetime.date` → epoch days. `datetime.datetime` (a `date` subclass) is rejected so the time-of-day part is never silently dropped.
- **Time**: naive `datetime.time` → millis-of-day. Timezone-aware times and sub-millisecond microseconds are rejected (`Datum::Time` carries millis; silent truncation would change comparison semantics).
- **Timestamp**: naive `datetime.datetime` → `{millis, nanos}` with a Euclidean split so pre-epoch instants keep a non-negative nano remainder — same convention as `BinaryRow::get_timestamp_raw` and the DataFusion pushdown conversion. Timezone-aware datetimes are rejected (parity with the SQL path, which refuses zoned literals for TIMESTAMP).
- **LocalZonedTimestamp**: timezone-aware `datetime.datetime` normalized through `astimezone(utc)`, so any tzinfo implementation (zoneinfo, pytz, fixed offsets) is handled by Python itself. Naive datetimes are rejected as ambiguous.
- **Decimal**: `decimal.Decimal` or `int`, rescaled losslessly to the column's scale via `as_tuple()` (avoids Decimal context rounding). Values needing rounding, exceeding the column precision, non-finite values (NaN/Infinity), `bool`, and binary `float`s are rejected.
- All rejections raise `ValueError` (type mismatch); `NotImplementedError` remains for genuinely unsupported types (Bytes/complex).
- Enables the `chrono` feature of pyo3 (chrono is already in the workspace dependency tree; no new crates).

### Tests

- Rust: `cargo test -p pypaimon_rust` — 51 passed. New unit tests cover epoch-day conversion (incl. pre-epoch), datetime-as-date rejection, time tz/sub-milli rejection, naive/aware timestamp acceptance+rejection for both TIMESTAMP and TIMESTAMP_LTZ, pre-epoch Euclidean split, and decimal exact-scale / lossless-rescale / scientific-notation / negative / int-literal / rounding / overflow / NaN / float / bool cases.
- Python: `uv run --no-sync pytest tests/test_read.py` — 39 passed. New end-to-end tests build a table with DATE/TIMESTAMP/DECIMAL columns across two files and assert the filter both prunes splits (stats-based) and returns the matching rows via `TableRead.read`.

### API and Format

No API surface change (the same `with_filter(dict)` accepts more literal types). No storage format change.

### Documentation

Behavior documented on `py_to_datum` and the new conversion helpers.

### DIFF
--- a/bindings/python/Cargo.toml
+++ b/bindings/python/Cargo.toml
@@ -28,16 +28,17 @@ doc = false
 
 [dependencies]
 arrow = { workspace = true, features = ["pyarrow"] }
+chrono = "0.4.38"
 datafusion = { workspace = true }
 datafusion-ffi = { workspace = true }
 futures = "0.3"
 paimon = { path = "../../crates/paimon", features = ["storage-all"] }
 paimon-datafusion = { path = "../../crates/integrations/datafusion", features = ["fulltext"] }
-pyo3 = { version = "0.28", features = ["abi3-py310"] }
+pyo3 = { version = "0.28", features = ["abi3-py310", "chrono"] }
 serde_json = "1.0"
 tokio = { workspace = true }
 
 [dev-dependencies]
 # Enables `Python::attach` to auto-initialize an interpreter for Rust unit tests.
 # Only applied to the test build; the maturin/extension-module build does not pull dev-dependencies.
-pyo3 = { version = "0.28", features = ["abi3-py310", "auto-initialize"] }
+pyo3 = { version = "0.28", features = ["abi3-py310", "auto-initialize", "chrono"] }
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #444, opened 2026-07-03
feat(datafusion): support SHOW CREATE TABLE via get_table_definition

shyjsarah:
## Summary

- `SHOW CREATE TABLE` on a Paimon table returned no DDL because DataFusion 53 rewrites the statement into a query against `information_schema.views`, whose `definition` column is populated by `TableProvider::get_table_definition()`. `PaimonTableProvider` did not override this method, so the column came back empty.
- Add a cached `table_definition` string on `PaimonTableProvider`, built once in `try_new` from the table's identifier and schema (fields, primary keys, partition keys, options).
- Add `data_type_to_sql` covering all 22 `DataType` variants (recursive for `Array`/`Map`/`Multiset`/`Row`/`Vector`).
- Override `TableProvider::get_table_definition()` to return the cached DDL.

## Test plan

- [x] `cargo check -p paimon-datafusion`
- [x] `cargo clippy -p paimon-datafusion --lib --tests` (zero warnings)
- [x] New tests in `sql_context_tests.rs` (4 cases): simple table, table with primary key, table with partition + options, table with various data types
- [x] Full `sql_context_tests` suite (39/39) — no regression
- [ ] Manual: `SHOW CREATE TABLE <db>.<table>` returns the DDL string


crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,119 @@ impl PaimonTableProvider {
...
+        "CREATE TABLE {}.{} (",
+        identifier.database(),
+        identifier.object()
+    );
+
+    for (i, field) in schema.fields().iter().enumerate() {
+        if i > 0 {
+            ddl.push_str(", ");
+        }
+        let _ = write!(
```
QuakeWang: This drops column nullability. For example, a table created with `payload BLOB NOT NULL` is rendered as `payload BLOB`, and replaying the DDL changes the schema. `data_type_to_sql` should preserve `NOT NULL` where the Paimon type is non-nullable.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,119 @@ impl PaimonTableProvider {
...
+        DataType::Char(t) => format!("CHAR({})", t.length()),
+        DataType::VarChar(t) => format!("VARCHAR({})", t.length()),
+        DataType::Date(_) => "DATE".to_string(),
+        DataType::Time(t) => format!("TIME({})", t.precision()),
+        DataType::Timestamp(t) => format!("TIMESTAMP({})", t.precision()),
+        DataType::LocalZonedTimestamp(t) => {
+            format!("TIMESTAMP_LTZ({})", t.precision())
+        }
+        DataType::Array(t) => format!("ARRAY<{}>", data_type_to_sql(t.element_type())),
+        DataType::Map(t) => format!(
```
QuakeWang: The generated MAP syntax does not match what `SQLContext` currently parses. `sqlparser` accepts `Map(key_type, value_type)` here, while this emits `MAP<key_type, value_type>`, so `SHOW CREATE TABLE` returns non-replayable DDL for map columns.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,119 @@ impl PaimonTableProvider {
...
+            "MAP<{}, {}>",
+            data_type_to_sql(t.key_type()),
+            data_type_to_sql(t.value_type())
+        ),
+        DataType::Multiset(t) => format!("MULTISET<{}>", data_type_to_sql(t.element_type())),
+        DataType::Row(t) => {
```
QuakeWang, line 224: ROW is not the syntax accepted by the current CREATE TABLE path. `SQLContext` maps `SqlType::Struct` to Paimon Row, so this should emit `STRUCT<field type, ...>` instead of `ROW<field: type, ...>`, otherwise row/struct tables cannot round-trip.

crates/integrations/datafusion/tests/sql_context_tests.rs
```diff
@@ -1262,3 +1262,163 @@ async fn test_multiple_temporary_views_in_same_database() {
...
+             g FLOAT, \
+             h DATE, \
+             i TIMESTAMP(3), \
+             j BLOB) \
+             WITH ('data-evolution.enabled' = 'true')",
+        )
+        .await
+        .expect("CREATE TABLE should succeed");
+
+    let definition = collect_definition(&sql_context, "paimon.test_db.t").await;
```
QuakeWang, line 1410: These tests only check substrings. Please add a round-trip assertion that executes the returned definition, especially for NOT NULL, MAP, and STRUCT columns.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in 4795977 — `data_type_to_sql` now preserves `NOT NULL` for non-nullable types. A follow-up (c0ffdf6) moved the `NOT NULL` suffix out of `data_type_to_sql` and into `build_table_definition`, so it only renders at the column level and doesn't leak into nested type arguments like `MAP(INT NOT NULL, ...)`.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in 8dfcb93 — `MAP<...>` changed to `MAP(key_type, value_type)` to match what `SQLContext` parses.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in f3b3c0d — `ROW<name: type, ...>` changed to `STRUCT<name type, ...>` to match `SqlType::Struct` → Paimon Row mapping.

crates/integrations/datafusion/tests/sql_context_tests.rs
shyjsarah: Added in ff4f2e1 — `test_show_create_table_round_trip` creates a table with NOT NULL, ARRAY, MAP, STRUCT, and nested composite types, drops it, re-executes the rendered DDL, and asserts schema equivalence (fields, primary keys, partition keys) via `assert_schema_equivalent`. This caught the NOT NULL nesting bug fixed in c0ffdf6.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,131 @@ impl PaimonTableProvider {
...
+///
+/// Mirrors the syntax accepted by `SQLContext::handle_create_table`:
+/// `CREATE TABLE <db>.<table> (<col> <type>, ..., PRIMARY KEY (...)) [PARTITIONED BY (...)] [WITH ('k'='v', ...)]`.
+fn build_table_definition(table: &Table) -> String {
+    let identifier = table.identifier();
+    let schema = table.schema();
+    let mut ddl = String::new();
+    let _ = write!(
+        ddl,
+        "CREATE TABLE {}.{} (",
```
JingsongLi, line 115: `SHOW CREATE TABLE` should emit replayable SQL, but the table name is written without identifier quoting. A table created with quoted identifiers such as `CREATE TABLE paimon.test_db."select" ("order" INT)` would be rendered as `CREATE TABLE test_db.select (order INT)`, which is invalid or changes the identifier semantics when re-executed. The same issue applies to column names, primary/partition keys, and nested struct field names below. Please quote identifiers when required and escape embedded quotes, then add a round-trip case with a reserved-word or otherwise quoted identifier.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,131 @@ impl PaimonTableProvider {
...
+    }
+
+    let options = schema.options();
+    if !options.is_empty() {
+        ddl.push_str(" WITH (");
+        for (i, (k, v)) in options.iter().enumerate() {
+            if i > 0 {
+                ddl.push_str(", ");
+            }
+            let _ = write!(ddl, "'{}' = '{}'", k, v);
```
JingsongLi: This needs SQL string escaping before interpolation. Option values are arbitrary table metadata, so a value containing a single quote, for example `WITH ('comment' = 'Bob's table')`, makes the returned definition invalid when users copy or re-execute `SHOW CREATE TABLE`. Please render string literals with SQL escaping (doubling `'` to `''`) for both keys and values, and add a round-trip test with an option value containing a quote.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -89,6 +97,131 @@ impl PaimonTableProvider {
...
+        DataType::Decimal(t) => format!("DECIMAL({}, {})", t.precision(), t.scale()),
+        DataType::Double(_) => "DOUBLE".to_string(),
+        DataType::Float(_) => "FLOAT".to_string(),
+        DataType::Binary(t) => format!("BINARY({})", t.length()),
+        DataType::VarBinary(t) => format!("VARBINARY({})", t.length()),
+        DataType::Blob(_) => "BLOB".to_string(),
+        DataType::Char(t) => format!("CHAR({})", t.length()),
+        DataType::VarChar(t) => format!("VARCHAR({})", t.length()),
+        DataType::Date(_) => "DATE".to_string(),
+        DataType::Time(t) => format!("TIME({})", t.precision()),
```
JingsongLi: This arm (and the `TIMESTAMP_LTZ`, `MULTISET`, and `VECTOR` arms below) emits syntax that the current `SQLContext` cannot round-trip. `sql_data_type_to_paimon_type` has no `SqlType::Time` branch and falls through to `Unsupported SQL data type`; similarly `TIMESTAMP_LTZ`, `MULTISET`, and `VECTOR` are not accepted by that converter. A table loaded from existing Paimon metadata with any of these types will therefore return a `SHOW CREATE TABLE` definition that cannot be executed by paimon-rust. Please either add parser/converter support for the emitted syntax with round-trip tests, or avoid advertising these variants as replayable DDL.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in 78c15c0. `SHOW CREATE TABLE` now quotes database/table names, column names, primary key columns, partition columns, and nested STRUCT field names, escaping embedded double quotes through the standard doubled-quote form.

I also added `test_show_create_table_round_trip_with_quoted_identifiers_and_options`, which covers reserved-word identifiers such as `"select"`, `"order"`, `"group"`, and nested `"from"`, then re-executes the generated DDL and checks schema equivalence.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in 78c15c0. `TIMESTAMP_LTZ` is now rendered as `TIMESTAMP(p) WITH TIME ZONE`, which the current converter already maps back to `LocalZonedTimestamp`.

For `TIME`, `MULTISET`, and `VECTOR`, `SHOW CREATE TABLE` now returns an explicit unsupported error instead of emitting non-replayable DDL. I added `test_show_create_table_rejects_non_round_trippable_types` to cover those cases.

crates/integrations/datafusion/src/table/mod.rs
shyjsarah: Fixed in 78c15c0. Option keys and values are now rendered as SQL string literals with single quotes escaped by doubling them. I also sort options before rendering to keep the output deterministic.

The new round-trip test includes an option value `Bob's table` and verifies the generated DDL contains `Bob''s table` and can be executed again.

JingsongLi:
I think SHOW CREATE TABLE is not fully lossless for bounded char/binary types created outside the DataFusion SQL path.

`table/mod.rs` renders `BINARY(n)`, `VARBINARY(n)`, `CHAR(n)`, and `VARCHAR(n)` with their lengths. However, `sql_context.rs::sql_data_type_to_paimon_type` currently maps all string-like SQL types to `VarChar(MAX_LENGTH)` and all binary-like SQL types to `VarBinary(MAX_LENGTH)`, so re-executing the generated DDL does not recreate the same Paimon schema for tables created via catalog API / Java metadata with bounded `CharType`, `BinaryType`, or bounded `VarBinaryType`.

The existing round-trip tests use SQL-created tables, so they do not catch this case because the original schema already goes through the lossy SQL parser.

Could we either:
1. preserve char/binary type and length in the CREATE TABLE parser, or
2. reject these non-round-trippable types in SHOW CREATE TABLE, similar to TIME/MULTISET/VECTOR?

Otherwise SHOW CREATE TABLE may return DDL that executes successfully but recreates a different schema.

shyjsarah:
Fixed in fcfc2c6 by preserving bounded string and binary types in the CREATE TABLE converter.

`CHAR(n)`, `VARCHAR(n)`, `BINARY(n)`, and `VARBINARY(n)` now round-trip to the corresponding Paimon types with their lengths instead of being widened to max-length varchar/varbinary. I also extended the round-trip test to create bounded char/binary columns through the catalog API, so it covers the path where the original schema did not go through DataFusion SQL first.

JingsongLi (approved):

PR #445, opened 2026-07-03
[python] Push down like/startsWith/endsWith/contains predicates

JunRuiLee:
### Purpose

The Python binding's `METHOD_NOT_SUPPORTED` list rejects `like` / `startsWith` / `endsWith` / `contains` with `NotImplementedError`, even though the core `PredicateBuilder` has supported `StartsWith`/`EndsWith`/`Contains`/`Like` (and the stats/reader-pruning evaluators handle them) since #425. Any query with a string-prefix/suffix/substring filter therefore cannot be pushed down through the DataFrame-style read path. This PR is pure binding wiring — zero core change.

Linked issue: none.

### Brief change log

- Remove the four string operators from `METHOD_NOT_SUPPORTED` (`not` remains rejected) and add match arms in `leaf_to_predicate`:
  - `startsWith` / `endsWith` / `contains`: exactly 1 string literal → `PredicateBuilder::{starts_with,ends_with,contains}`. An empty pattern folds to `IsNotNull` in the core builder.
  - `like`: `[pattern]` or `[pattern, escape]`, mirroring SQL `LIKE .. ESCAPE ..`. The escape must be a single character; the core accepts only `'\\'` and rejects anything else. Prefix / suffix / contains-shaped patterns are rewritten by the core's LIKE optimization (`ab%` → StartsWith); residual patterns (`_`, multi-segment `%`, escaped wildcards) stay as a `Like` leaf.
- String operators now follow the normal leaf path: unknown field or non-string column raises `ValueError` (previously the blanket `NotImplementedError` preceded field resolution). The two tests asserting that precedence are updated accordingly; `not` keeps its precedence behavior.

### Tests

- Rust: `cargo test -p pypaimon_rust` — 38 passed. New unit tests cover each operator's leaf conversion, LIKE prefix-optimization and residual shapes, escape acceptance/rejection (non-backslash, multi-char, 3 literals), empty-pattern → IsNotNull folding, non-string column rejection, and literal-count validation.
- Python: `uv run --no-sync pytest tests/test_read.py` — 39 passed. New end-to-end tests build a two-file string table and assert `startsWith` prunes splits via file stats and each operator returns exactly the matching rows through `TableRead.read`, including a `b_nana` residual-LIKE evaluator case.

### API and Format

No API surface change (the same `with_filter(dict)` accepts more operators). No storage format change.

### Documentation

Behavior documented inline on the new match arms and `escape_char` helper.

### DIFF
--- a/bindings/python/src/predicate.rs
+++ b/bindings/python/src/predicate.rs
@@ -105,7 +105,7 @@ fn float_val(value: &Bound<'_, PyAny>) -> PyResult<f64> {
 
 /// Operators recognized by the lightweight dict format but not translatable to a
 /// Rust [`Predicate`] for pushdown.
-const METHOD_NOT_SUPPORTED: &[&str] = &["like", "startsWith", "endsWith", "contains", "not"];
+const METHOD_NOT_SUPPORTED: &[&str] = &["not"];
 
 /// Recursively convert a lightweight dict predicate into a Rust [`Predicate`].
 ///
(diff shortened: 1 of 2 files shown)


QuakeWang (approved):
+1

PR #446, opened 2026-07-03
Implement global index search modes

JingsongLi:
## Summary

This PR aligns Rust global-index search behavior with Java by adding `global-index.search-mode` and applying it across scalar, vector, and full-text global index search paths.

## Changes

- Add `global-index.search-mode` parsing with `fast`, `full`, and `detail` modes.
- Track indexed row-id coverage for scalar BTree global indexes and include unindexed row ranges in `full`/`detail` modes.
- Extend vector search to merge indexed results with raw fallback results, including raw metric resolution and Java-compatible scoring for `l2`, `cosine`, and `inner_product`.
- Extend full-text search to fill raw gaps when full-text index coverage is incomplete.
- Return an empty DataFusion execution plan for empty full-text search results instead of scanning the full table.

## Testing

- [x] `cargo fmt`
- [x] `cargo check -p paimon`
- [x] `cargo check -p paimon --features fulltext`
- [x] `cargo check -p paimon-datafusion --features fulltext`
- [x] `cargo test -p paimon global_index --lib`
- [x] `cargo test -p paimon table_scan --lib`
- [x] `cargo test -p paimon vector_search_builder --lib`
- [x] `cargo test -p paimon --features fulltext full_text --lib`

## Notes

The vector path follows Java's behavior of allowing raw fallback in `full`/`detail` mode even when no vector index files exist. The full-text path only raw-fills gaps when full-text index files are present, matching Java's scan behavior.

### DIFF
--- a/crates/paimon/src/table/full_text_search_builder.rs
+++ b/crates/paimon/src/table/full_text_search_builder.rs
@@ -121,17 +121,17 @@ impl<'a> FullTextSearchBuilder<'a> {
             None => return Ok(Vec::new()),
         };
 
-        let index_manifest_name = match snapshot.index_manifest() {
-            Some(name) => name.to_string(),
-            None => return Ok(Vec::new()),
+        let index_entries = match snapshot.index_manifest() {
+            Some(index_manifest_name) => {
+                let manifest_path = format!(
+                    "{}/manifest/{}",
+                    self.table.location().trim_end_matches('/'),
+                    index_manifest_name
+                );
+                IndexManifest::read(self.table.file_io(), &manifest_path).await?
+            }
+            None => Vec::new(),
         };
-
-        let manifest_path = format!(
-            "{}/manifest/{}",
-            self.table.location().trim_end_matches('/'),
-            index_manifest_name
-        );
-        let index_entries = IndexManifest::read(self.table.file_io(), &manifest_path).await?;
 
         evaluate_full_text_search(
             Some(self.table),
@@ -179,7 +179,7 @@ async fn evaluate_full_text_search(
         })
         .collect();
 
-    if fulltext_entries.is_empty() {
+    if fulltext_entries.is_empty() && search_mode == GlobalIndexSearchMode::Fast {
         return Ok(Vec::new());
     }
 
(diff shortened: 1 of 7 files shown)


crates/paimon/src/table/full_text_search_builder.rs
```diff
@@ -166,38 +183,255 @@ async fn evaluate_full_text_search(
...
+            })
+            .collect();
+
+        let results = futures::future::try_join_all(futures).await?;
+        for r in &results {
+            merged = merged.or(r);
+        }
+    }
+
+    if search_mode != GlobalIndexSearchMode::Fast {
```
leaves12138, line 227: This raw fallback is currently unreachable when `fulltext_entries` is empty, because the function returns `Ok(Vec::new())` earlier. In `full`/`detail` mode, no Tantivy coverage for this field means all rows are unindexed and should be searched through the raw full-text path, matching `GlobalIndexCoverage` semantics and the vector path (`vector_entries.is_empty() && search_mode == Fast` is the only early return there). As written, a table configured with `global-index.search-mode=full` or `detail` but with no full-text index for `body` returns `[]` instead of matching raw rows. The same issue exists one layer above when `execute` returns early on a snapshot without an index manifest. Could we make these early returns apply only to `Fast`, route the empty/no-manifest case through `unindexed_ranges_for_global_index_entries` + `read_raw_full_text_search`, and add a regression test similar to the vector raw-path test?

leaves12138 (changes requested):
Thanks for implementing the search modes. The BTree/table-scan path and the vector raw fallback look consistent with the intended full/detail semantics, and the focused tests/format checks pass locally. I found one correctness issue in the full-text path: full/detail mode still returns no rows when the target full-text index is absent, instead of searching the unindexed raw data.

JingsongLi:
Addressed the full-text review feedback in 47859ac: full/detail mode now routes no-manifest/no-fulltext-entry cases through raw full-text search instead of returning empty, and DataFileReader now materializes _ROW_ID for files with first_row_id so raw global-index fallback can produce correct row ranges. Added an end-to-end regression test for full-text raw search without an index manifest.\n\nVerification run:\n- cargo fmt\n- cargo check -p paimon\n- cargo check -p paimon --features fulltext\n- cargo check -p paimon-datafusion --features fulltext\n- cargo test -p paimon --features fulltext full_text --lib\n- cargo test -p paimon data_file_reader --lib\n- cargo test -p paimon table_scan --lib\n- cargo test -p paimon vector_search_builder --lib\n- cargo test -p paimon global_index --lib

leaves12138 (approved):
Thanks for the update. The previous full-text correctness concern is addressed: the no-index/no-manifest cases now flow through the raw full-text path in full/detail mode, and the new regression tests cover the scenarios I was worried about. I also checked the new `_ROW_ID` projection handling in the data-file reader. Locally, cargo fmt, the focused full-text/global-index/vector/data-file-reader tests pass. LGTM from code review; the CI check job is still failing and should be handled before merge.

JingsongLi:
Fixed the failing CI check in 6e32ac2. The failure was clippy-only: three evaluator functions exceeded the argument limit and one table-scan helper tripped type_complexity. I refactored the evaluator parameters into small context structs and added a type alias for the scan grouping map.\n\nLocal verification now includes the exact CI check command:\n- cargo fmt --all\n- cargo clippy --all-targets --workspace --features fulltext,vortex,mosaic -- -D warnings\n- cargo test -p paimon --features fulltext full_text --lib\n- cargo test -p paimon global_index --lib\n- cargo test -p paimon vector_search_builder --lib\n- cargo test -p paimon table_scan --lib

PR #447, opened 2026-07-03
[table] Fail closed when reading a query-auth.enabled table

plusplusjiajia:
### Purpose

  A `query-auth.enabled` table needs a per-user row filter / column masking that
  the Rust client can't yet enforce. Reading it would return unfiltered data, so
  this makes reads fail closed.

  ### Brief change log

  - Add `CoreOptions::query_auth_enabled()`.
  - Guard `TableRead::to_arrow` (the single row-data exit) to return
    `Error::Unsupported` when enabled — non-bypassable for any construction path.

  ### Tests

  - Builder path and direct `TableRead::new(..)` path both fail closed
    (`read_builder` + `table_read` unit tests).

  ### API and Format

  Adds `CoreOptions::query_auth_enabled()`.

### DIFF
--- a/crates/paimon/src/table/table_read.rs
+++ b/crates/paimon/src/table/table_read.rs
@@ -72,6 +72,7 @@ impl<'a> TableRead<'a> {
 
     /// Returns an [`ArrowRecordBatchStream`].
     pub fn to_arrow(&self, data_splits: &[DataSplit]) -> crate::Result<ArrowRecordBatchStream> {
+        let has_primary_keys = !self.table.schema.primary_keys().is_empty();
         let core_options = CoreOptions::new(self.table.schema.options());
         // Fail closed: this client can't yet enforce query-auth row filtering / column masking,
         // so refuse to read rather than return unfiltered data. Guarding this single data exit
@@ -84,8 +85,6 @@ impl<'a> TableRead<'a> {
                     .to_string(),
             });
         }
-
-        let has_primary_keys = !self.table.schema.primary_keys().is_empty();
         let merge_engine = core_options.merge_engine()?;
 
         // PK table with Deduplicate engine: splits that may hold multiple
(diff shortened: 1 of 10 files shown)


crates/paimon/src/table/table_read.rs
```diff
@@ -72,8 +72,20 @@ impl<'a> TableRead<'a> {
 
     /// Returns an [`ArrowRecordBatchStream`].
     pub fn to_arrow(&self, data_splits: &[DataSplit]) -> crate::Result<ArrowRecordBatchStream> {
-        let has_primary_keys = !self.table.schema.primary_keys().is_empty();
         let core_options = CoreOptions::new(self.table.schema.options());
+        // Fail closed: this client can't yet enforce query-auth row filtering / column masking,
+        // so refuse to read rather than return unfiltered data. Guarding this single data exit
+        // (not the `ReadBuilder`) makes the check non-bypassable.
+        if core_options.query_auth_enabled() {
```
QuakeWang: This guards row materialization, but scan planning still succeeds for a `query-auth.enabled` table. `TableScan::plan()` / `plan_with_trace()` can still expose splits, file metadata, row counts, and stats before `to_arrow()` is reached, and DataFusion also plans eagerly. To make this truly fail closed, please add the same `Unsupported` guard at the scan planning boundary as well.

plusplusjiajia: @QuakeWang Good catch. Added the same Unsupported guard at scan planning (plan() / plan_with_trace()), gated on !scan_all_files so only reads are blocked (writes/compaction stay allowed).

plusplusjiajia: Update: dropped the !scan_all_files exemption — it wasn't a clean read/write split (the files system table also uses it to surface metadata). plan() now fails closed unconditionally, so writes/compaction are blocked too until the client can enforce the auth filter.

QuakeWang (approved):
LGTM

PR #448, opened 2026-07-03
Apply exact residual filtering for ORC, Avro, and Row reads

JunRuiLee:
### Purpose

Make predicate evaluation exact across all file formats, so a consumer that returns scan batches verbatim gets exactly the rows matching the pushed-down predicate. Motivated by pypaimon delegating its read kernel to the Rust core: it returns the scan's Arrow batches directly without re-filtering, which was only safe for Parquet.

Before this change:
- **ORC** pruned by stripe stats only — non-matching rows in a selected stripe survived.
- **Avro** / **Row** ignored the predicate entirely.

These caused silent over-reads (wrong data, no error). Now every format applies the predicate exactly. Exact `LIMIT` remains the caller's responsibility (matching `with_limit`'s documented hint semantics).

### Changes

- Consolidate the two duplicate Arrow-batch predicate evaluators (Parquet's and Vortex's) into a shared `arrow::residual` module; the Vortex copy previously fell open on `startsWith`/`like`/`between`, now evaluated exactly everywhere.
- Apply the residual filter in ORC (after stripe pruning), Avro, and Row.
- Readers scan predicate columns (`widen_scan_fields`) so a predicate on a non-projected column is applied; `DataFileReader` projects to the requested output by name. Residual columns are resolved by name, not batch position (ORC emits file-schema order).
- Parquet keeps its zero-overhead `RowFilter` fast path when all predicates are fully enforced, else applies the residual pass (covers `Or`/`Not`/unsupported leaves).
- Reject `_ROW_ID` projection combined with a data predicate (`Unsupported`) — positional row-id assignment desyncs when rows are filtered.

### Tests

`cargo test -p paimon --lib` (+`--features vortex`) all pass. Covers per-format exactness incl. non-projected predicate columns, `Or`/`Not`, misordered batch columns, and the `_ROW_ID` guard.

No public API or storage-format change.


crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -223,7 +235,32 @@ impl FormatFileReader for ParquetFormatReader {
...
         let batch_stream = batch_stream_builder.build()?;
-        Ok(batch_stream.map(|r| r.map_err(Error::from)).boxed())
+
+        if all_enforced {
+            // Fast path: the row filter enforced every predicate exactly during
+            // decode. Return the stream directly — no residual pass.
+            return Ok(batch_stream.map(|r| r.map_err(Error::from)).boxed());
+        }
+
+        // Residual backstop: at least one predicate (`Or`/`Not`/unsupported leaf)
```
QuakeWang, line 284: This residual backstop only works when the compound predicate reaches the format reader. The normal ReadBuilder/TableRead path still passes data predicates through reader_pruning_predicates(), which drops And/Or/Not, so public reads will not actually get exact filtering for these predicates. Please keep the full read predicate for residual filtering and use a separate pruned copy only for stats/pushdown.

crates/paimon/src/arrow/residual.rs
```diff
@@ -0,0 +1,1184 @@
...
+            let Some(scalar) = literal_scalar_for_arrow_filter(literal, data_type)
+                .map_err(|e| ArrowError::ComputeError(e.to_string()))?
+            else {
+                // The literal cannot be converted to an Arrow scalar for this
+                // column type (e.g. a decimal literal whose scale differs from
+                // the column). Erroring is required for correctness: returning
+                // all-true here would silently pass every row (a wrong-read),
+                // and this residual pass is the only place the predicate is
+                // enforced when the row filter did not accept the leaf.
+                return Err(unconvertible_literal_error(op, data_type));
```
QuakeWang, line 354: For schema-evolved files this can turn a valid predicate into a read error. Example: table column promoted INT -> BIGINT, old file column is INT, predicate value > i32::MAX. literal_scalar_for_arrow_filter() returns None for the old INT column, but the exact result for that file is simply no rows, not DataInvalid. This needs boundary handling/tests for promoted numeric types.

crates/paimon/src/arrow/format/mod.rs
```diff
@@ -54,8 +54,13 @@ pub(crate) struct FilePredicates {
...
 #[async_trait]
 pub(crate) trait FormatFileReader: Send + Sync {
-    /// Read a single data file, returning a stream of RecordBatches
-    /// containing only the projected columns (using names from the file's schema).
+    /// Read a single data file, returning a stream of RecordBatches containing
+    /// at least the projected columns (using names from the file's schema). A
+    /// reader MAY include extra columns it needed to scan (e.g. predicate columns
+    /// for residual filtering); the caller (`DataFileReader`) projects to the
+    /// requested output by name, so extra columns are harmless. Each batch MUST
+    /// already have the pushed-down predicate applied exactly — residual filtering
```
QuakeWang: This contract is currently stronger than the implementations. Blob still ignores predicates, and Mosaic only does row-group stats pruning without row-level residual filtering. Either the contract should be narrowed to the formats changed here, or the remaining readers need to reject/apply predicates exactly.

crates/paimon/src/arrow/format/parquet.rs
JunRuiLee: Fixed. `ReadBuilder`/`TableRead` were pruning `And`/`Or`/`Not` out of the data predicate before the reader, so the residual pass never saw compound predicates on the public path. Rather than carry a separate pruned copy, I just stopped pruning — pushdown/stats already skip compound nodes they can't use, so the reader now gets the full predicate and filters exactly. Added a regression test through the real `ReadBuilder` -> `to_arrow` path.

crates/paimon/src/arrow/residual.rs
JunRuiLee: Good catch, fixed. The residual now evaluates the leaf against the predicate's declared type instead of the file column's type: it casts the decoded column up to the (wider, promoted) type first, so an out-of-range value on an old narrower column yields no rows rather than an error. Added tests for the INT->BIGINT boundary (out-of-range -> 0 rows, in-range -> exact).

crates/paimon/src/arrow/format/mod.rs
JunRuiLee: Fixed — narrowed the doc. It now states exactness per format (Parquet/ORC/Avro/Row/Vortex apply the predicate exactly; Blob ignores predicates and Mosaic does stats-only pruning) instead of claiming a blanket guarantee.

crates/paimon/src/arrow/residual.rs
```diff
@@ -0,0 +1,1247 @@
...
+            _ => unreachable!(),
+        };
+        let mask = sanitize_filter_mask(evaluate_column_predicate(array, &scalar, comparison_op)?);
+        combined = combine_filter_masks(&combined, &mask, matches!(op, PredicateOperator::In));
+    }
+
+    Ok(combined)
+}
+
+fn evaluate_column_predicate(
```
QuakeWang, line 569: This uses Arrow's generic comparison kernels for Binary/VarBinary as well, but Paimon's Datum::Bytes ordering is Java signed-byte order (`0xFF < 0x00`). Arrow compares byte slices unsigned, so binary range predicates can return the wrong rows. Please add a Binary/VarBinary-specific comparison path that matches Paimon's comparator, with a residual test.

crates/paimon/src/arrow/residual.rs
```diff
@@ -0,0 +1,1247 @@
...
+            Datum::LocalZonedTimestamp { millis, nanos } => {
+                let Some(array) = timestamp_scalar(*millis, *nanos, ts.precision(), Some("UTC"))?
+                else {
+                    return Ok(None);
+                };
+                array
+            }
+            _ => return Ok(None),
+        },
+        DataType::Decimal(decimal) => match literal {
```
QuakeWang: This rejects decimal literals whose scale differs from the file column scale, but Paimon Datum compares decimals by mathematical value across scales. A valid predicate like DECIMAL 1.0 against a DECIMAL(…, 2) column can now fail during residual filtering. Please rescale exactly when possible, or evaluate decimal predicates with the same semantics as Datum.

crates/paimon/src/table/data_file_reader.rs
```diff
@@ -69,6 +69,26 @@ impl DataFileReader {
...
 
+    /// Reject projecting `_ROW_ID` alongside a data predicate. `_ROW_ID` is
+    /// assigned positionally from post-filter batch row counts, so a residual
+    /// filter that drops rows would desync it. (`_ROW_ID` predicates travel via
+    /// `row_ranges`, not `predicates`, so they do not trip this.)
+    fn reject_row_id_with_predicate(
+        read_type: &[DataField],
+        predicates: &[Predicate],
+    ) -> crate::Result<()> {
+        let projects_row_id = read_type
```
QuakeWang, line 80: This guard treats any predicate as row-dropping. Constant predicates such as AlwaysTrue cannot desync _ROW_ID but are rejected here. Consider normalizing away no-op predicates, or only rejecting predicates that can actually filter rows.

crates/paimon/src/arrow/residual.rs
JunRuiLee: Fixed. Binary/VarBinary comparisons (ordering and Between) now go through Paimon's signed-byte comparator (java_bytes_cmp), matching Datum::Bytes (0xFF < 0x00) instead of Arrow's unsigned order. Between delegates to the shared comparison path so it can't diverge. Regression tests added for both ordering and Between.

crates/paimon/src/arrow/residual.rs
JunRuiLee: Fixed. Decimal value predicates are now evaluated row-wise via datum_cmp, so cross-scale comparison is by mathematical value (e.g. DECIMAL 1.0 matches a DECIMAL(_,2) column, and d > 1.05 on a DECIMAL(_,1) column is exactly d >= 1.1) — no rescale requirement and no spurious error. Tests cover the equal-value cross-scale and finer-scale-literal cases.

crates/paimon/src/table/data_file_reader.rs
JunRuiLee: Fixed. The guard now only rejects predicates that can actually filter rows — a constant AlwaysTrue is allowed.

Broader note: the residual filter this PR adds makes format-reader reads exact, but the data-evolution read path still passes no predicate to the reader, so index/fallback row-ranges aren't residual-filtered there. That gap is orthogonal to this PR and to #449's fallback-coverage fix; I'll follow up on it separately.

crates/paimon/src/table/read_builder.rs
```diff
@@ -233,10 +232,13 @@ impl<'a> ReadBuilder<'a> {
...
         };
 
+        // Pass the FULL data predicate through (including `And`/`Or`/`Not`).
+        // Pushdown/stats skip compound nodes; the residual pass enforces the full
+        // predicate exactly. Pruning here would drop compound predicates.
         Ok(TableRead::new(
             self.table,
             read_type,
-            reader_pruning_predicates(self.filter.data_predicates.clone()),
+            self.filter.data_predicates.clone(),
```
JingsongLi, line 246: This still does not make the public read path exact for `data_evolution_enabled` tables. `new_read()` now preserves the full data predicate here, but `TableRead::read_with_evolution()` builds `DataEvolutionReader` without passing `self.data_predicates`, and `DataEvolutionReader` creates every `DataFileReader` with `Vec::new()`. `TableScan` also disables normal data-predicate pushdown for data-evolution tables (`pushdown_data_predicates = &[]`), so a filtered scan without a usable global index, or with only coarse row ranges, will still return non-matching rows from the data-evolution reader. This regresses the PR goal that scan batches are exactly filtered; the evolution reader needs to receive and apply the same residual predicates after column merge (or before when it is safe).

JunRuiLee: Fixed. `TableRead::read_with_evolution` now passes the data predicates into `DataEvolutionReader`, which filters every batch exactly on both branches (raw-convertible and column-merge): the read set is widened with predicate columns, `_ROW_ID` is attached first, then the residual filter runs and the batch is projected back to the caller's read type. Inner file readers still receive no predicates, so the two branches stay consistent.

Tests go through the real `ReadBuilder` -> `to_arrow` path: leaf and `Or` predicates, cross-file predicate on the merge branch, `_ROW_ID` + predicate (original ids survive), and null-fill semantics (comparisons drop, `IS NULL` keeps).

JingsongLi (approved):
+1

PR #449, opened 2026-07-04
fix(scan): use evaluated fields for global index fallback

QuakeWang:
### Purpose

Global index coverage currently includes `extra_field_ids`, but Rust can only read BTree entries by `index_field_id`. For predicates like `primary AND extra`, the scanner may evaluate only the primary-field conjunct while fallback coverage is computed from all predicate fields, including an extra field that was not actually evaluated.

This PR tracks which fields were actually evaluated by the global index scanner and uses only those fields when adding raw fallback ranges. Extra-field-only predicates continue to fall back until composite-key BTree reads are supported.

### Brief change log

  - Return evaluated field ids together with global index row ranges.
  - Compute raw fallback ranges from evaluated fields only.
  - Add regression tests for extra-field-only fallback and mixed primary/extra predicates.

### Tests

### DIFF
--- a/crates/paimon/src/table/global_index_scanner.rs
+++ b/crates/paimon/src/table/global_index_scanner.rs
@@ -36,13 +36,19 @@ use std::sync::Mutex;
 
 type BoxedCmp = Box<dyn Fn(&[u8], &[u8]) -> Ordering + Send + Sync>;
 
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #450, opened 2026-07-04
Add batch vector search core

JingsongLi:
## Summary

Adds a Rust core batch vector search API so multiple query vectors can be evaluated in one call while preserving one result per input vector. The existing single-vector search builder now delegates to the batch path, keeping the DataFusion UDTF behavior unchanged.

## Changes

- Introduce `BatchVectorSearchBuilder` and `Table::new_batch_vector_search_builder()`.
- Refactor vector search evaluation to merge and top-k results independently per query vector.
- Add batch raw fallback scanning so unindexed row ranges are read once for multiple query vectors.
- Add backend batch hooks for Lumina and vindex; Lumina uses its native `n > 1` search path when possible and falls back to per-vector search for filtered or mixed-limit queries.

## Testing

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p paimon table::vector_search_builder`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_java_vindex_table`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_without_matching_index_returns_empty`

## Notes

The current DataFusion `vector_search` table function signature and behavior are unchanged. A true Spark-style lateral/table-function join over outer-row vectors should be implemented as a follow-up logical/physical DataFusion extension on top of this core batch API.

The full local `vector_search_tests` suite could not complete because the two Lumina query tests require `liblumina_py.dylib`; the vindex and missing-index tests pass, and the Lumina build test is already ignored unless `LUMINA_LIB_PATH` is set.

### DIFF
--- a/crates/paimon/src/lumina/reader.rs
+++ b/crates/paimon/src/lumina/reader.rs
@@ -122,6 +122,15 @@ impl LuminaVectorGlobalIndexReader {
         self.search(vector_search)
     }
 
+    pub fn visit_batch_vector_search<S: Read + Seek + Send + 'static>(
+        &mut self,
+        vector_searches: &[VectorSearch],
+        stream_fn: impl FnOnce(&str) -> crate::Result<S>,
+    ) -> crate::Result<Vec<Option<HashMap<u64, f32>>>> {
+        self.ensure_loaded(stream_fn)?;
+        self.search_batch(vector_searches)
+    }
+
     fn search(&self, vector_search: &VectorSearch) -> crate::Result<Option<HashMap<u64, f32>>> {
         let index_meta = self
             .index_meta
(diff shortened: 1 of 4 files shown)


leaves12138 (approved):
LGTM. The batch vector search API keeps the single-vector path as a thin wrapper over the batch implementation, and the evaluator preserves per-query result ordering while sharing index/raw scans where possible. The Lumina fast path is used only when there is no row-id filter and all limits match, with sensible fallback to per-query search for filtered or mixed-limit cases; VIndex currently uses the safe per-query path.

Validated locally with:
- `cargo test -p paimon vector_search_builder --all-features`
- `cargo test -p paimon 'lumina::reader' --all-features`
- `cargo test -p paimon 'vindex::reader' --all-features`

GitHub CI is also green on this head.

PR #452, opened 2026-07-04
Support DataFusion lateral vector search joins

JingsongLi:
## Summary

Adds DataFusion support for `CROSS JOIN LATERAL vector_search(...)` when the query vector comes from the left input column, planning it as batched vector search per input batch. The existing literal JSON `vector_search` path remains unchanged.

## Changes

- Registers a custom optimizer rule and query planner in `SQLContext` to rewrite supported lateral joins into `LateralVectorSearchExec`.
- Extends the `vector_search` UDTF with a lateral marker provider for column query vector arguments.
- Executes batch vector search per left record batch, reads target rows by `_ROW_ID`, and emits joined left/right rows.
- Adds DataFusion coverage with per-row query vectors that produce different top-k matches.

## Testing

- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_lateral_join_uses_query_vectors -- --nocapture`
- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_java_vindex_table -- --nocapture`
- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_without_matching_index_returns_empty -- --nocapture`
- [x] `cargo test -p paimon table::vector_search_builder -- --nocapture`
- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo clippy -p paimon-datafusion --test read_tables -- -D warnings`

## Notes

- First version supports `CROSS JOIN LATERAL vector_search(...)` / inner joins with no join predicate. Unsupported shapes continue to fail through the marker provider scan path.
- The lateral query vector currently must be a column expression yielding `List<Float32>` or `FixedSizeList<Float32>`.

### DIFF
--- a/crates/integrations/datafusion/src/lateral_vector_search.rs
+++ b/crates/integrations/datafusion/src/lateral_vector_search.rs
@@ -1,0 +1,698 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::any::Any;
+use std::cmp::Ordering;
+use std::collections::HashMap;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use datafusion::arrow::array::{
+    new_empty_array, Array, ArrayRef, FixedSizeListArray, Float32Array, Int64Array, ListArray,
+    RecordBatch, UInt32Array,
+};
+use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
+use datafusion::catalog::default_table_source::source_as_provider;
+use datafusion::common::stats::Precision;
+use datafusion::common::tree_node::Transformed;
+use datafusion::common::{
+    internal_err, DFSchemaRef, DataFusionError, JoinType, Result as DFResult, Statistics,
+};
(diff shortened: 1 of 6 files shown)


leaves12138 (approved):
Reviewed the lateral `vector_search` implementation. The existing literal JSON `vector_search` path remains intact, and the new lateral path is cleanly routed through a logical rewrite plus a custom physical extension planner. The execution path batches query vectors from the left input, preserves the per-query result mapping, and reads target rows by `_ROW_ID` without falling back to a full table scan.

Validation passed locally:
- `git diff --cached --check`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_lateral_join_uses_query_vectors -- --nocapture`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_java_vindex_table -- --nocapture`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_without_matching_index_returns_empty -- --nocapture`
- `cargo test -p paimon table::vector_search_builder -- --nocapture`
- `cargo clippy -p paimon-datafusion --test read_tables -- -D warnings`

No blocking issues found.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -0,0 +1,696 @@
...
+            target_table: self.target_table.clone(),
+            target_schema: Arc::clone(&self.target_schema),
+            target_column: self.target_column.clone(),
+            query_vector_expr: exprs.into_iter().next().unwrap(),
+            limit: self.limit,
+            schema: Arc::clone(&self.schema),
+        }))
+    }
+
+    fn dyn_hash(&self, mut state: &mut dyn Hasher) {
```
QuakeWang, line 280: nit: Should `input` also participate in `dyn_hash` / `dyn_eq`? The node's semantics depend on the child plan too, and DataFusion extension nodes generally expect Eq/Hash to represent the full logical node. Non-blocking, but including it would make the contract tighter.

QuakeWang (approved):
LGTM. The change is well-scoped, reuses the existing batch vector search/read path cleanly, and the added lateral join test covers the main behavior. I only left one non-blocking nit around the extension node Eq/Hash contract.

crates/integrations/datafusion/src/lateral_vector_search.rs
JingsongLi: Fixed by including the child logical plan in both `dyn_hash` and `dyn_eq`. Thanks for the catch.

PR #457, opened 2026-07-05
docs(python): fix README fence and document current read/write APIs

hechao-ustc:
### Purpose

The Python README had a broken markdown fence at the end and was missing documentation for the lower-level `PaimonCatalog → Table → ReadBuilder / WriteBuilder` APIs, which are the primary interface for programmatic read and write when not using SQL. 

### changes
- Fix the broken closing fence                                                                             
- Add "Native Read / Write" examples

### DIFF
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -92,7 +92,7 @@ write_builder.new_commit().commit(commit_messages)
 
 # --- Time travel: read a past version ---
 # Supported options: scan.version, scan.timestamp-millis, scan.snapshot-id, or scan.tag-name
-read_builder_tt = table.new_read_builder({"scan.version": "2"})
+read_builder_tt = table.new_read_builder({"scan.snapshot-id": "1"})
 scan_tt = read_builder_tt.new_scan()
 plan_tt = scan_tt.plan()
 batches_tt = read_builder_tt.new_read().read(plan_tt.splits())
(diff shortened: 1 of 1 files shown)


bindings/python/README.md
```diff
@@ -50,6 +50,57 @@ ctx.sql("DROP TEMPORARY TABLE paimon.default.my_temp")
...
+)
+write_builder = table.new_write_builder()
+writer = write_builder.new_write()
+writer.write_arrow(batch)
+commit_messages = writer.prepare_commit()
+write_builder.new_commit().commit(commit_messages)
+
+# --- Time travel: read a past version ---
+# Supported options: scan.version, scan.timestamp-millis, scan.snapshot-id, or scan.tag-name
+read_builder_tt = table.new_read_builder({"scan.version": "2"})
```
QuakeWang: This says the example reads a past version, but after the preceding native write `scan.version=2` resolves to the latest snapshot and returns the newly written rows too. Could we use `scan.snapshot-id=1` or `scan.version=1` here so the time-travel example actually reads the earlier snapshot?

JingsongLi:
Looks good to me.

bindings/python/README.md
hechao-ustc: Sure, it’s already done.

QuakeWang (approved):
+1

PR #458, opened 2026-07-05
Add Variant support and DataFusion functions

JingsongLi:
## Summary

Adds Java-compatible `VARIANT` support to the Rust implementation and registers Spark-style Variant scalar functions in the DataFusion integration. This covers the end-to-end path from schema type handling and row/Arrow encoding to SQL JSON parsing and path extraction.

## Changes

- Add Paimon `VARIANT` type support, including schema JSON fixtures, Datum support, BinaryRow/BTree encoding, Arrow mapping, and row-format read/write handling for `value` + `metadata` buffers.
- Add a Java-compatible Variant core module for JSON parsing, metadata dictionary construction, path access, JSON null handling, and scalar extraction/casting.
- Register DataFusion scalar UDFs: `parse_json`, `try_parse_json`, `variant_get`, `try_variant_get`, and `is_variant_null`.
- Document Variant table creation, inserts, queries, path syntax, supported cast targets, null semantics, raw `SessionContext` registration, and current limitations.

## Testing

- [x] `cargo test -p paimon variant --lib`
- [x] `cargo test -p paimon-datafusion variant_functions --lib`
- [x] `cargo check -p paimon-datafusion`
- [x] `cargo check --workspace`
- [x] `git diff --check`
- [x] `mkdocs build --strict`

## Notes

This PR implements the common scalar-function surface first. `schema_of_variant`, `schema_of_variant_agg`, `to_variant_object`, `variant_explode`, `variant_explode_outer`, and direct `variant_get` casts to `ARRAY` / `MAP` / `STRUCT` are left for follow-up work.

### DIFF
--- a/crates/integrations/datafusion/src/variant_functions.rs
+++ b/crates/integrations/datafusion/src/variant_functions.rs
@@ -201,7 +201,16 @@ impl ScalarUDFImpl for VariantGetFunc {
         if args.arg_fields.len() != 2 && args.arg_fields.len() != 3 {
             return plan_err(format!("{} expects 2 or 3 arguments", self.name()));
         }
-        let output = variant_get_output_type(args.scalar_arguments.get(2).and_then(|v| *v))?;
+        let output = match args.arg_fields.len() {
+            2 => variant_get_output_type(None)?,
+            3 => {
+                let Some(type_arg) = args.scalar_arguments.get(2).and_then(|v| *v) else {
+                    return plan_err("variant_get type argument must be a string literal");
+                };
+                variant_get_output_type(Some(type_arg))?
+            }
+            _ => unreachable!("argument count checked above"),
+        };
         Ok(Arc::new(Field::new(
             self.name(),
             output.arrow_type().clone(),
@@ -847,5 +856,22 @@ mod tests {
             .await
             .unwrap_err();
         assert!(err.to_string().contains("Invalid Variant cast"));
+    }
+
+    #[tokio::test]
+    async fn variant_get_rejects_non_literal_type_argument() {
+        let ctx = SessionContext::new();
+        register_variant_functions(&ctx);
+        let sql = r#"
+            SELECT variant_get(parse_json('{"age":26}'), '$.age', type_name)
+            FROM (VALUES ('int')) AS t(type_name)
+        "#;
+        let err = match ctx.sql(sql).await {
+            Ok(df) => df.collect().await.unwrap_err(),
+            Err(err) => err,
+        };
+        assert!(err
+            .to_string()
+            .contains("variant_get type argument must be a string literal"));
     }
 }
--- a/crates/integrations/datafusion/src/variant_functions.rs
+++ b/crates/integrations/datafusion/src/variant_functions.rs
@@ -1,0 +1,851 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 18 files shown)


crates/paimon/src/variant.rs
```diff
@@ -0,0 +1,1465 @@
...
+    pub scale: i8,
+}
+
+impl VariantDecimal {
+    pub fn to_plain_string(self) -> String {
+        decimal_to_plain_string(self.unscaled, self.scale, true)
+    }
+}
+
+pub fn validate_payload(value: &[u8], metadata: &[u8]) -> Result<()> {
```
leaves12138, line 313: This validation is used by the row/binary-row writers before accepting external Arrow / buffers, but it only checks the metadata version and total sizes. For example,  and  passes , so a non-null Variant with an empty value buffer can be persisted and only fail later when / reads it. Please validate at least the root value buffer here (and ideally recursively validate object/array offsets and metadata ids), or avoid using this helper at ingestion boundaries where malformed Variant bytes must be rejected.

crates/integrations/datafusion/src/variant_functions.rs
```diff
@@ -0,0 +1,851 @@
...
+
+    fn return_type(&self, _arg_types: &[ArrowDataType]) -> DFResult<ArrowDataType> {
+        internal_err("return_field_from_args should be used for variant_get")
+    }
+
+    fn return_field_from_args(&self, args: ReturnFieldArgs) -> DFResult<FieldRef> {
+        if args.arg_fields.len() != 2 && args.arg_fields.len() != 3 {
+            return plan_err(format!("{} expects 2 or 3 arguments", self.name()));
+        }
+        let output = variant_get_output_type(args.scalar_arguments.get(2).and_then(|v| *v))?;
```
leaves12138: If the third argument is not a scalar literal,  is , so the planner silently chooses the default  return type. I reproduced this with  where : the result schema is still the Variant struct instead of an integer (or a planning error). Since DataFusion needs a fixed output type, please reject non-literal third arguments during planning, rather than treating them as omitted.

leaves12138 (changes requested):
Thanks for the large Variant implementation. The core shape looks good and the targeted Variant tests pass locally, but I found a couple of correctness issues that should be fixed before merging.

crates/paimon/src/variant.rs
leaves12138, line 313: This validation is used by the row/binary-row writers before accepting external Arrow `value`/`metadata` buffers, but it only checks the metadata version and total sizes. For example, `value = []` and `metadata = [0x01]` passes `VariantType::validate_payload`, so a non-null Variant with an empty value buffer can be persisted and only fail later when `VariantRef`/`to_json` reads it. Please validate at least the root value buffer here (and ideally recursively validate object/array offsets and metadata ids), or avoid using this helper at ingestion boundaries where malformed Variant bytes must be rejected.

crates/integrations/datafusion/src/variant_functions.rs
leaves12138: If the third argument is not a scalar literal, `scalar_arguments.get(2)` is `None`, so the planner silently chooses the default `Variant` return type. I reproduced this with `SELECT variant_get(parse_json('{"age":26}'), '$.age', ty) FROM t` where `ty = 'int'`: the result schema is still the Variant struct instead of an integer (or a planning error). Since DataFusion needs a fixed output type, please reject non-literal third arguments during planning, rather than treating them as omitted.

leaves12138 (changes requested):
Thanks for the large Variant implementation. The core shape looks good and the targeted Variant tests pass locally, but I found a couple of correctness issues that should be fixed before merging.

crates/paimon/src/variant.rs
JingsongLi: Addressed in 0194ca4: `validate_payload` now validates the metadata dictionary, root value size, scalar payloads, object metadata ids/key order, and recursively checks object/array child ranges/offsets. I also updated row/binary-row tests to use real Variant payloads and added malformed payload coverage.

crates/integrations/datafusion/src/variant_functions.rs
JingsongLi: Addressed in 0194ca4: `variant_get` now treats the 2-arg form as the only omitted-type case; if the 3rd argument is present but not a scalar string literal, planning fails instead of silently returning Variant. Added a regression test for a column-backed type argument.

leaves12138 (approved):
Thanks for the update. The two blocking issues from my previous review are fixed: Variant payload validation now rejects malformed root/object payloads, and variant_get rejects non-literal type arguments during planning. I reran targeted checks locally and they passed.

PR #459, opened 2026-07-06
feat(variant): support shredding read and write

JingsongLi:
## Summary

Adds Variant shredding support aligned with Java, including configured and inferred shredding schemas. The format-level shredding read/write logic is kept reusable so future formats can plug in a physical writer factory instead of duplicating the state machine.

## Changes

- Add Variant shredding schema construction, inference, cast, and rebuild utilities.
- Add Arrow helpers for configured/inferred physical fields, logical-to-physical batch conversion, and shredded Variant reassembly on read.
- Add reusable `ShreddingFormatWriter` / `ShreddingFormatReader` plus a physical writer factory abstraction.
- Keep Parquet focused on raw Parquet IO and a `ParquetPhysicalWriterFactory`; route table format options into `DataFileWriter` through its constructor.
- Add Parquet roundtrip coverage for configured and inferred Variant shredding schemas.
- Document Variant shredding usage, configured/inferred modes, and table options.

## Testing

- [x] `cargo clippy --all-targets --workspace --features fulltext,vortex,mosaic -- -D warnings`
- [x] `cargo check -p paimon --features mosaic,vortex --lib`
- [x] `cargo fmt --check`
- [x] `git diff --check`

## Notes

No migration is required. Existing non-shredded Variant files continue to read through the no-op path unless a shredded physical shape is detected.

### DIFF
--- a/crates/paimon/src/variant.rs
+++ b/crates/paimon/src/variant.rs
@@ -19,6 +19,11 @@
 
 use std::collections::{HashMap, HashSet};
 
+use crate::spec::{
+    ArrayType, BigIntType, BooleanType, DataField, DataType, DateType, DecimalType, DoubleType,
+    FloatType, LocalZonedTimestampType, RowType, TimestampType, VarBinaryType, VarCharType,
+    VariantType,
+};
 use crate::{Error, Result};
 use base64::{engine::general_purpose, Engine as _};
 
@@ -204,6 +209,10 @@ impl<'a> VariantRef<'a> {
         get_decimal(self.value, self.pos)
     }
 
+    pub fn get_binary(&self) -> Result<&'a [u8]> {
+        get_binary(self.value, self.pos)
+    }
+
     pub fn to_json(&self) -> Result<String> {
         let mut out = String::new();
         write_json(self.value, self.metadata, self.pos, &mut out)?;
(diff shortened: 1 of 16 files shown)


crates/paimon/src/variant.rs
```diff
@@ -1694,14 +1913,1179 @@ fn java_string_cmp(left: &str, right: &str) -> std::cmp::Ordering {
...
+            scale: i8::try_from(decimal.scale() as i32).map_err(|_| Error::DataTypeInvalid {
+                message: "Variant shredding decimal scale exceeds i8".to_string(),
+            })?,
+        }),
+        DataType::VarChar(_) | DataType::Char(_) => Some(VariantScalarSchema::String),
+        DataType::VarBinary(_) | DataType::Binary(_) | DataType::Blob(_) => {
+            Some(VariantScalarSchema::Binary)
+        }
+        DataType::Date(_) => Some(VariantScalarSchema::Date32),
+        DataType::Timestamp(_) | DataType::LocalZonedTimestamp(_) => {
```
leaves12138: Collapsing `Timestamp` and `LocalZonedTimestamp` into one scalar schema loses the Variant timestamp kind. `try_typed_shred` accepts both `VariantKind::Timestamp` and `VariantKind::TimestampNtz` for this schema, but `append_scalar_shredded_value` always rebuilds with `append_timestamp`, so a shredded `TIMESTAMP_NTZ` value is read back as `TIMESTAMP`. I reproduced this against the latest head with a local unit test constructing a `TIMESTAMP_NTZ` Variant: expected `TimestampNtz`, got `Timestamp`. Can we split the scalar schema (for example `Timestamp` vs `TimestampNtz`) and call `append_timestamp_ntz` for NTZ values?

leaves12138 (changes requested):
Thanks for the large Variant shredding work. I found one correctness issue that should be fixed before merge: shredding a TIMESTAMP_NTZ Variant value can read it back as TIMESTAMP.

crates/paimon/src/variant.rs
JingsongLi: Fixed. I split the scalar schema into `Timestamp` and `TimestampNtz`: `DataType::LocalZonedTimestamp` now maps to the Variant TIMESTAMP kind, while `DataType::Timestamp` maps to TIMESTAMP_NTZ. Typed shredding now only accepts the matching Variant kind, and rebuild calls `append_timestamp_ntz` for NTZ values. Added `cast_and_rebuild_shredded_timestamp_ntz_preserves_kind` to cover the reported case.

leaves12138 (approved):
Re-reviewed the latest head. The TIMESTAMP_NTZ shredding roundtrip issue is fixed by splitting the timestamp scalar schema, and the new targeted test passes. I also ran `cargo fmt --check`, `cargo test -p paimon cast_and_rebuild_shredded_timestamp_ntz_preserves_kind --lib`, `cargo test -p paimon variant --lib`, and `cargo test -p paimon shredd --lib`.

PR #465, opened 2026-07-07
Add video_frame SQL UDF

JingsongLi:
## Summary

Adds a `video_frame(video, frame_index)` SQL UDF to the Python DataFusion binding so callers can extract PNG frames by zero-based frame index while keeping the existing `video_snapshot` timestamp-based behavior.

## Changes

- Register video-related SQL UDFs together, including the new `video_frame` built-in.
- Reuse shared frame-to-image encoding for timestamp-based and frame-index-based extraction.
- Return NULL for NULL inputs, negative frame indices, out-of-range frame indices, still images, non-video inputs, and decode failures.
- Extend Python DataFusion tests for built-in registration and frame-index extraction.

## Testing

- [x] `cargo fmt --check`
- [x] `cargo check -p pypaimon_rust`
- [x] `uv run --no-sync pytest tests/test_datafusion.py -k 'video_snapshot or video_frame or video_builtins'`

## Notes

`video_frame` uses zero-based frame indices: `video_frame(video, 0)` returns the first decoded video frame.

### DIFF
--- a/bindings/python/python/pypaimon_rust/functions.py
+++ b/bindings/python/python/pypaimon_rust/functions.py
@@ -117,15 +117,44 @@ def _decode_video_snapshot(
             break
 
         if candidate is not None:
-            try:
-                image = candidate.to_image()
-            except ImportError as e:
-                raise ImportError(
-                    "Pillow is required to encode video_snapshot images"
-                ) from e
-            output = io.BytesIO()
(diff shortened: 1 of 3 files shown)


leaves12138 (approved):
LGTM. Reviewed the implementation and tests, and verified locally with cargo check and targeted Python UDF tests.

PR #466, opened 2026-07-07
fix(datafusion): keep dynamic options out of show create

shyjsarah:
## Summary
- Generate Paimon table definitions from persisted table metadata for `SHOW CREATE TABLE`
- Capture the table definition before session dynamic options are merged into the table used for planning/scans
- Add regression coverage for `paimon.blob-as-descriptor=true` so it does not appear in `SHOW CREATE TABLE` output

## Validation
- cargo test -p paimon-datafusion --test sql_context_tests test_show_create_table_excludes_session_dynamic_options
- cargo test -p paimon-datafusion --test sql_context_tests
- cargo test -p paimon-datafusion (fails in existing fixture-dependent table scan tests because `/tmp/paimon-warehouse` lacks `default.partitioned_log_table` / `default.multi_partitioned_log_table`)

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -382,6 +382,7 @@ impl SchemaProvider for PaimonSchemaProvider {
         await_with_runtime(async move {
             match catalog.get_table(&identifier).await {
                 Ok(table) => {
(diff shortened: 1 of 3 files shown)


shyjsarah:
Closing this PR because it was based on an outdated fork branch before #444. Reopening with a minimal patch based on current apache/paimon-rust main.

PR #468, opened 2026-07-07
fix(datafusion): keep dynamic options out of show create

shyjsarah:
## Summary
- Follow up on #444 by building `SHOW CREATE TABLE` definitions from persisted table metadata before session dynamic options are merged for planning/scans
- Keep the existing blob reader registry helper path and only add an optional persisted definition override for the catalog provider path
- Add regression coverage for multiple session dynamic options (`scan.version` and `blob-as-descriptor`) while ensuring persisted table options remain in the output

## Validation
- cargo test -p paimon-datafusion --test sql_context_tests test_show_create_table_excludes_session_dynamic_options
- cargo test -p paimon-datafusion --test sql_context_tests
- cargo test -p paimon-datafusion --test blob_tests


JingsongLi (approved):
+1

PR #470, opened 2026-07-07
perf(scan): stop split construction after limit coverage

### LINKED ISSUE #469: Optimize LIMIT scan planning by stopping split construction early
### Search before asking

### Description

`ReadBuilder::with_limit()` can pass a row-count hint into paimon-core scan planning, but scan planning can still build every candidate split before the final limited plan is returned.

For small LIMIT reads on tables with many candidate split groups, especially data-evolution tables, this means the final plan may contain only a few splits while planning still pays the cost of constructing the rest. The optimization should allow scan planning to stop split construction once known split-level row counts already cover the requested limit.

The limit hint must remain conservative. It should only be used when split-level `merged_row_count()` is safe as a planning hint. Cases where rows may be reduced after planning, such as data predicates or effective row ranges, should continue to build the complete candidate split set.

Expected direction:

- Keep manifest planning complete for now.
- Push safe LIMIT planning into split construction.
- Stop building candidate splits once known `merged_row_count()` values cover the requested limit.
- Preserve fail-open behavior when known row counts do not cover the limit.
- Preserve safety gates for predicates and row ranges.
- Treat `LIMIT 0` as an empty plan.
- Expose scan trace state so tests and diagnostics can distinguish construction-time early stop from final split pruning.

This is related to, but separate from, #220. That issue covered the DataFusion fetch/limit contract; this issue is about paimon-core scan planning reducing split construction work after the limit hint has already reached core.

### Willingness to contribute

hhhizzz:
### Purpose

Linked issue: close #469

Push safe LIMIT planning into split construction so limited scans can stop building candidate splits once known split-level row counts cover the requested limit.

This keeps manifest planning complete and preserves the existing safety gates: LIMIT early-stop is only applied when split-level `merged_row_count()` remains conservative, so scans with data predicates or effective row ranges still build the full candidate set.

### Brief change log

- Add an internal LIMIT pushdown accumulator that runs while scan planning constructs `DataSplit`s.
- Stop split construction once known `merged_row_count()` values cover the requested LIMIT.
- Preserve fallback behavior when the LIMIT is not covered by known row counts, so the original candidate split set is returned instead of being unsafely pruned.
- Return an empty plan for `LIMIT 0` without building split candidates.
- Extend scan trace output with `split_candidates_built` and `limit_early_stopped` so tests and diagnostics can distinguish final split pruning from construction-time early stop.
- Add/extend data-evolution integration coverage for successful early stop and row-range safety.

### Tests

Targeted tests run:

- `rtk cargo test -p paimon test_incremental_limit_accumulator -- --nocapture`
  - `2 passed`
- `rtk cargo test -p paimon-integration-tests test_limit_pushdown -- --nocapture`
  - `3 passed`

Full `cargo test` was not run locally.

### API and Format

No storage format changes.

The public scan trace surface is extended with additive fields for LIMIT planning diagnostics:

- `split_candidates_built`
- `limit_early_stopped`

Existing read results and LIMIT semantics are unchanged; callers still need to enforce the final LIMIT at the query/read layer.

### Documentation

No user-facing documentation update is required. This is an internal scan-planning optimization with trace diagnostics covered by tests.

### DIFF
--- a/crates/integration_tests/tests/read_tables.rs
+++ b/crates/integration_tests/tests/read_tables.rs
@@ -21,13 +21,15 @@ use arrow_array::{
     Array, ArrowPrimitiveType, Int32Array, Int64Array, ListArray, MapArray, PrimitiveArray,
     RecordBatch, StringArray, StructArray,
 };
+use arrow_schema::{DataType as ArrowDataType, Field as ArrowField, Schema as ArrowSchema};
 use futures::TryStreamExt;
 use paimon::api::ConfigResponse;
 use paimon::catalog::{Identifier, RESTCatalog};
 use paimon::common::Options;
 use paimon::spec::{DataType, IntType, Predicate, Schema, VarCharType};
-use paimon::{Catalog, CatalogOptions, Error, FileSystemCatalog, Plan};
+use paimon::{Catalog, CatalogOptions, Error, FileSystemCatalog, Plan, ScanTrace};
 use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, OnceLock};
 
 #[path = "../../paimon/tests/mock_server.rs"]
 mod mock_server;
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #472, opened 2026-07-07
feat(index): add btree global index drop

QuakeWang:
### Purpose

BTree global index already supports explicit creation, but there was no matching drop API. Once a BTree index exists, users could not remove its index manifest entries through the table/DataFusion API, which also made safe rebuild flows awkward because create rejects overlapping existing global index ranges.

### Brief change log

- Add a BTree global index drop builder on `Table`.
- Drop matching current-snapshot index manifest entries via `CommitMessage.deleted_index_files`.
- Keep the operation idempotent: no snapshot, no index manifest, or no matching BTree index returns success with zero dropped entries.
- Add `CALL sys.drop_global_index(table => '...', index_column => '...', index_type => 'btree')`.
- Reject non-BTree index types and unsupported `partitions` / `dry_run` arguments explicitly.

### Tests

- `cargo test -p paimon table::btree_global_index_drop_builder --lib`
- `cargo test -p paimon-datafusion --test procedures drop_global_index`
- `cargo test -p paimon-datafusion --test procedures create_global_index`
- `cargo test -p paimon table::btree_global_index_build_builder --lib`
- `cargo check -p paimon`

### API and Format

Adds a Rust table API and DataFusion procedure for BTree global index drop.

### Documentation

No docs update in this PR; this only exposes the narrow BTree drop lifecycle already represented by index manifest delete commits.

### DIFF
--- a/crates/integrations/datafusion/src/procedures.rs
+++ b/crates/integrations/datafusion/src/procedures.rs
@@ -24,6 +24,7 @@
 //! - `CALL sys.rollback_to_timestamp(table => '...', timestamp => ...)`
 //! - `CALL sys.create_tag_from_timestamp(table => '...', tag => '...', timestamp => ...)`
 //! - `CALL sys.create_global_index(table => '...', index_column => '...', index_type => 'btree')`
+//! - `CALL sys.drop_global_index(table => '...', index_column => '...', index_type => 'btree')`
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #477, opened 2026-07-07
feat: write value stats for append data files

QuakeWang:
### Purpose

Rust append writes currently emit empty `_VALUE_STATS`, so manifest-level stats pruning cannot skip data files written by Rust even when min/max/null-count information is available.

### Brief change log

  - Accumulate min/max/null-count value stats across batches in `DataFileWriter`.
  - Persist stats in existing `DataFileMeta.value_stats` and dense `value_stats_cols`.
  - Skip unsupported variable-length value stats and add nanosecond timestamp extraction support.
  - Add tests for value stats writing, timestamp(9), skipped variable-length columns, and scan pruning.

### Tests

### DIFF
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -16,10 +16,14 @@
 // under the License.
 
 use super::shredding::PhysicalFormatWriterFactory;
-use super::{FilePredicates, FormatFileReader, FormatFileWriter};
+use super::{FilePredicates, FormatFileReader, FormatFileWriter, FormatWriteResult};
 use crate::arrow::filtering::{predicates_may_match_with_schema, StatsAccessor};
 use crate::io::{FileRead, OutputFile};
-use crate::spec::{DataField, DataType, Datum, Predicate, PredicateOperator};
+use crate::spec::stats::BinaryTableStats;
+use crate::spec::{
+    BinaryRowBuilder, CoreOptions, DataField, DataType, Datum, MetadataStatsMode, Predicate,
+    PredicateOperator,
+};
 use crate::table::{ArrowRecordBatchStream, RowRange};
 use crate::Error;
 use arrow_array::{BooleanArray, RecordBatch};
@@ -39,6 +43,7 @@ use parquet::file::metadata::{
 use parquet::file::page_index::column_index::ColumnIndexMetaData;
 use parquet::file::properties::WriterProperties;
 use parquet::file::statistics::Statistics as ParquetStatistics;
+use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::ops::Range;
 use std::sync::Arc;
@@ -49,20 +54,30 @@ pub(crate) struct ParquetFormatReader;
 /// Streams data directly to storage via `AsyncArrowWriter` + opendal.
 pub(crate) struct ParquetFormatWriter {
     inner: AsyncArrowWriter<Box<dyn crate::io::AsyncFileWrite>>,
+    write_fields: Option<Vec<DataField>>,
+    stats_modes: Option<Vec<MetadataStatsMode>>,
+    stats_dense_store: bool,
 }
 
 pub(crate) struct ParquetPhysicalWriterFactory {
     output: OutputFile,
     compression: String,
     zstd_level: i32,
+    format_options: HashMap<String, String>,
 }
 
 impl ParquetPhysicalWriterFactory {
-    pub(crate) fn new(output: &OutputFile, compression: &str, zstd_level: i32) -> Self {
+    pub(crate) fn new(
+        output: &OutputFile,
+        compression: &str,
+        zstd_level: i32,
+        format_options: HashMap<String, String>,
+    ) -> Self {
         Self {
             output: output.clone(),
             compression: compression.to_string(),
             zstd_level,
+            format_options,
         }
     }
 }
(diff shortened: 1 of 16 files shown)


JingsongLi (changes requested):
In Java Side, the performance of the accumulator is poor, and the only truly useful link is the one that extracts stats from Parquet. Perhaps we should optimize its implementation instead of the accumulator.

QuakeWang:
> In Java Side, the performance of the accumulator is poor, and the only truly useful link is the one that extracts stats from Parquet. Perhaps we should optimize its implementation instead of the accumulator.

@JingsongLi Thanks, addressed.

I changed the implementation to avoid per-row stats accumulation. Append value stats are now extracted from Parquet writer metadata on close, aligned with the Java ParquetSimpleStatsExtractor path.

Added/updated tests for append value stats, stats pruning, and timestamp stats.

crates/paimon/src/table/data_file_writer.rs
```diff
@@ -247,23 +252,19 @@ impl DataFileWriter {
         file_source: Option<i32>,
         first_row_id: Option<i64>,
         write_cols: Option<Vec<String>>,
+        write_fields: &[DataField],
+        parquet_metadata: Option<&parquet::file::metadata::ParquetMetaData>,
     ) -> DataFileMeta {
+        let (value_stats, value_stats_cols) =
+            crate::arrow::format::parquet::extract_value_stats(parquet_metadata, write_fields);
```
JingsongLi: Just return stats from parquet? It is not good to expose `ParquetMetaData` in core.

crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -625,6 +631,168 @@ fn build_row_group_column_indices(
         .collect()
 }
 
+pub(crate) fn extract_value_stats(
```
JingsongLi: We need to support default stats mode: truncate(16), otherwise the stats are too big.

QuakeWang:
@JingsongLi Addressed, thanks. Conflicts are resolved, ParquetMetaData is no longer exposed to core, and string value stats now use default truncate(16) with test coverage.

JingsongLi:
Thanks for the update. I think there are still a couple of Java compatibility gaps here.

1. The manifest stats mode is currently hard-coded to the default truncate(16) behavior. In Java, stats collectors are created from `metadata.stats-mode`, `fields.<name>.stats-mode`, and `metadata.stats-keep-first-n-columns`. With this implementation, a table configured with `metadata.stats-mode = none/counts/full` or a field-level override would still get truncate(16)-style stats in Rust-written manifests. Could we thread the table stats options into this extraction path and apply the same collector semantics as Java?

2. `supports_manifest_value_stats` currently skips unsupported min/max types completely. Java does not do that: for unsupported Parquet min/max types such as BINARY/VARBINARY, `ParquetSimpleStatsExtractor` still returns `SimpleColStats(null, null, nullCount)`, and dense stats storage keeps the column when the null count is known. The current Rust behavior drops those columns from `value_stats_cols`, so their null counts are lost. Also, Java supports TIME_WITHOUT_TIME_ZONE through IntStatistics, while Rust's conversion code has Time support but the support list excludes `DataType::Time`. Could we keep null counts for unsupported min/max columns and include Time to match Java?

crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -87,11 +108,23 @@ impl ParquetFormatWriter {
...
+        write_fields: Option<&[DataField]>,
+        format_options: &HashMap<String, String>,
     ) -> crate::Result<Self> {
         let async_write = output.async_writer().await?;
         let codec = parse_compression(compression, zstd_level);
         let inner = create_parquet_arrow_writer(async_write, schema, codec)?;
-        Ok(Self { inner })
+        let stats_modes = write_fields
+            .map(|fields| {
+                CoreOptions::new(format_options)
```
leaves12138: Parsing the per-column stats modes is not sufficient here because Java also passes `metadata.stats-dense-store` (default `true`) into `SimpleStatsConverter`. Rust currently ignores that option and `extract_value_stats` always returns dense, filtered stats with `Some(column_names)`. With `metadata.stats-dense-store=false`, Java requires full-width stats and `valueStatsCols = null`; with dense storage enabled, Java also uses `null` when all write columns are present. Please thread this flag through and add compatibility tests. A local one-column `counts` regression with dense store disabled currently returns `Some(["id"])` instead of `None`.

leaves12138 (changes requested):
The existing stats-mode, null-count, timestamp, pruning, and MERGE-focused tests pass locally. I also added a minimal compatibility regression with `metadata.stats-dense-store=false`; it fails because the produced file has `value_stats_cols = Some(["id"])` instead of `None`.

Java threads `metadata.stats-dense-store` into `SimpleStatsConverter`. In non-dense mode it stores full-width stats and leaves `valueStatsCols` null. In dense mode it filters disabled columns, but still leaves `valueStatsCols` null when every write column is represented. This implementation never reads the option and always emits `Some(column_names)`, so Rust-written manifests are not compatible with the configured Java storage semantics. Please thread the dense-store flag into extraction/serialization and add coverage for `false` plus the all-columns-present dense case.

JingsongLi:
The dedicated bloom/vector table will incorrectly crop data. A normal Parquet file only writes some fields, but 'writable' equals' None '; When the statistics cover all 'normal fields',' value_dats_coms' is also 'None'. The reading end therefore performs position mapping according to the complete table schema. Tested the schema '[payload (blob), a, b]' and data 'a=100, b=0', filtered the statistics that 'a>50' would cause misreading of 'b', and cut the file to 'finalization files=0'. The Java implementation will write 'normalColumnNames' to' writable Cols'.

JingsongLi:
`normalize_insert_batch_to_table_schema` prioritizes name and position for each column, and will reuse the source column. The target '[a, b]' and source '[b, x]' are converted to '[b, b]' instead of obtaining '[b, x]' based on the positional semantics of 'Insert ROW/*'. Temporary regression testing failed to stabilize.

JingsongLi (approved):
+1

leaves12138 (approved):
LGTM.

PR #482, opened 2026-07-08
feat(datafusion): support reading branch tables

shyjsarah:
### Purpose

Linked issue: N/A

Add read-only DataFusion support for Java-compatible Paimon branch references such as `table$branch_b1`, `table$branch_b1$snapshots`, and system-table time travel such as `table$files VERSION AS OF 'tag1'`.

### Brief change log

- Parse Paimon object names into base table, optional branch, and optional system table components.
- Validate parsed branch names before they are used in metadata paths, including direct `Table::copy_with_branch` calls.
- Add branch-aware table copies and make scans, time travel, and system tables use branch-aware snapshot/tag managers.
- Route DataFusion table and system-table lookup through branch-aware loading.
- Preserve system-table semantics when rewriting `VERSION AS OF` / `TIMESTAMP AS OF` queries, including point-in-time manifests, partitions, and table-index metadata.
- Reject writes against branch references in DataFusion and core write/commit paths, including `$branch_main` and index build/drop builders.
- Document branch reference reads and read-only behavior.

### Tests

- `cargo fmt --all --manifest-path /Users/shaoyijie/codex/code/paimon-rust/Cargo.toml`
- `cargo test --manifest-path /Users/shaoyijie/codex/code/paimon-rust/Cargo.toml -p paimon test_branch_reference_rejects_write_and_index_builders`
- `cargo test --manifest-path /Users/shaoyijie/codex/code/paimon-rust/Cargo.toml -p paimon-datafusion test_select_branch_table_reads_branch_snapshot --test sql_context_tests`
- `cargo test --manifest-path /Users/shaoyijie/codex/code/paimon-rust/Cargo.toml -p paimon-datafusion test_branch_partitions_system_table_reads_branch_snapshot --test sql_context_tests`
- `cargo check --manifest-path /Users/shaoyijie/codex/code/paimon-rust/Cargo.toml -p paimon-datafusion --lib`
- `git diff --check`

Earlier full verification on this PR also included:

- `cargo test -p paimon catalog::tests --lib`
- `cargo test -p paimon-datafusion system_tables::tests::branch_reference_does_not_drop_branch_name --lib`
- `cargo test -p paimon-datafusion --test read_tables --no-run`
- `cargo test -p paimon --lib --no-run`
- `cargo test -p paimon-integration-tests --no-run`
- `cargo test -p paimon-rest-server --test e2e --no-run`

### API and Format

This changes the public Rust API by adding branch helpers on `Table`, object-name parsing helpers under `paimon::catalog`, and `WriteBuilder::try_new_commit` for fallible commit creation. Existing `WriteBuilder::new_commit` and `TableCommit::new` remain source-compatible.

Branch write validation is enforced in core commit paths so public mutating builders and direct `TableCommit::new` users cannot bypass the read-only branch-reference contract.

This does not change the storage format.

### Documentation

Updated `docs/src/sql.md` to document branch reference reads and the read-only behavior for DataFusion branch references.


crates/paimon/src/table/mod.rs
```diff
@@ -302,6 +342,44 @@ impl Table {
...
+        let mut options = schema.options().clone();
+        options.insert("branch".to_string(), branch.clone());
+        Ok(Self {
+            file_io: self.file_io.clone(),
+            identifier: self.identifier.clone(),
+            location: self.location.clone(),
+            schema: schema.copy_with_replaced_options(options),
+            schema_manager,
+            branch,
+            branch_reference: true,
```
JingsongLi, line 401: This marks branch copies as read-only for `WriteBuilder`, but it does not cover other public mutating builders that bypass `WriteBuilder` and call `TableCommit::new` directly, such as the global-index build/drop builders. A user can still call a branch copy's index builder and commit root metadata. Please centralize the branch-reference write guard in a fallible commit path, or explicitly reject branch references in every public mutating builder before exposing branch copies.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1425,6 +1482,24 @@ impl SQLContext {
...
+    fn ensure_main_branch_write_target(name: &ObjectName, operation: &str) -> DFResult<()> {
+        let object = name
+            .0
+            .last()
+            .and_then(|part| part.as_ident())
+            .map(|ident| ident.value.as_str())
+            .ok_or_else(|| DataFusionError::Plan(format!("Invalid table reference: {name}")))?;
+        let parsed = parse_object_name(object).map_err(to_datafusion_error)?;
+        if let Some(branch) = parsed.branch() {
+            if branch != DEFAULT_MAIN_BRANCH {
```
JingsongLi: `$branch_main` is still a branch reference, but this guard allows it through. For intercepted writes (`UPDATE`, `DELETE`, `MERGE`, `TRUNCATE`, `ALTER`, `INSERT OVERWRITE`), this can fall through to table-not-found or `IF EXISTS` no-op behavior instead of the documented read-only rejection. Please reject whenever `parsed.branch().is_some()` and add `$branch_main` coverage for these write handlers.

crates/paimon/src/table/mod.rs
shyjsarah: Thanks for catching this. Fixed in `847ad3a`.

I centralized the branch-reference write guard in core with `Table::ensure_not_branch_reference_for_write()` and now enforce it from:
- `WriteBuilder` commit/write/update/delete paths
- `TableCommit` mutating commit paths
- global index build/drop builders that can bypass `WriteBuilder`

I also added coverage for a `$branch_main`-style branch reference to verify both write-builder commit creation and index build execution fail with the read-only branch error.

crates/integrations/datafusion/src/sql_context.rs
shyjsarah: Fixed in `e49afcb`.

`ensure_main_branch_write_target` now rejects any parsed branch reference, including `$branch_main`, instead of treating `main` as writable. I also added `$branch_main` coverage for the intercepted write handlers: `UPDATE`, `DELETE`, `MERGE INTO`, `TRUNCATE TABLE`, `ALTER TABLE`, and `INSERT OVERWRITE`.

Plain `INSERT INTO ...$branch_main` remains covered through the table provider write rejection.

JingsongLi:
I found one correctness issue with branch-qualified system tables.

`system_tables::load` switches the loaded `Table` to the requested branch for branch-qualified system tables, but `$partitions` is special-cased. `provider_for_table` routes it to `partitions::build(catalog, identifier, table)`, where `identifier` is still the base table identifier. Later `PartitionsTable::scan` calls `catalog.list_partitions(&identifier)`, so the branch qualifier is lost.

As a result, `SELECT * FROM t$branch_b1$partitions` returns partitions from the main branch instead of branch `b1`.

I verified this locally with a partitioned table:
1. write partition `id=1`
2. create branch `b1`
3. write partition `id=2` on main
4. query `t$branch_b1$partitions`

The result contains both main partitions (`[1, 1]`) instead of only the branch partition (`[1]`).

Could you fix `$partitions` to respect the branch reference as well? Either use the branch-bound `Table` to list/scan partition entries, or pass a branch-qualified identifier to `catalog.list_partitions`. Please also add a regression test for `table$branch_b1$partitions`.

shyjsarah:
Thanks for the detailed repro. Fixed in `eadcc9b`.

`$partitions` now preserves branch semantics by using the branch-bound `Table` for branch references instead of calling `catalog.list_partitions` with the base identifier. I also fixed the filesystem partition fallback to resolve the snapshot via the branch-aware snapshot manager while reading manifest files from the table root, which matches how branch snapshots reference existing manifest lists.

Added a regression test for `table$branch_b1$partitions`: main returns `id=1,id=2`, while the branch-qualified partitions table returns only `id=1`.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -649,6 +669,37 @@ impl SQLContext {
         }
     }
 
+    async fn load_table_for_read(
```
leaves12138: Could this branch-aware loader be extracted and reused by all read entry points? The built-in `full_text_search`, `vector_search`, `hybrid_search`, and `blob_view` functions bypass this helper and call `Catalog::get_table` with the full `table$branch_b1` identifier, which fails for `FileSystemCatalog`. A local regression query against `full_text_search('paimon.default.branch_orders$branch_b1', ...)` currently fails with `TableNotExist`.

leaves12138 (changes requested):
Thanks for the comprehensive branch work. The normal table scan, system-table, and write-guard paths look coherent, and the focused branch tests pass locally.

There is one remaining correctness gap in the built-in read TVFs. `full_text_search`, `vector_search`, `hybrid_search`, and `blob_view` still parse the full branch-qualified object name and call `Catalog::get_table` directly. With `FileSystemCatalog`, a query such as `full_text_search('paimon.default.branch_orders$branch_b1', ...)` therefore tries to load a literal `branch_orders$branch_b1` directory and fails with `TableNotExist`; I reproduced this locally.

After fixing table loading, the search execution path also needs to stay branch-aware: `FullTextSearchBuilder` and `BatchVectorSearchBuilder` currently construct a root `SnapshotManager`, so they would read the main branch's index manifest rather than the branch snapshot. Please reuse a shared branch-aware loader for all read TVFs and make the search builders resolve snapshots through the branch-bound `Table`, with regression coverage for at least full-text and vector search.

crates/integrations/datafusion/src/sql_context.rs
shyjsarah: Fixed in `fc2f5f9`. I extracted the branch-aware loading logic into a shared helper and reused it in `full_text_search`, `vector_search`, `hybrid_search`, and `blob_view`.

I also updated the full-text and vector search builders to read snapshots through `Table::snapshot_manager()`, while manifest paths still resolve from the table root.

Regression tests were added for all four read entry points. The full-text and vector tests remove the main snapshot directory to verify that branch reads do not fall back to the main branch. All CI checks are passing.

leaves12138 (approved):
Re-reviewed fc2f5f9. The shared branch-aware loader now covers full_text_search, vector_search, hybrid_search, and blob_view, and the search builders resolve snapshots through the branch-bound Table while keeping manifests at the table root. The focused branch regressions pass locally, and all CI checks are green. Looks good to me.

PR #484, opened 2026-07-08
Support dedicated vector storage writes

JingsongLi:
## Summary

Add writer support for dedicated vector storage so tables configured with `vector.file.format` write vector columns into `*.vector.<format>` files instead of ordinary data files. This aligns append and data-evolution partial writes with the existing reader support for `.vector.*` files.

## Changes

- Add `vector.file.format` and `vector.target-file-size` option accessors.
- Validate dedicated vector storage tables require data evolution, row tracking, and a normal anchor column.
- Split append writes so VECTOR columns are written to dedicated vector files while scalar/blob columns keep their existing layout.
- Split data-evolution partial writes by normal versus VECTOR columns so updates also produce dedicated vector files.
- Add focused coverage for `.vector.parquet`, `.vector.vortex`, option defaults, schema validation, and blob regression.

## Testing

- [x] `cargo test -p paimon --lib vector`
- [x] `cargo test -p paimon --lib blob_write`
- [x] `cargo test -p paimon --features vortex --lib test_vector_write_uses_dedicated_vortex_file`
- [x] `git diff --check`

## Notes

No migration is required. Existing inline VECTOR writes are unchanged when `vector.file.format` is not configured.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -47,6 +47,8 @@ const COMMIT_MAX_RETRY_WAIT_OPTION: &str = "commit.max-retry-wait";
 const FILE_COMPRESSION_OPTION: &str = "file.compression";
 const FILE_COMPRESSION_ZSTD_LEVEL_OPTION: &str = "file.compression.zstd-level";
 const FILE_FORMAT_OPTION: &str = "file.format";
+const VECTOR_FILE_FORMAT_OPTION: &str = "vector.file.format";
+const VECTOR_TARGET_FILE_SIZE_OPTION: &str = "vector.target-file-size";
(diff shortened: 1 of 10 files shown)


leaves12138 (approved):
Looks good to me. I reviewed the updated dedicated vector storage writer changes, fixed the Clippy large-enum-variant issue by boxing the dedicated writer variant, and verified the relevant paths locally.

Validated locally:
- `cargo fmt --check`
- `cargo clippy -p paimon --all-targets --features vortex -- -D warnings`
- `cargo test -p paimon --lib vector`
- `cargo test -p paimon --features vortex --lib test_vector_write_uses_dedicated_vortex_file`

The full workspace Clippy check could not be reproduced locally because this environment lacks a dynamically embeddable Python for PyO3, but the CI blocker was isolated to the `paimon` crate and is fixed in the latest commit.

PR #485, opened 2026-07-08
fix(core): align BinaryRow timestamp write path with euclidean representation

liwuhen:
### Purpose

Follow-up to a review comment from @JingsongLi on #430.

DataFusion temporal literal pushdown converts microsecond/nanosecond timestamp
literals with euclidean division (`div_euclid` / `rem_euclid`), but the
BinaryRow write/extract path used truncating division. For pre-epoch,
sub-millisecond timestamps the two paths disagree:

- `-1us` (`1969-12-31 23:59:59.999999`) is stored as `(millis=0, nanos=-1000)`
- the pushed literal becomes `(millis=-1, nanos=999_000)`

Because `Datum::Timestamp` compares lexicographically on `(millis, nanos)`, a
pushed equality/range predicate on such a value is false-negative and can drop
matching rows during stats-based file pruning. This aligns the write path with
the pushdown representation and with Paimon Java's canonical
`(millisecond, nanoOfMillisecond ∈ [0, 999999])` form.

### Brief change log

- Switch the three microsecond→`(millis, nanos)` conversions in
  `crates/paimon/src/spec/binary_row.rs` from truncating `/ 1000`, `% 1000` to
  `div_euclid(1_000)` / `rem_euclid(1_000)`:
  - `extract_datum_from_arrow` — `Timestamp`
  - `extract_datum_from_arrow` — `LocalZonedTimestamp`
  - `write_typed_value` — `TypedColumn::TimestampUs` (batch write path)

### Tests

- Unit test `test_negative_sub_millisecond_timestamp_uses_euclidean_parts`
  (binary_row.rs): verifies the extraction and binary-row write paths both
  normalize `-1us` to `(millis=-1, nanos=999_000)`. **Fails on the pre-fix code.**
- Integration test `test_negative_temporal_filter_pushdown_via_datafusion_scan`
  (read_tables.rs): writes pre-epoch fractional timestamps in separate files and
  verifies pushed equality and range predicates return the correct rows through a
  DataFusion scan.
- Commands run locally:
  - `cargo test -p paimon --lib spec::binary_row`
  - `cargo test -p paimon-datafusion --test read_tables test_negative_temporal_filter_pushdown_via_datafusion_scan`

### API and Format

No change to the BinaryRow binary layout (still `i64` millis + `i32`
nano-of-milli). The *encoded values* for negative sub-millisecond timestamps
change to the canonical euclidean form; reads of both the old and new encodings
reconstruct the same microsecond instant, so existing data remains readable.

### Documentation

### DIFF
--- a/crates/integrations/datafusion/tests/read_tables.rs
+++ b/crates/integrations/datafusion/tests/read_tables.rs
@@ -499,6 +499,87 @@ async fn test_temporal_filter_pushdown_via_datafusion_scan() {
     );
 }
 
+/// Regression test for pre-epoch, sub-millisecond timestamps.
+///
+/// The BinaryRow write path and the DataFusion literal pushdown path must
+/// agree on the `(millis, nanos)` representation of a timestamp. Previously the
+/// write path used truncating division while pushdown used euclidean division,
+/// so `1969-12-31 23:59:59.999999` (`-1us`) was stored as `(0, -1000)` but the
+/// pushed literal became `(-1, 999_000)`. A pushed equality/range predicate on
+/// such a value was therefore false-negative and dropped matching rows.
+#[tokio::test]
+async fn test_negative_temporal_filter_pushdown_via_datafusion_scan() {
+    let (_tmp, sql_context) = common::setup_sql_context().await;
+    sql_context
+        .sql(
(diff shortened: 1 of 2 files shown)


QuakeWang (approved):
LGTM

JingsongLi:
+1

PR #486, opened 2026-07-08
feat: Implement Blob View support

JingsongLi:
## Summary

Adds Java-compatible Blob View support to Paimon Rust, including BlobViewStruct serialization, table options, DataFusion SQL functions, REST-backed read resolution, write validation, and SQL documentation. The read path also avoids unnecessary BLOB IO by filtering before resolution when predicates do not depend on resolved BLOB fields and by batching descriptor range reads per blob file.

## Changes

- Add a Java-compatible BlobViewStruct spec plus CoreOptions support for `blob-view-field` and `blob-view.resolve.enabled`.
- Register DataFusion `blob_view` / `sys.blob_view` scalar functions; Blob View columns are configured through the `blob-view-field` table option.
- Store Blob View fields inline, validate canonical BlobViewStruct payloads on writes, and preserve raw references when resolution is disabled or no REST environment is available.
- Resolve Blob View fields on reads through RESTEnv-backed upstream table lookups, reusing the REST catalog environment for cross-table resolution.
- Optimize BLOB reads by filtering before Blob View / BlobDescriptor resolution when safe, applying the same predicate to the Blob View prescan, and merging nearby descriptor ranges per URI.
- Document Blob View usage, RESTCatalog-backed resolution, and dynamic options in `docs/src/sql.md`.

## Testing

- `cargo fmt`
- `cargo check -p paimon -p paimon-datafusion`
- `cargo test -p paimon --test rest_catalog_test -- --nocapture`
- `cargo test -p paimon table::data_evolution_reader::tests::test_evolution_read -- --nocapture`
- `cargo test -p paimon table::blob_resolver::tests::test_merge_blob_read_requests_merges_nearby_ranges -- --nocapture`
- `cargo test -p paimon --test rest_catalog_test test_rest_env_get_table_reuses_catalog_environment -- --nocapture`
- `cargo test -p paimon-datafusion --test blob_tests -- --nocapture`
- `cargo test -p paimon-datafusion --test blob_tests test_blob_view -- --nocapture`
- `cargo test -p paimon-datafusion sql_context -- --nocapture`
- `cargo test -p paimon-datafusion physical_plan::scan -- --nocapture`
- `git diff --check`

## Notes

- Blob View resolution currently applies to data-evolution reads and requires a REST-backed table environment. FileSystemCatalog reads preserve serialized BlobViewStruct bytes.
- `blob-view.resolve.enabled=false` preserves serialized BlobViewStruct bytes instead of resolving upstream BLOB values.


crates/paimon/src/table/data_evolution_reader.rs
```diff
@@ -375,6 +387,121 @@ impl DataEvolutionReader {
...
+            return Ok(None);
+        };
+
+        let prescan = DataEvolutionReader::new(
+            self.file_io.clone(),
+            self.schema_manager.clone(),
+            self.table_schema_id,
+            self.table_fields.clone(),
+            view_fields.clone(),
+            Vec::new(),
```
leaves12138: Java's `DataEvolutionTableRead#createBlobViewReader` applies the current predicate to the Blob View prescan before collecting `BlobViewStruct`s. This prescan drops the predicate, so Rust preloads/resolves every Blob View reference in the split, including rows that will be filtered out later. That can change behavior (for example, a filtered-out row pointing to a missing upstream table still fails the scan) and can do unnecessary upstream reads. Could we filter the prescan to the same final row set when the predicate is safe to evaluate before Blob View resolution, and add a regression test for a filtered-out invalid Blob View reference?

leaves12138 (changes requested):
I found one Java-consistency issue in the Blob View prescan path; please see the inline comment. The current CI failure appears to be runner disk exhaustion rather than a test failure.

crates/paimon/src/table/data_evolution_reader.rs
JingsongLi: Addressed in b2cd863: Blob View prescan now applies the current predicates when they are safe to evaluate before Blob View/descriptor resolution, and I added a REST-backed regression test where a filtered-out invalid BlobViewStruct no longer fails the scan.

crates/paimon/tests/rest_catalog_test.rs
```diff
@@ -254,6 +345,146 @@ async fn test_catalog_get_table() {
...
+        .get_table(&Identifier::new("default", "upstream_table"))
+        .await
+        .expect("upstream table should load via RESTEnv");
+
+    assert_eq!(upstream.identifier().full_name(), "default.upstream_table");
+    assert!(upstream.rest_env().is_some());
+}
+
+#[tokio::test]
+async fn test_blob_view_prescan_filters_invalid_filtered_out_reference() {
```
leaves12138, line 386: The regression test fixes the semantic gap, but it currently fails on `unit (windows-latest)`: `test_blob_view_prescan_filters_invalid_filtered_out_reference` panics in opendal fs listing with `StripPrefixError(())`. This test uses `FileSystemCatalog`, whose own filesystem catalog tests are already `#[cfg(not(windows))]` because directory listing on Windows is still tracked separately. Could we either make this test `#[cfg(not(windows))]` or rewrite it to avoid the Windows-unsupported filesystem listing path? The CI failure is from this test.

leaves12138 (changes requested):
The Blob View prescan semantic issue looks fixed and the new regression passes locally on Linux, but the new test is currently failing on Windows CI; please see the inline comment.

crates/paimon/tests/rest_catalog_test.rs
JingsongLi: Addressed in 604d40a by gating this FileSystemCatalog-backed regression with `#[cfg(not(windows))]`, matching the existing filesystem catalog test policy for the opendal fs listing StripPrefixError. Verified locally with rest_catalog_test and workspace clippy.

leaves12138 (approved):
LGTM. The Blob View prescan now matches the Java behavior by filtering before preloading when safe, the regression test passes locally on Linux, Windows CI is fixed by skipping the FileSystemCatalog-only regression there, and all GitHub checks are green.

PR #490, opened 2026-07-09
Support Java-style blob field directives

JingsongLi:
## Summary

Adds Java-compatible BLOB field comment directives so SQL schemas can mark binary columns as raw BLOB, descriptor BLOB, or Blob View fields. The directive handling lives in the core schema layer, while DataFusion only passes through column comments and supports the `BYTES` SQL alias.

## Changes

- Parse `__BLOB_FIELD`, `__BLOB_DESCRIPTOR_FIELD`, and `__BLOB_VIEW_FIELD` during core schema creation and `ALTER TABLE ... ADD COLUMN`.
- Normalize configured binary BLOB fields to `BlobType`, update `blob-field` / `blob-descriptor-field` / `blob-view-field` options, and strip directive prefixes from stored comments.
- Add DataFusion SQL coverage for `BYTES` and Java-style BLOB directives.
- Document BLOB column directives and their equivalent table options in `docs/src/sql.md`.

## Testing

- [x] `cargo test -p paimon blob_comment_directive --lib`
- [x] `cargo test -p paimon blob_field_option --lib`
- [x] `cargo test -p paimon test_blob_view_options --lib`
- [x] `cargo test -p paimon-datafusion blob_comment_directive --lib`
- [x] `cargo test -p paimon-datafusion test_create_table_blob --lib`
- [x] `cargo test -p paimon-datafusion test_alter_table_add_blob --lib`
- [x] `git diff --check`

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -16,17 +16,22 @@
 // under the License.
 
 use crate::spec::core_options::{
-    first_row_supports_changelog_producer, CoreOptions, BUCKET_KEY_OPTION,
-    QUERY_AUTH_ENABLED_OPTION, SEQUENCE_FIELD_OPTION,
+    first_row_supports_changelog_producer, CoreOptions, BLOB_DESCRIPTOR_FIELD_OPTION,
+    BLOB_FIELD_OPTION, BLOB_VIEW_FIELD_OPTION, BUCKET_KEY_OPTION, QUERY_AUTH_ENABLED_OPTION,
+    SEQUENCE_FIELD_OPTION,
 };
 use crate::spec::types::{ArrayType, DataType, MapType, MultisetType, RowType};
 use crate::spec::{
-    remove_field_scoped_options, rename_field_scoped_options, AggregationConfig, ColumnMove,
-    ColumnMoveType, PartialUpdateConfig,
+    remove_field_scoped_options, rename_field_scoped_options, AggregationConfig, BlobType,
+    ColumnMove, ColumnMoveType, PartialUpdateConfig,
 };
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use std::collections::{HashMap, HashSet};
+
+const BLOB_FIELD_DIRECTIVE: &str = "__BLOB_FIELD";
+const BLOB_DESCRIPTOR_FIELD_DIRECTIVE: &str = "__BLOB_DESCRIPTOR_FIELD";
+const BLOB_VIEW_FIELD_DIRECTIVE: &str = "__BLOB_VIEW_FIELD";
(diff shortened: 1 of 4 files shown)


crates/paimon/src/spec/schema.rs
```diff
@@ -618,6 +628,66 @@ fn reassign_field_ids(data_type: DataType, next_id: &mut i32) -> DataType {
...
+
+fn parse_blob_comment_directive(comment: Option<&str>) -> Option<BlobCommentDirective> {
+    let comment = comment?;
+    let mut parts = comment.splitn(2, ';');
+    let directive = parts.next().unwrap_or_default().trim();
+    let option_key = match directive {
+        BLOB_FIELD_DIRECTIVE => BLOB_FIELD_OPTION,
+        BLOB_DESCRIPTOR_FIELD_DIRECTIVE => BLOB_DESCRIPTOR_FIELD_OPTION,
+        BLOB_VIEW_FIELD_DIRECTIVE => BLOB_VIEW_FIELD_OPTION,
+        _ => return None,
```
QuakeWang: This should not silently fall back to a normal comment when the column comment starts with `__BLOB` but does not match one of the supported directives.

Java Paimon rejects unknown blob directives in this case, e.g. `__BLOB_UNKNOWN`, so a typo like `__BLOB_FIEL; picture` should fail fast instead of creating a plain binary column. Otherwise the user gets a non-BLOB schema with no `blob-field` option and only discovers the issue later.

Could we make this parser return an error for unknown `__BLOB*` directives and add a create/add-column test for that case?

JingsongLi: Add `test_blob_comment_directive_rejects_unknown_create_directive`.

leaves12138 (commented):
One small Java-compatibility gap remains in the directive parser.

Java `ColumnDirectiveUtils.matchDirective` accepts only a marker by itself or a marker immediately followed by `;`. For example, `__BLOB_FIELD; comment` is valid, but `__BLOB_FIELD ; comment` is rejected because the character after `__BLOB_FIELD` is a space, not `;`.

The Rust parser currently does `splitn(2, ';')` and then trims the directive part, so `COMMENT '__BLOB_FIELD ; payload bytes'` is accepted, stripped, and converted into a BLOB column. I verified this locally with a temporary regression test: `Schema::builder()` returned `Ok` with `blob-field=payload`, whereas Java would throw an unsupported directive error.

Could we align the matching with Java by trimming the whole comment first and accepting only `marker` or `marker;...` (no whitespace between the marker and semicolon), and add a regression test for this malformed directive?

leaves12138 (approved):
Looks good to me. The Java-style BLOB directive parsing and validation issues from the previous review are addressed.

PR #491, opened 2026-07-09
Enable Mosaic reads by default

JingsongLi:
## Summary

Make Mosaic reads always available instead of hiding them behind a Cargo feature, and upgrade `paimon-mosaic-core` to 0.2.0 so array and map columns are supported by the reader.

## Changes

- Treat `paimon-mosaic-core` as a regular dependency and remove the Mosaic Cargo feature declarations.
- Remove Mosaic feature gates from the core reader registration, unit tests, and DataFusion fixture test.
- Delegate Mosaic Arrow type validation to `paimon-mosaic-core` and add an array/map round-trip test.
- Remove `mosaic` from CI feature lists and update docs to describe Mosaic reads as always available.

## Testing

- [x] `cargo check -p paimon`
- [x] `cargo check -p paimon --no-default-features --features storage-memory`
- [x] `cargo check -p paimon-datafusion --no-default-features`
- [x] `cargo test -p paimon mosaic --lib`
- [x] `cargo test -p paimon-datafusion --no-default-features --test mosaic_tables`

## Notes

`cargo check -p paimon --no-default-features` still requires at least one storage feature because of existing storage enum behavior, so this PR verifies the no-default path with `storage-memory` enabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Clippy
-        run: cargo clippy --all-targets --workspace --features fulltext,vortex,mosaic -- -D warnings
+        run: cargo clippy --all-targets --workspace --features fulltext,vortex -- -D warnings
 
   build:
     runs-on: ${{ matrix.os }}
@@ -71,7 +71,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - name: Build
-        run: cargo build --features fulltext,vortex,mosaic
+        run: cargo build --features fulltext,vortex
 
   unit:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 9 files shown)


leaves12138 (commented):
One compatibility concern: removing the `mosaic` feature name itself is a breaking change for downstream crates that already depend on it.

After this PR, Mosaic is always enabled, but existing dependency declarations such as:

```toml
paimon = { version = "0.3.0", features = ["mosaic"] }
paimon-datafusion = { version = "0.3.0", features = ["mosaic"] }
```

will fail Cargo resolution because neither crate defines a `mosaic` feature anymore. I verified this locally with a small downstream `cargo check`, which fails with:

```text
package `check-mosaic-feature` depends on `paimon` with feature `mosaic` but `paimon` does not have that feature
```

Could we keep a no-op compatibility feature, e.g. `mosaic = []` in `paimon` and `mosaic = ["paimon/mosaic"]` or `mosaic = []` in `paimon-datafusion`, while still making the dependency non-optional and enabling reads by default? That preserves existing user Cargo.toml files and CI commands while achieving the goal of this PR.

leaves12138 (approved):
Looks good to me. The Mosaic feature compatibility concern is intentionally not a blocker here.

PR #492, opened 2026-07-09
Support REST format table reads

JingsongLi:
## Summary

Adds read support for RESTCatalog format tables while keeping format-table implementation separate from ordinary Paimon table paths.

## Changes

- Detect `type=format-table` metadata and route read, scan, and write builders through dedicated format-table implementations.
- Load REST table `path` into schema options for format tables and scan external format-table files directly.
- Add partition predicate pruning for format table scans, including direct partition path scans for exact partition predicates and leading equality path pruning.
- Keep format-table writes fail-closed with explicit unsupported errors.
- Add RESTCatalog integration coverage for plain and partitioned format-table reads.

## Testing

- `cargo fmt --check && git diff --check`
- `cargo test -p paimon --test rest_catalog_test`
- `cargo test -p paimon --test rest_catalog_test test_rest_catalog_prunes_format_table_partition_filter`
- `cargo test -p paimon --test rest_catalog_test test_rest_catalog_reads_format_table`
- `cargo test -p paimon table_scan`
- `cargo test -p paimon read_builder`

## Notes

Data-column predicates continue to be passed to the file reader for format-level pruning/filtering. Format-table scan now uses partition predicates to avoid unnecessary directory scans where possible.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -31,6 +31,8 @@ const SOURCE_SPLIT_TARGET_SIZE_OPTION: &str = "source.split.target-size";
 const SOURCE_SPLIT_OPEN_FILE_COST_OPTION: &str = "source.split.open-file-cost";
 const PARTITION_DEFAULT_NAME_OPTION: &str = "partition.default-name";
 const PARTITION_LEGACY_NAME_OPTION: &str = "partition.legacy-name";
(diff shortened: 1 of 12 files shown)


leaves12138 (approved):
Looks good to me. The latest changes align the REST format-table read path with the Java behavior for partition column materialization, read-side limit handling, and value-only partition paths.

PR #493, opened 2026-07-09
feat(index): Support vector index rerank

JingsongLi:
### Purpose

This pull request adds rerank support for vector index search, aligned with the Java implementation. When a positive refine factor is configured, the index search first retrieves a larger candidate set and then reranks those candidates by reading raw vectors and computing exact scores.

### Brief change log

- Support vector index rerank with `refine_factor` / `refine-factor` / `rerank_factor` / `rerank-factor` option aliases.
- Resolve rerank options from search options first, then table options, with field/index-specific prefixes taking precedence over global options.
- Expand index search limit to `limit * refine_factor` and rerank candidates back to the original query limit.
- Read raw vectors for the union of candidate row IDs while scoring only the candidates relevant to each query.
- Keep Java-compatible ranking semantics: higher score first, and smaller row id first when scores tie.
- Optimize rerank performance with bounded top-K heaps, row-id-to-query routing, zero-copy Arrow vector scoring, and streaming per-query top-K collection.

### Tests

- `cargo fmt --all`
- `cargo test -p paimon vector_search --lib`
- `git diff --check`

### API and Format

This change adds vector search option aliases for rerank/refine factor. It does not change the storage format.

### Documentation

No user-facing documentation is updated in this pull request. The supported options are covered by unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 [Latest Version]: https://img.shields.io/crates/v/paimon.svg
 [crates.io]: https://crates.io/crates/paimon
 
-The rust implementation of Apache Paimon. 1
+The rust implementation of Apache Paimon. 
 
 ## Issue Tracker
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 [Latest Version]: https://img.shields.io/crates/v/paimon.svg
 [crates.io]: https://crates.io/crates/paimon
 
-The rust implementation of Apache Paimon. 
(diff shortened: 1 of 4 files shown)


leaves12138 (approved):
LGTM. The rerank implementation is aligned with the Java behavior, keeps the expected score ordering semantics, and the follow-up optimization avoids unnecessary allocations and full sorts. The accidental README change has been removed, and the check job is now passing.

PR #496, opened 2026-07-10
feat: support case-insensitive column matching on reads (read-time parameter)

### LINKED ISSUE #495: Support case-insensitive column matching on reads (read-time parameter)
### Motivation

Spark + Java Paimon support **case-insensitive column matching** on reads. In Java this lives at the **engine / catalog layer**: Spark's `spark.sql.caseSensitive` (a session setting) drives Catalyst to resolve projection and filter column references case-insensitively and normalize them to the schema's field names before pushing down to the connector; Paimon core exposes it as a per-call parameter (`RowType.getFieldIndex(name, caseSensitive)`). It is **not** a table property.

The Rust implementation (`paimon-rust`) has no equivalent — projection and predicate column resolution are always exact-case. Engines that read Paimon through the Rust core / C bindings / DataFusion / Python (e.g. Apache Doris) therefore cannot offer the case-insensitive behavior users get from Spark. A user querying `SELECT NAME` or `WHERE NAME = ...` against a schema field `name` fails, even though the same query works via Spark.

### Proposal

Expose case sensitivity as a **read-time parameter** (mirroring Java's per-call `caseSensitive`), controlled by the reader/engine rather than stored on the table. Default stays **case-sensitive**, preserving today's exact-match behavior; a reader opts in per read. When disabled, column-name resolution on the **read path** matches by ASCII case-folding and reports an error when a name is ambiguous (multiple schema fields collide under folding), mirroring Spark's `AMBIGUOUS` behavior.

Surface:
- Core: `ReadBuilder::with_case_sensitive(bool)`; `PredicateBuilder::new_with_case_sensitive(fields, bool)`.
- DataFusion: derived from the session (`!sql_parser.enable_ident_normalization`) for scan and filter pushdown.
- Python: `PyReadBuilder::with_case_sensitive(bool)`.
- C binding: `paimon_read_builder_with_case_sensitive`, plus a `case_sensitive` parameter on the `paimon_predicate_*` constructors (used by Doris and the Go binding).

Out of scope for the initial change:
- Nested/dotted column paths (top-level only).
- The `_ROW_ID` system column stays an exact reserved name.
- Write path / DDL.

### Notes

Default stays case-sensitive, so existing behavior is unchanged unless a reader explicitly requests case-insensitive matching. No table option and no storage-format change.

JunRuiLee:
### Purpose

Linked issue: close #495

Spark + Java Paimon support case-insensitive column matching on reads. In Java this lives at the **engine / catalog layer** — Spark's `spark.sql.caseSensitive` (a session setting) drives Catalyst to resolve projection/filter columns case-insensitively and normalize them to the schema names before pushdown; Paimon core exposes it as a per-call parameter (`RowType.getFieldIndex(name, caseSensitive)`). It is **not** a table property.

`paimon-rust` had no equivalent — column resolution was always exact-case — so callers reading Paimon through the Rust core / C bindings / Python couldn't offer the behavior users get from Spark.

This PR adds case-insensitive column matching as a **read-time parameter** (mirroring Java's per-call `caseSensitive`), not a table option. Default stays **case-sensitive**, so existing behavior is unchanged; a reader opts in per read. When enabled, column names resolve by **ASCII case-folding** and an **ambiguous** reference (schema fields colliding under folding) is reported, mirroring Spark's `AMBIGUOUS` behavior.

### Scope of the opt-in

Case-insensitive matching is offered on the **direct ReadBuilder API** paths: **core**, the **C binding**, and **Python**. It is **not** offered through **DataFusion / SQL**: DataFusion resolves projection/filter columns against the provider schema during logical planning, *before* `TableProvider::scan` runs, and `enable_ident_normalization` only lowercases unquoted identifiers — it does not make schema resolution case-insensitive. A genuine case mismatch (`SELECT name` against a `Name` field) therefore fails at planning and never reaches Paimon's resolution. So the DataFusion path always matches **case-sensitively**; a new end-to-end negative test pins this.

### Brief change log

- **Core**: `ReadBuilder::with_case_sensitive(bool)` (default `true`) and `PredicateBuilder::new_with_case_sensitive(fields, bool)` (the existing `new` delegates with `true`). When disabled, column resolution folds case, errors on ambiguity, and stores the **canonical** schema name in the predicate leaf so downstream name-based lookups (e.g. index pruning) keep matching. Projection duplicate detection folds case so a case-colliding projection is reported as a duplicate.
- **Order-independent projection**: projection name resolution is deferred to scan/read build time and uses the case sensitivity effective then, so `with_projection` and `with_case_sensitive` may be called in any order. `with_projection` does best-effort early validation, rejecting only columns that cannot match under any case sensitivity (an obvious typo); case-dependent outcomes surface from `new_read`.
- **DataFusion**: always resolves column names **case-sensitively** for scan and filter translation (see "Scope" above). `supports_filters_pushdown` classifies case-sensitively and still refuses to report `Exact` for a schema with ASCII-case-folding collisions, as a conservative guard so a needed residual filter is never dropped.
- **Python**: `PyReadBuilder.with_case_sensitive(bool)` threads the flag into projection and predicate conversion (type stub added).
- **C binding**: `paimon_read_builder_with_case_sensitive(rb, bool)` (new symbol), plus **additive** `paimon_predicate_*_with_case_sensitive` variants of each predicate constructor. The existing `paimon_predicate_*` symbols keep their original signatures and case-sensitive behavior, so the current C ABI and its consumers (including the in-repo Go binding, which prepares fixed libffi call interfaces per symbol) are unaffected. Compile-time signature guards pin the existing ABI. (Go's own opt-in case-insensitive API is left to a follow-up.)

ASCII-only folding throughout (`eq_ignore_ascii_case`); no Unicode case folding.

**Scope**: read path, top-level columns only. Nested/dotted paths, the `_ROW_ID` system column (kept an exact reserved name), the write path, and the DataFusion/SQL path are out of scope.

### Tests

- `cargo test -p paimon --lib` — projection/predicate case-insensitive match, default-sensitive rejection, ambiguity error, canonical-name storage, duplicate detection, order-independence (projection ↔ case-sensitive either order), and deferral of case-dependent errors to `new_read`.
- `cargo test -p paimon-datafusion` — `filter_pushdown` translation/classify guard, and an end-to-end SQL negative test asserting that `SELECT name` against a `Name` field fails at planning (documenting why the SQL path stays case-sensitive).
- Python predicate/read tests (pyo3, `PYO3_PYTHON=python3.12`).

### API and Format

- New public API: `ReadBuilder::with_case_sensitive`, `PredicateBuilder::new_with_case_sensitive`; C `paimon_read_builder_with_case_sensitive` and **additive** `paimon_predicate_*_with_case_sensitive` variants. Existing `PredicateBuilder::new` and the existing `paimon_predicate_*` C symbols are unchanged (case-sensitive), preserving ABI compatibility.
- **No table option and no storage-format change.** Default behavior is unchanged (case-sensitive).

### Documentation

Behavior described above and in the linked issue; the switch is a read-time API parameter, controlled by the reader/engine (not stored on the table), and is order-independent with projection. Case-insensitive matching is available via the direct ReadBuilder API (core / C / Python); the DataFusion/SQL path stays case-sensitive.

### DIFF
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -113,10 +113,12 @@ pub unsafe extern "C" fn paimon_read_builder_free(rb: *mut paimon_read_builder) 
 ///
 /// The `columns` parameter is a null-terminated array of null-terminated C strings.
 /// Output order follows the caller-specified order. An empty list is a valid
-/// zero-column projection. Column-name resolution is deferred to
-/// `paimon_read_builder_new_read` (order-independent with
-/// `paimon_read_builder_with_case_sensitive`), so unknown, duplicate, or
-/// ambiguous names are reported there, not by this call.
+/// zero-column projection. An obvious typo — a name that matches no field under
+/// any case sensitivity — is rejected by this call. Case-dependent resolution
+/// (a name that matches only case-insensitively, or a case-fold ambiguity) is
+/// deferred to `paimon_read_builder_new_read`, which uses the case sensitivity
+/// effective then, so this stays order-independent with
+/// `paimon_read_builder_with_case_sensitive`.
 ///
 /// # Safety
 /// `rb` must be a valid pointer from `paimon_table_new_read_builder`, or null (returns error).
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -167,10 +167,16 @@ pub unsafe extern "C" fn paimon_read_builder_with_projection(
     std::ptr::null_mut()
 }
 
-/// Set whether column-name matching (projection and predicate resolution) is
-/// case-sensitive for this ReadBuilder. Defaults to `true` (exact match). When
-/// `false`, column names are matched by ASCII case-folding and an ambiguous
+/// Set whether column-name matching for **projection** is case-sensitive for
+/// this ReadBuilder. Defaults to `true` (exact match). When `false`, projected
+/// column names are matched by ASCII case-folding and an ambiguous
 /// (case-colliding) request errors.
+///
+/// This does **not** affect predicate resolution: a predicate is resolved when
+/// it is constructed, so its case sensitivity is chosen by which constructor
+/// you call — `paimon_predicate_*` (case-sensitive) or the additive
+/// `paimon_predicate_*_with_case_sensitive` variant — independently of this
+/// setting.
 ///
 /// # Safety
 /// `rb` must be a valid pointer from `paimon_table_new_read_builder`, or null (returns error).
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -811,12 +811,28 @@ unsafe fn build_leaf_predicate(
     }
 }
 
-/// Create an equality predicate: `column = datum`.
+/// Create an equality predicate: `column = datum` (case-sensitive column match).
+///
+/// For case-insensitive column matching use
+/// `paimon_predicate_equal_with_case_sensitive`.
 ///
 /// # Safety
 /// `table` and `column` must be valid pointers.
 #[no_mangle]
 pub unsafe extern "C" fn paimon_predicate_equal(
+    table: *const paimon_table,
+    column: *const std::ffi::c_char,
+    datum: paimon_datum,
+) -> paimon_result_predicate {
+    build_leaf_predicate_datum(table, column, &datum, true, |pb, col, d| pb.equal(col, d))
+}
+
+/// Create an equality predicate with configurable column-name case sensitivity.
+///
+/// # Safety
+/// `table` and `column` must be valid pointers.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_predicate_equal_with_case_sensitive(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
@@ -827,12 +843,27 @@ pub unsafe extern "C" fn paimon_predicate_equal(
     })
 }
 
-/// Create a not-equal predicate: `column != datum`.
+/// Create a not-equal predicate: `column != datum` (case-sensitive column match).
 ///
 /// # Safety
 /// `table` and `column` must be valid pointers.
 #[no_mangle]
 pub unsafe extern "C" fn paimon_predicate_not_equal(
+    table: *const paimon_table,
+    column: *const std::ffi::c_char,
+    datum: paimon_datum,
+) -> paimon_result_predicate {
+    build_leaf_predicate_datum(table, column, &datum, true, |pb, col, d| {
+        pb.not_equal(col, d)
+    })
+}
+
+/// Create a not-equal predicate with configurable column-name case sensitivity.
+///
+/// # Safety
+/// `table` and `column` must be valid pointers.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_predicate_not_equal_with_case_sensitive(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
@@ -843,12 +874,27 @@ pub unsafe extern "C" fn paimon_predicate_not_equal(
     })
 }
 
-/// Create a less-than predicate: `column < datum`.
+/// Create a less-than predicate: `column < datum` (case-sensitive column match).
 ///
 /// # Safety
 /// `table` and `column` must be valid pointers.
 #[no_mangle]
 pub unsafe extern "C" fn paimon_predicate_less_than(
+    table: *const paimon_table,
+    column: *const std::ffi::c_char,
+    datum: paimon_datum,
+) -> paimon_result_predicate {
+    build_leaf_predicate_datum(table, column, &datum, true, |pb, col, d| {
+        pb.less_than(col, d)
+    })
+}
+
+/// Create a less-than predicate with configurable column-name case sensitivity.
+///
+/// # Safety
+/// `table` and `column` must be valid pointers.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_predicate_less_than_with_case_sensitive(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
@@ -859,12 +905,27 @@ pub unsafe extern "C" fn paimon_predicate_less_than(
     })
 }
 
-/// Create a less-or-equal predicate: `column <= datum`.
+/// Create a less-or-equal predicate: `column <= datum` (case-sensitive column match).
 ///
 /// # Safety
 /// `table` and `column` must be valid pointers.
 #[no_mangle]
 pub unsafe extern "C" fn paimon_predicate_less_or_equal(
+    table: *const paimon_table,
+    column: *const std::ffi::c_char,
+    datum: paimon_datum,
+) -> paimon_result_predicate {
+    build_leaf_predicate_datum(table, column, &datum, true, |pb, col, d| {
+        pb.less_or_equal(col, d)
+    })
+}
+
+/// Create a less-or-equal predicate with configurable column-name case sensitivity.
+///
+/// # Safety
+/// `table` and `column` must be valid pointers.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_predicate_less_or_equal_with_case_sensitive(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
@@ -875,12 +936,27 @@ pub unsafe extern "C" fn paimon_predicate_less_or_equal(
     })
 }
 
-/// Create a greater-than predicate: `column > datum`.
+/// Create a greater-than predicate: `column > datum` (case-sensitive column match).
 ///
 /// # Safety
 /// `table` and `column` must be valid pointers.
 #[no_mangle]
 pub unsafe extern "C" fn paimon_predicate_greater_than(
+    table: *const paimon_table,
+    column: *const std::ffi::c_char,
+    datum: paimon_datum,
+) -> paimon_result_predicate {
+    build_leaf_predicate_datum(table, column, &datum, true, |pb, col, d| {
+        pb.greater_than(col, d)
+    })
+}
+
+/// Create a greater-than predicate with configurable column-name case sensitivity.
+///
+/// # Safety
+/// `table` and `column` must be valid pointers.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_predicate_greater_than_with_case_sensitive(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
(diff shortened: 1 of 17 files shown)


bindings/c/src/table.rs
```diff
@@ -772,8 +820,11 @@ pub unsafe extern "C" fn paimon_predicate_equal(
     table: *const paimon_table,
     column: *const std::ffi::c_char,
     datum: paimon_datum,
+    case_sensitive: bool,
```
QuakeWang, line 847: Existing `paimon_predicate_*` symbols now require an extra argument, but the Go libffi wrappers still call them with the old signatures. This leaves `case_sensitive` undefined at the ABI boundary and breaks all predicate constructors used by the Go binding.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -349,9 +353,16 @@ impl TableProvider for PaimonTableProvider {
         filters: &[Expr],
         limit: Option<usize>,
     ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        // Derive column-name case sensitivity from the DataFusion session:
+        // when DataFusion normalizes (lowercases) unquoted identifiers (its
+        // default), Paimon should match case-insensitively; when normalization
+        // is off, match exactly.
+        let case_sensitive = !state.config_options().sql_parser.enable_ident_normalization;
```
QuakeWang: `enable_ident_normalization` only lowercases unquoted identifiers. DataFusion resolves columns against the schema before `TableProvider::scan`; with a schema field `Name`, `SELECT name` fails with `column 'name' not found`, so this code never runs and case-insensitive SQL reads do not work.

bindings/c/src/table.rs
JunRuiLee: Confirmed. Restored the existing `paimon_predicate_*` signatures (case-sensitive) and added additive `paimon_predicate_*_with_case_sensitive` variants, so the Go libffi CIFs in `bindings/go/predicate.go` keep matching with no change to the Go binding. Added compile-time signature guards (`const _: unsafe extern "C" fn(...) = paimon_predicate_*`) that fail the build if one of these symbols changes shape again. Go's own opt-in case-insensitive API is left to a follow-up. (36ad296)

crates/integrations/datafusion/src/table/mod.rs
JunRuiLee: Confirmed — DataFusion resolves projection/filter columns against the provider schema before `TableProvider::scan`, and `enable_ident_normalization` only lowercases unquoted identifiers rather than making schema resolution case-insensitive, so `SELECT name` against a `Name` field fails at planning and the scan-level folding is never reached. Dropped the derivation: the DataFusion/SQL path now always matches case-sensitively (both `scan` and the variant-extraction planner pass `case_sensitive = true`), with a comment explaining why. Added an end-to-end negative test that asserts `SELECT name` vs a `Name` field fails at planning, pinning this behavior. Case-insensitive matching stays a direct-ReadBuilder feature (core/C/Python). (36ad296)

crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -81,22 +89,34 @@ pub(crate) fn analyze_filters(filters: &[Expr], fields: &[DataField]) -> FilterP
...
+    // direct ReadBuilder API callers (and its own unit tests), but the DataFusion
+    // TableProvider/SQL path always passes `case_sensitive = true`: the planner
+    // resolves columns against the schema before `scan`, so SQL reads are
+    // case-sensitive. Reporting `Exact` tells DataFusion to drop its residual
+    // filter, so it must only be returned when column resolution is unambiguous.
+    // The `Inexact` cap for schemas with ASCII case-folding collisions (e.g.
+    // `Name`/`name`) is a conservative guard: if a reference were ever resolved
+    // case-insensitively, `Exact` would have dropped the residual while the
+    // translation pushed nothing — filtering neither side. Keep the residual.
+    let allow_exact = !has_ascii_case_collision(fields);
```
QuakeWang: DataFusion now always uses `case_sensitive = true`, but this guard still downgrades every otherwise `Exact` filter to `Inexact` whenever any two schema fields differ only by case. For example, a filter on `pt` becomes `Inexact` solely because unrelated `Name` / `name` fields exist.

bindings/c/src/table.rs
```diff
@@ -157,6 +167,26 @@ pub unsafe extern "C" fn paimon_read_builder_with_projection(
     std::ptr::null_mut()
 }
 
+/// Set whether column-name matching (projection and predicate resolution) is
```
QuakeWang: This setting does not control predicate resolution. Predicates are resolved earlier by the separate predicate constructors, so setting this to `false` alone only affects projection matching.

crates/integrations/datafusion/src/filter_pushdown.rs
JunRuiLee: Good catch. Since the DataFusion path always resolves case-sensitively, a reference matches exactly one field and an unrelated ASCII case-folding collision (e.g. `Name`/`name`) can never make a resolved filter ambiguous — the `Inexact` cap was only over-downgrading. Dropped the `has_ascii_case_collision` guard so classification depends solely on `requires_residual` / `is_exact_filter_pushdown`, and retargeted the test to assert a filter on `dt` stays `Exact` in a schema that also has a `Name`/`name` pair. (e205628)

bindings/c/src/table.rs
JunRuiLee: Correct — a predicate is resolved when it is constructed, so this setting only affects projection matching. Rather than have the read builder retroactively rewrite an already-resolved predicate (which would break the independent predicate-constructor ABI), I fixed the doc: `paimon_read_builder_with_case_sensitive` now documents that it controls projection only, and that predicate case sensitivity is chosen by the constructor variant (`paimon_predicate_*` vs `paimon_predicate_*_with_case_sensitive`), independent of this setting. (e205628)

bindings/c/src/table.rs
```diff
@@ -111,8 +112,11 @@ pub unsafe extern "C" fn paimon_read_builder_free(rb: *mut paimon_read_builder)
 /// Set column projection for a ReadBuilder.
 ///
 /// The `columns` parameter is a null-terminated array of null-terminated C strings.
-/// Output order follows the caller-specified order. Unknown or duplicate names
-/// are validated immediately; an empty list is a valid zero-column projection.
+/// Output order follows the caller-specified order. An empty list is a valid
+/// zero-column projection. Column-name resolution is deferred to
+/// `paimon_read_builder_new_read` (order-independent with
+/// `paimon_read_builder_with_case_sensitive`), so unknown, duplicate, or
```
QuakeWang: Unknown columns are not always deferred to `paimon_read_builder_new_read`. The `validate_projection_possible` call below still rejects names that cannot match under either case mode in this function.

crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -81,18 +89,29 @@ pub(crate) fn analyze_filters(filters: &[Expr], fields: &[DataField]) -> FilterP
...
     filter: &Expr,
     fields: &[DataField],
+    case_sensitive: bool,
     is_exact_filter_pushdown: F,
 ) -> TableProviderFilterPushDown
 where
     F: Fn(&Predicate) -> bool,
 {
-    let translator = FilterTranslator::new(fields);
+    // `FilterTranslator` still supports case-insensitive column resolution for
```
QuakeWang, line 104: `FilterTranslator` is private to the DataFusion integration, so direct `ReadBuilder` callers never reach this case-insensitive branch. Every non-test call site passes `true`, leaving the `false` path test-only.

crates/paimon/src/table/read_builder.rs
```diff
@@ -376,27 +459,91 @@ impl<'a> PaimonReadBuilder<'a> {
...
+pub(super) fn validate_projection_possible(
+    full_name: String,
+    fields: &[DataField],
+    projection_names: &[String],
+) -> Result<()> {
+    for name in projection_names {
+        if name == crate::spec::ROW_ID_FIELD_NAME {
+            continue;
+        }
+        let matches_any = fields.iter().any(|f| f.name().eq_ignore_ascii_case(name));
```
QuakeWang: This scans the full schema once per projected name, and the same lookup runs again in `new_scan` and `new_read`. The default case-sensitive path regresses from the previous `O(fields + projections)` lookup to `O(fields × projections)`, which is significant for wide tables.

JunRuiLee: Fixed. Both `resolve_projected_fields` and `validate_projection_possible` now build the name index once (O(fields)) and then look up each projected name in it, restoring O(fields + projections): an exact-name `HashMap` for the case-sensitive path, and a folded-name `HashMap` (folding a collision to `None`) for the case-insensitive path, which keeps the ambiguity check. Dropped the now-unused `find_projection_field`. (2e8c912)

bindings/c/src/table.rs
JunRuiLee: Fixed the doc to match. An obvious typo — a name that matches no field under any case sensitivity — is rejected by this call (via `validate_projection_possible`); only case-dependent outcomes (a name matching only case-insensitively, or a case-fold duplicate/ambiguity) are deferred to `paimon_read_builder_new_read`, which uses the case sensitivity effective then. (2e8c912)

crates/integrations/datafusion/src/filter_pushdown.rs
JunRuiLee: True that every current call site passes `true` — but the `case_sensitive` parameter is intentional, not dead code. It is the explicit carrier of the "SQL path is always case-sensitive" contract: `scan` and the variant-extraction planner both pass `true` on purpose, and the end-to-end negative test (`SELECT name` vs a `Name` field fails at planning) pins that. The `false` path keeps `FilterTranslator`/`classify` consistent with the rest of the case-insensitivity machinery and is exercised by unit tests. I would rather keep the contract visible in the signature than hardcode `true` internally and lose that. Happy to revisit if you feel strongly.

QuakeWang: Understood. Since this is intentional and the production path is explicitly pinned to `true`, I don't feel strongly about it. Resolving.

crates/integrations/datafusion/src/table/mod.rs
```diff
@@ -426,12 +439,17 @@ impl TableProvider for PaimonTableProvider {
         filters: &[&Expr],
     ) -> DFResult<Vec<TableProviderFilterPushDown>> {
         let fields = self.table.schema().fields();
+        // SQL reads resolve columns case-sensitively (see `scan`), so classify
+        // pushdown the same way. `classify_filter_pushdown` still caps at
```
QuakeWang: nit: This comment is stale after `e205628c`; `classify_filter_pushdown` no longer caps case-colliding schemas at `Inexact`.

crates/paimon/src/table/read_builder.rs
```diff
@@ -141,6 +147,31 @@ impl<'a> ReadBuilder<'a> {
         Ok(self)
     }
 
+    /// Set whether column-name matching (projection and predicate column
+    /// resolution) is case-sensitive. Defaults to `true` (exact match). When set
+    /// to `false`, names are matched by ASCII case-folding and an ambiguous
+    /// (case-colliding) request errors. Mirrors Java's
```
QuakeWang: nit: Current Java Paimon `RowType` only exposes `getFieldIndex(String fieldName)`; there is no `getFieldIndex(name, caseSensitive)` overload, so this reference is inaccurate.

QuakeWang (approved):
LGTM. Left a couple of non-blocking nits.

crates/integrations/datafusion/src/table/mod.rs
JunRuiLee: Fixed — dropped the stale sentence; the comment now just says SQL reads classify case-sensitively. (30e24ee)

crates/paimon/src/table/read_builder.rs
JunRuiLee: Good catch — Java `RowType` only has `getFieldIndex(String)`. Removed the inaccurate overload reference; the doc now describes the per-read case sensitivity engines like Spark drive from `spark.sql.caseSensitive`, which is the actual analogue. (30e24ee)

JingsongLi (approved):
+1

PR #497, opened 2026-07-10
feat(datafusion): support REST Catalog CREATE VIEW

JingsongLi:
## Summary

Add end-to-end persistent `CREATE VIEW` support for Paimon REST Catalogs through `SQLContext`. The defining query is planned in the view's owning namespace, its schema is inferred by DataFusion, and the Java-compatible view request is persisted through the catalog API.

## Changes

- add `CreateViewRequest`, `ViewSchema::new`, `Catalog::create_view`, and `ViewAlreadyExist`
- implement REST create-view transport, HTTP 409 mapping, `IF NOT EXISTS`, and HTTP 501 fallback
- support `CREATE VIEW [IF NOT EXISTS] name [(columns)] AS query` in `SQLContext`
- infer field types/nullability, preserve the canonical DataFusion query, and bind bare relations/functions to the target catalog and database
- validate explicit column lists and reject unsupported persistent-view modifiers instead of silently ignoring them
- document the supported syntax and current limitations in the DataFusion README and SQL guide

## Testing

- [x] `cargo test -p paimon --test rest_object_models_test --test rest_api_test --test rest_catalog_test`
- [x] `cargo test -p paimon-datafusion --lib sql_context::tests -- --test-threads=1`
- [x] `cargo test -p paimon-datafusion --lib persistent_rest_catalog_view_ -- --test-threads=1`
- [x] `cargo clippy -p paimon -p paimon-datafusion --lib --tests -- -D warnings`
- [x] `mkdocs build --strict`

## Notes

- Persistent `CREATE VIEW` is implemented by REST Catalog; other catalog implementations keep the default `Unsupported` behavior.
- `CREATE OR REPLACE VIEW`, materialized/secure views, comments/options, persistent `ALTER VIEW` / `DROP VIEW`, and persistent function DDL remain out of scope.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -30,10 +30,11 @@
 //! - `ALTER TABLE db.t RENAME COLUMN old TO new`
 //! - `ALTER TABLE db.t RENAME TO new_name`
 //! - `ALTER TABLE db.t DROP PARTITION (col = val, ...)`
+//! - `CREATE VIEW [IF NOT EXISTS] view [(col, ...)] AS query`
 //! - `TRUNCATE TABLE db.t`
 //! - `TRUNCATE TABLE db.t PARTITION (col = val, ...)`
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use datafusion::arrow::array::{
@@ -48,6 +49,7 @@ use datafusion::datasource::{MemTable, TableProvider};
 use datafusion::error::{DataFusionError, Result as DFResult};
 use datafusion::execution::SessionStateBuilder;
 use datafusion::prelude::{DataFrame, SessionContext};
+use datafusion::sql::planner::IdentNormalizer;
 use datafusion::sql::sqlparser::ast::{
     AlterTableOperation, BinaryLength, CharacterLength, ColumnDef, ColumnOption, CreateTable,
     CreateTableOptions, CreateView, Delete, Expr as SqlExpr, FromTable, Insert, Merge, ObjectName,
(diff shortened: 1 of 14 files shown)


crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1514,6 +1569,72 @@ impl SQLContext {
     }
 }
 
+fn validate_persistent_create_view(create_view: &CreateView) -> DFResult<()> {
+    let unsupported = if create_view.or_alter {
+        Some("CREATE OR ALTER VIEW is not supported")
+    } else if create_view.or_replace {
+        Some("CREATE OR REPLACE VIEW is not supported")
+    } else if create_view.secure {
```
leaves12138, line 1577: Please reject `COPY GRANTS` here as well and add it to the unsupported-clause regression cases. `sqlparser::ast::CreateView` exposes this modifier through `copy_grants`, but the current validator never checks it, so `CREATE VIEW ... COPY GRANTS AS SELECT ...` succeeds while silently dropping the requested grant-copy semantics. This also contradicts the documentation that vendor-specific modifiers are unsupported.

crates/paimon/src/catalog/rest/rest_catalog.rs
```diff
@@ -278,6 +278,22 @@ impl Catalog for RESTCatalog {
...
+        &self,
+        identifier: &Identifier,
+        schema: crate::catalog::ViewSchema,
+        ignore_if_exists: bool,
+    ) -> Result<()> {
+        let result = self
+            .api
+            .create_view(identifier, schema)
+            .await
+            .map_err(|error| map_rest_error_for_view(error, identifier));
```
leaves12138: The create operation needs an operation-specific 404 mapping. For `POST /databases/{db}/views`, HTTP 404 means the parent database does not exist, but `map_rest_error_for_view` converts it to `Error::ViewNotExist`. Java `RESTCatalog.createView` maps the same response to `DatabaseNotExistException`. Please preserve that contract here and add a missing-database test.

leaves12138 (changes requested):
Thanks for adding persistent REST view creation. The main flow and the existing targeted tests look good, but I found two correctness issues that should be addressed before merge. I reproduced both with focused regression cases locally.

crates/integrations/datafusion/src/sql_context.rs
JingsongLi: Fixed in 25838b1. validate_persistent_create_view now rejects CreateView::copy_grants explicitly, and COPY GRANTS is covered by the unsupported-clause regression test. The focused persistent REST view test suite passes.

crates/paimon/src/catalog/rest/rest_catalog.rs
JingsongLi: Fixed in 25838b1. RESTCatalog create_view now uses an operation-specific mapper that converts a create-time 404 to DatabaseNotExist for the parent database, while preserving the existing 409 and 501 mappings. Added a missing-database regression test; all create-view catalog tests pass.

leaves12138 (approved):
Re-reviewed the latest changes. Both previous issues are fixed correctly: COPY GRANTS is explicitly rejected with regression coverage, and CREATE VIEW now maps a missing parent database to DatabaseNotExist. The focused DataFusion and REST Catalog tests, formatting check, and diff check all pass locally. LGTM.

PR #499, opened 2026-07-11
feat(datafusion): support REST Catalog CREATE FUNCTION

JingsongLi:
## Summary

Add persistent `CREATE FUNCTION` support for SQL scalar functions in Paimon REST Catalogs. The implementation uses the existing REST SQL function expansion path to validate a candidate definition before persisting it, so newly created functions follow the same execution semantics as catalog functions read from REST.

## Changes

- Add the Java-compatible create-function REST request, catalog API, POST endpoint, and create-specific 404/409/501 error mapping.
- Support bare, two-part, and three-part creation targets, named nullable parameters, a single nullable return value, and scalar `RETURN` expressions.
- Align the recommended SQL syntax with Databricks: `LANGUAGE SQL` is optional with SQL as the default, and determinism is inferred from the planned expression instead of requiring `IMMUTABLE`; explicit `IMMUTABLE` remains accepted for compatibility.
- Add token-level adapters for `IF NOT EXISTS` and `OR ALTER`, including comments and quoted identifiers, without rewriting function bodies.
- Validate candidate functions before POST using typed placeholders, owning-catalog/database resolution, REST dependency expansion, logical and physical DataFusion planning, recursion detection, volatility checks, and scalar-only restrictions.
- Document supported syntax, resolution rules, validation behavior, and unsupported function forms.

## Testing

- [x] `cargo test -p paimon --test rest_object_models_test`
- [x] `cargo test -p paimon --test rest_api_test --test rest_catalog_test`
- [x] `cargo test -p paimon-datafusion --lib persistent_rest_catalog_function_`
- [x] `cargo clippy -p paimon -p paimon-datafusion --all-targets -- -D warnings`
- [x] `cargo fmt --all -- --check` and `git diff --check`

## Notes

Persistent function creation is implemented by REST Catalog; other catalog implementations return `Unsupported` by default. Aggregate, table, multi-return, non-SQL, Stable/Volatile, lambda/file, temporary, alter, and drop function forms remain out of scope.

A full local `paimon-datafusion --lib` run passed 260 tests; seven existing table-scan tests require pre-populated `partitioned_log_table` fixtures under `/tmp/paimon-warehouse`, which are not present in this environment.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -1557,22 +1557,14 @@ impl SQLContext {
             .unwrap_or_default()
             .iter()
             .map(|field| {
-                let value = serde_json::to_value(field.data_type()).map_err(|error| {
-                    DataFusionError::Plan(format!(
-                        "Invalid CREATE FUNCTION argument type '{:?}': {error}",
-                        field.data_type()
-                    ))
-                })?;
-                let sql_type = value.as_str().ok_or_else(|| {
-                    DataFusionError::Plan(format!(
-                        "CREATE FUNCTION argument type '{:?}' cannot be represented in SQL",
-                        field.data_type()
-                    ))
-                })?;
-                Ok(format!(
-                    "CAST(NULL AS {})",
-                    sql_type.strip_suffix(" NOT NULL").unwrap_or(sql_type)
-                ))
+                let sql_type =
+                    crate::table::data_type_to_sql(field.data_type()).map_err(|error| {
+                        DataFusionError::Plan(format!(
+                            "Invalid CREATE FUNCTION argument type '{:?}': {error}",
+                            field.data_type()
+                        ))
+                    })?;
+                Ok(format!("CAST(NULL AS {sql_type})"))
             })
             .collect::<DFResult<Vec<_>>>()?
             .join(", ");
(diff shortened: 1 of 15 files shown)


crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1430,6 +1442,165 @@ impl SQLContext {
...
+            .unwrap_or_default()
+            .iter()
+            .map(|field| {
+                let value = serde_json::to_value(field.data_type()).map_err(|error| {
+                    DataFusionError::Plan(format!(
+                        "Invalid CREATE FUNCTION argument type '{:?}': {error}",
+                        field.data_type()
+                    ))
+                })?;
+                let sql_type = value.as_str().ok_or_else(|| {
```
leaves12138: Complex signature types are accepted by the DDL parser and converted into Paimon `DataType`s, but this assumes every serialized type is a JSON string. `ARRAY`, `MAP`, and `ROW` serialize as structured JSON, so, for example, `CREATE FUNCTION array_answer(x ARRAY<BIGINT>) RETURNS BIGINT RETURN 42` fails here before the REST request is sent. The same assumption in `sql_function.rs` also makes complex return types unusable (for example, `RETURNS ARRAY<BIGINT>`). Please reuse/extract the recursive SQL type renderer currently in `table/mod.rs` for both validation placeholders and return casts, or explicitly reject and document unsupported complex signature types, with regression tests for both argument and return types.

leaves12138 (changes requested):
I found one blocking issue around complex argument/return types. The existing targeted suites and formatting checks pass, but focused regression cases with `ARRAY<BIGINT>` reproduce failures before the function can be created or executed.

crates/integrations/datafusion/src/sql_context.rs
JingsongLi: Fixed in b0dbc50. CREATE FUNCTION validation placeholders and REST SQL function return casts now both reuse the recursive table::data_type_to_sql renderer, so ARRAY/MAP/ROW types no longer depend on JSON string serialization. Added regression tests for ARRAY input creation and ARRAY return creation/execution. Verified 17 persistent CREATE FUNCTION tests, 15 REST SQL function tests, fmt/diff checks, and clippy with -D warnings.

leaves12138 (approved):
Re-reviewed b0dbc50. The shared recursive SQL type renderer now handles both CREATE FUNCTION validation placeholders and REST SQL function return casts. The added ARRAY regressions pass, and I also verified MAP and ROW argument signatures locally. The persistent CREATE FUNCTION suite, REST SQL function suite, and formatting check all pass. LGTM.

PR #500, opened 2026-07-11
[spec] Support parsing REST catalog Predicate and Transform JSON

plusplusjiajia:
### Purpose

  Deserialize the REST catalog's serialized `Predicate` and `Transform` JSON
  (the row filter / column masking a `query-auth.enabled` table returns) into
  `spec::Predicate` / `spec::Transform`. Parse-only — needed by the query-auth
  enforcement follow-up.

  ### Changes

  - `Predicate::from_rest_json` + `collect_leaf_field_indices`
  - `spec::Transform` / `TransformInput` + `Transform::from_rest_json` +
    `collect_field_indices`
  - Field references resolve by name; the local field type drives literal
    conversion. Tests pin wire strings from Java's `PredicateJsonSerdeTest` /
    `TransformJsonSerdeTest`.

  ### Scope

  Structural parsing only, following the crate's existing pattern for
  REST-serialized payloads (`Schema` uses `#[derive(Deserialize)]`; the shared
  `DataType` parser already accepts lenient type strings for schema columns).
  It does not replicate the Java parser bit-for-bit:

  - `CAST` is parsed structurally; cast applicability is checked where the mask
    is applied (arrow cast + output-type equality), not here.
  - `LIKE` uses SQL escape semantics; row-filter evaluation reuses the shared
    residual evaluator (`arrow/residual.rs`), so `Float/Double.compareTo` and
    `LIKE _` edge semantics are Arrow's — a documented limitation

### DIFF
--- a/crates/paimon/src/spec/mod.rs
+++ b/crates/paimon/src/spec/mod.rs
@@ -102,7 +102,8 @@ pub(crate) use predicate::extract_datum;
 pub(crate) use predicate::java_bytes_cmp;
 pub(crate) use predicate::like_match;
 pub use predicate::{
-    field_idx_to_partition_idx, Datum, Predicate, PredicateBuilder, PredicateOperator,
+    field_idx_to_partition_idx, Datum, Predicate, PredicateBuilder, PredicateOperator, Transform,
+    TransformInput,
 };
 pub(crate) mod murmur_hash;
 mod partition_statistics;
--- a/crates/paimon/src/spec/predicate.rs
+++ b/crates/paimon/src/spec/predicate.rs
@@ -606,11 +606,9 @@ impl fmt::Display for Predicate {
 // ---------------------------------------------------------------------------
 
 impl Predicate {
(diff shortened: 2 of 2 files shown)


JingsongLi:
Incompatible with Java legitimate wire format: Java will generate 'LEAF+transform=NULL+function=TRUS/FALSE', which Rust currently explicitly rejects. The result is a failed valid query auth condition (fail closed, not unauthorized, but will erroneously reject the query).

plusplusjiajia:
> Incompatible with Java legitimate wire format: Java will generate 'LEAF+transform=NULL+function=TRUS/FALSE', which Rust currently explicitly rejects. The result is a failed valid query auth condition (fail closed, not unauthorized, but will erroneously reject the query).

@JingsongLi Thanks for catching this! Fixed — TRUE/FALSE leaves with the NULL transform now parse to AlwaysTrue/AlwaysFalse, with the PredicateJsonSerdeTest wire strings pinned in tests.

JingsongLi (approved):
+1

PR #507, opened 2026-07-11
feat: support rowkind.field write semantics

### LINKED ISSUE #502: Support rowkind.field for CDC-style writes on deduplicate PK tables
### Describe the feature

I'd like to write changelog-style rows into a primary-key table by putting the row kind (`+I` / `-U` / `+U` / `-D`) in a user column, matching Java's `rowkind.field` option.

Today the Rust write path fails loud with something like "KeyValueFileWriter does not support rowkind.field", even though the option can already be parsed.

### Motivation

This is useful for CDC / ETL pipelines that already carry a short string op column and don't want to invent a separate `_VALUE_KIND` convention on the client side.

### Proposed scope

- Parse short-string row kinds (`+I`, `-U`, `+U`, `-D`)
- Honor `rowkind.field` on **deduplicate** PK tables only
- Validate the configured field (exists, STRING, not conflicting with system columns)
- Support `ignore-delete` / `ignore-update-before` filtering when writing
- Integration tests aligned with the Java ITCase scenarios I could map over

Out of scope for this issue: other merge engines, lookup/compact, DataFusion SQL surface.

### Additional context

Happy to send a PR if this direction looks good.

Pandas886:
### Purpose

Linked issue: close #502

I ran into the current fail-loud path when trying to write CDC-style rows with a user op column (`rowkind.field`). Java already supports this, and the Rust option is parsed but writing still rejects it. This PR wires up the write path for **deduplicate** PK tables.

### Brief change log

- Add short-string `RowKind` parsing (`+I` / `-U` / `+U` / `-D`, case-insensitive)
- Introduce `RowKindGenerator` to map a STRING column onto internal `_VALUE_KIND`
- Schema checks: PK table, deduplicate only, field exists and is unbounded STRING, reject bad combos
- `TableWrite` enrich step: convert row kinds, optionally drop retracts via `ignore-delete` / `ignore-update-before`
- Reject batches that already contain `_VALUE_KIND` when `rowkind.field` is set (fail loud, no silent override)
- Integration tests for the main happy paths + illegal tokens / conflict cases

I deliberately left other merge engines alone — they still get a clear validation error if someone sets `rowkind.field` there.

### Tests

Locally:

```bash
cargo test -p paimon --test rowkind_field_test
cargo test -p paimon row_kind
cargo fmt --all -- --check
```

All green on my machine.

### API and Format

- Public API: short-string helpers on `RowKind`, plus the existing `rowkind.field` option now actually works on write
- Storage format: unchanged (still writes normal PK files with `_VALUE_KIND`)

### Documentation

No user-facing docs update in this PR. Happy to add a short note to the write docs if reviewers want it.

### Notes

This is independent of the incremental-read PR stack I also opened. Review can land in either order.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -71,6 +71,13 @@ pub(crate) const DISABLE_ALTER_COLUMN_NULL_TO_NOT_NULL_OPTION: &str =
 const MERGE_ENGINE_OPTION: &str = "merge-engine";
 const CHANGELOG_PRODUCER_OPTION: &str = "changelog-producer";
 const ROWKIND_FIELD_OPTION: &str = "rowkind.field";
+const IGNORE_DELETE_OPTION: &str = "ignore-delete";
+const IGNORE_UPDATE_BEFORE_OPTION: &str = "ignore-update-before";
+const IGNORE_DELETE_FALLBACK_KEYS: &[&str] = &[
+    "first-row.ignore-delete",
+    "deduplicate.ignore-delete",
+    "partial-update.ignore-delete",
+];
 const DEFAULT_COMMIT_MAX_RETRIES: u32 = 10;
 const DEFAULT_COMMIT_TIMEOUT_MS: u64 = 120_000;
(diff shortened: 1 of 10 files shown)


JingsongLi (approved):
+1

PR #508, opened 2026-07-11
feat: support batch incremental delta scans

Pandas886:
### Purpose

Linked issue: close #503

I'd like a small public API for batch incremental reads — basically "files between snapshot A (exclusive) and B (inclusive)" — without callers reinventing snapshot walking on top of `TableScan`.

This PR is intentionally thin: **Delta + Auto selection only**. Changelog / Diff come in follow-up PRs so review stays focused.

### Brief change log

- Add `IncrementalScanMode`, `IncrementalScan`, `IncrementalPlan`, `IncrementalSplit`
- Range semantics: left-open / right-closed, empty range, out-of-bounds errors
- Delta: plan data files from APPEND snapshots in the range via existing scan helpers
- Auto: `changelog-producer=none` → Delta; otherwise resolves to Changelog (planning Changelog still fails loud until the next PR)
- Push partition/bucket filters through from `ReadBuilder` when planning
- `TableRead::to_incremental_arrow` for the data-split case

No new manifest reader code — it reuses what `TableScan` already does.

### Tests

```bash
cargo test -p paimon --test incremental_batch_scan_test
cargo test -p paimon --lib
cargo fmt --all -- --check
cargo clippy -p paimon --all-targets --features fulltext,vortex -- -D warnings
```

### API and Format

- New public scan types under `paimon::table` (also re-exported at crate root)
- No storage format change

### Documentation

None yet — waiting until the SQL TVF lands so we document the whole story once. Can add a short rustdoc note if preferred.

### Follow-ups

- #509 Changelog + AuditLogTable
- #510 Diff mode
- #511 DataFusion `paimon_incremental_query`

PR #509, opened 2026-07-11
feat: support existing changelog incremental reads

### LINKED ISSUE #504: Batch incremental scan: read existing changelog manifests + AuditLogTable
### Describe the feature

Extend the batch incremental scan API to read **existing** changelog files from snapshots (`changelog_manifest_list`), and expose an `AuditLogTable`-style read that surfaces row kinds as `+I` / `-U` / `+U` / `-D`.

### Motivation

Tables written with `changelog-producer=input` (and other producers that already materialize changelog files) should be readable incrementally without generating changelog on the Rust side.

### Proposed scope

- Changelog planning over snapshots in `(start, end]`
- Skip OVERWRITE / missing changelog lists cleanly
- `AuditLogTable` (or equivalent) with `rowkind` as the first column
- Tests with input-changelog fixtures (no compact/lookup generation in this PR)

Depends on the Delta/Auto incremental API landing first.

### Out of scope

- Generating lookup/compact changelogs in Rust
- Diff mode (derive changelog by comparing before/after images)

Pandas886:
### Purpose

Linked issue: close #504

**Depends on:** #508 (Delta/Auto). Please review that one first — this branch is stacked on top of it, so the diff against `main` currently includes both commits.

This adds reading of **already-written** changelog files (`changelog_manifest_list`) plus an `AuditLogTable` view that exposes `rowkind` as the first column.

Important: this does **not** generate lookup/compact changelogs. If a snapshot has no changelog list, we just skip it / return empty for that snapshot rather than inventing files.

### Brief change log

- Implement Changelog planning over `(start, end]`
- `TableScan::plan_snapshot_changelog` reusing the existing split pipeline
- `AuditLogTable` wrapper: `rowkind` first, optional `_SEQUENCE_NUMBER`
- `TableRead::to_audit_log_arrow` for Delta (all `+I`) and Changelog (from `_VALUE_KIND`)
- Auto → Changelog when producer is not `none`
- Tests with `changelog-producer=input` fixtures and multi-snapshot ranges

Diff mode is still unsupported here on purpose.

### Tests

```bash
cargo test -p paimon --test incremental_batch_scan_test
cargo test -p paimon --test audit_log_table_test
cargo fmt --all -- --check
```

### API and Format

- New `AuditLogTable` + Changelog path on the incremental API
- No storage format change; only reads existing changelog manifests/files

### Documentation

No standalone docs in this PR (SQL docs come with the DataFusion follow-up).

### Review tip

If you're looking at the full branch vs `main`, the interesting commit for *this* PR is the top one (`feat: support existing changelog incremental reads`). The parent is #508.

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -811,6 +811,11 @@ pub const VALUE_KIND_FIELD_NAME: &str = "_VALUE_KIND";
 /// Must match Java Paimon's `SpecialFields.VALUE_KIND` (Integer.MAX_VALUE - 2).
 pub const VALUE_KIND_FIELD_ID: i32 = i32::MAX - 2;
 
+pub const ROW_KIND_FIELD_NAME: &str = "rowkind";
+
+/// Must match Java Paimon's `SpecialFields.ROW_KIND` (Integer.MAX_VALUE - 4).
+pub const ROW_KIND_FIELD_ID: i32 = i32::MAX - 4;
+
 /// Data field for paimon table.
 ///
 /// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-common/src/main/java/org/apache/paimon/types/DataField.java#L40>
--- a/crates/paimon/src/table/audit_log_table.rs
+++ b/crates/paimon/src/table/audit_log_table.rs
@@ -18,14 +18,14 @@
 use super::incremental_scan::{IncrementalPlan, IncrementalScan, IncrementalScanMode};
 use super::{ArrowRecordBatchStream, Table};
 use crate::spec::{
-    BigIntType, DataField, DataType, VarCharType, SEQUENCE_NUMBER_FIELD_ID,
-    SEQUENCE_NUMBER_FIELD_NAME,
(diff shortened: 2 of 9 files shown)


JingsongLi:
Please resolve conflicts.

JingsongLi:
The Delta audit at `table_read.rs:243` calls `audit_raw_stream(plan, false)` and then hard-codes all rowkinds to `+I`. Even when `changelog-producer=none` is set for the PK table, Delta files still contain `_VALUE_KIND`; Java incremental audit tests explicitly require types such as `-D` to be preserved, so delete/update records are incorrectly reported as inserts. It is recommended that PK Delta also read `_VALUE_KIND` and add regression tests for DELETE/UPDATE operations. Additionally, the `rowkind` field uses the ID `-1` and `varchar(8)`, which is inconsistent with Java’s `Integer.MAX_VALUE-4` and the maximum `varchar` length; synchronization is recommended. The existing incremental/audit/exception `rowkind` tests have passed 16 out of 16, but do not cover Delta delete records.

Pandas886:
Fixed in 2f72b51 — PK Delta audit now preserves _VALUE_KIND, and rowkind matches Java’s field ID and varchar length. Added regression coverage too. Thanks for catching this.

JingsongLi (approved):
+1

PR #510, opened 2026-07-11
feat: support batch incremental diff reads

### LINKED ISSUE #505: Batch incremental scan: Diff mode for deduplicate PK tables
### Describe the feature

Add **Diff** incremental planning/reading: compare table state at two snapshots and emit the net changes (insert / update-before / update-after / delete) when no pre-materialized changelog is available.

### Motivation

Many tables run with `changelog-producer=none`. Diff is the practical way to still produce an audit/changelog stream for batch incremental jobs.

### Proposed scope (phase 1)

- Diff planning as per-(partition, bucket) before/after pairs
- Fail loud (no silent wrong answers) for:
  - non-deduplicate merge engines
  - bucket rescale between snapshots
  - deletion vectors / nested/decimal types we can't safely compare yet
- Audit-style row kinds for the diff result

Depends on the base incremental scan + AuditLog read path.

### Out of scope

- Full DV-aware diff
- Non-deduplicate merge engines

Pandas886:
### Purpose

Linked issue: close #505

**Depends on:** #508 and #509 (stacked). Against `main` you'll see the whole stack; the new work is Diff planning + before/after comparison.

For tables with `changelog-producer=none`, Diff is how you still get an audit-style change stream without materializing changelog during write.

### Brief change log

- Diff planning: before/after snapshot state → `DiffPair` per `(partition, bucket)`
- Reject bucket rescale between snapshots (explicit `Unsupported`, no silent wrong results)
- Read path: PK-sorted merge of before/after images, emit `+I` / `-U` / `+U` / `-D`, skip equal rows
- Phase-1 limits (fail loud):
  - deduplicate only
  - no deletion-vector support yet
  - nested / decimal types rejected for now
- `diff.parallelism` option (default 4) for pair execution
- After-image mode on plain incremental read (non-audit) returns the post-change rows only

I used input-delete rows for delete cases in tests so we don't need compact/lookup generation in this PR.

### Tests

```bash
cargo test -p paimon --test incremental_batch_scan_test
cargo test -p paimon --test audit_log_table_test
cargo fmt --all -- --check
```

### API and Format

- Diff mode becomes usable on the existing incremental / AuditLog APIs
- No storage format change

### Documentation

Still holding docs for the DataFusion PR so users see one end-to-end story.

### Known gaps (intentionally)

- Real DV-backed tables: unit-level fail-loud only, no full e2e with DV files yet
- Non-deduplicate engines: hard reject

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -89,6 +89,8 @@ const IGNORE_DELETE_FALLBACK_KEYS: &[&str] = &[
     "deduplicate.ignore-delete",
     "partial-update.ignore-delete",
 ];
+const DIFF_PARALLELISM_OPTION: &str = "diff.parallelism";
+const DEFAULT_DIFF_PARALLELISM: usize = 4;
 const DEFAULT_COMMIT_MAX_RETRIES: u32 = 10;
 const DEFAULT_COMMIT_TIMEOUT_MS: u64 = 120_000;
 const DEFAULT_COMMIT_MIN_RETRY_WAIT_MS: u64 = 1_000;
@@ -510,6 +512,17 @@ impl<'a> CoreOptions<'a> {
         self.options
             .get(IGNORE_UPDATE_BEFORE_OPTION)
             .is_some_and(|v| v.eq_ignore_ascii_case("true"))
+    }
+
+    /// Parallelism for batch incremental Diff pair reads (`diff.parallelism`).
+    ///
+    /// Default is 4; values below 1 are clamped to 1.
+    pub fn diff_parallelism(&self) -> usize {
+        self.options
+            .get(DIFF_PARALLELISM_OPTION)
+            .and_then(|s| s.parse().ok())
+            .unwrap_or(DEFAULT_DIFF_PARALLELISM)
+            .max(1)
     }
 
     pub fn data_evolution_enabled(&self) -> bool {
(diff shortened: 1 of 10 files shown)


JingsongLi:
There are multiple correctness issues:
- When comparing the null Arrow value without checking its validity, 'NULL ->0' will be judged as unchanged; The minimum regression I added actually returns' [] 'and expects' (1,0)'.
- Diff only compares projection columns; When the value changes without projection, it will be missed, and if the projection does not include PK, it will directly report an error. The second minimum regression also returns an empty result.
- Claiming to support types larger than the actual support range of comparator, runtime failures such as timestamp/time/financial; NaN will also be treated as equal.
- Using serialized 'Binary Row' bytes for key range lexicographic comparison and pruning files based on it has unreliable logical order and a risk of data leakage.

Pandas886:
Thanks, I addressed these in 7b47806 and rebased onto latest main. Diff now compares the full schema before projection (including NULLs), rejects unsupported types, and no longer uses BinaryRow bytes for range pruning. NaN follows Java Float.compare. I also added the NULL/projection regressions. Could you take another look?

JingsongLi (commented):
Can you add API doc in docs?

Pandas886:
Added the API guide in `docs/src/incremental-reading.md` and linked it from the docs navigation. I also tightened the Diff guards and updated the regression tests. Thanks for the reminder!

JingsongLi:
`validate_diff_bucket_layout` compares only the partitions that exist in both the before and after snapshots. If the table is empty before and after a bucket rescale, or if the partitions do not overlap at all, the validation will be skipped and the Diff will continue to run, which may produce incorrect results. The added regression test for “different partitions, bucket 1→2” has been confirmed to fail. We recommend validating the bucket configuration directly based on the schemas of the two snapshots, similar to Java’s `TimeTravelUtil`.

Pandas886:
Good catch — fixed in 56cf37b. Diff now compares the bucket count from both snapshot schemas before reading manifests, following Java's TimeTravelUtil. I also added an empty-snapshot regression test. Thanks!

JingsongLi (approved):
+1

PR #511, opened 2026-07-11
feat: support DataFusion paimon_incremental_query

### LINKED ISSUE #506: DataFusion: paimon_incremental_query table function
### Describe the feature

Expose batch incremental reads in DataFusion SQL via a table function, roughly:

```sql
SELECT * FROM paimon_incremental_query(
  'table_name',
  start_snapshot_exclusive,
  end_snapshot_inclusive
  [, 'auto' | 'delta' | 'changelog' | 'diff']
);
```

Also support the `$audit_log` style suffix so row kinds show up as a first column when needed.

### Motivation

The Rust core incremental API is useful, but most people will want to poke at this from SQL. DataFusion should only parse args and adapt the execution plan — snapshot planning stays in `paimon` core.

### Depends on

Core incremental Delta / Changelog / Diff support landing first (or the PR can sit on top of that stack).

### Out of scope

- Reimplementing snapshot planning inside the DataFusion crate
- Lookup/compact changelog generation

Pandas886:
### Purpose

Linked issue: close #506

This PR adds the DataFusion SQL surface for the batch incremental read APIs now available on `main`.

```sql
SELECT * FROM paimon_incremental_query('t', 0, 2);
SELECT * FROM paimon_incremental_query('t', 0, 2, 'diff');
SELECT rowkind, id, value FROM paimon_incremental_query('t$audit_log', 0, 2, 'changelog');
```

### Brief change log

- Register TVF `paimon_incremental_query` in `SQLContext`
- 3-arg form defaults to `auto`; optional 4th arg: `auto` / `delta` / `changelog` / `diff`
- Delegate physical planning to core `IncrementalScan` / `AuditLogTable`
- Support `$audit_log` suffix for rowkind output
- Support projection and residual filters
- Resolve qualified table names against the named catalog; unqualified names follow the current catalog
- Update `docs/src/sql.md` with usage and examples

### Tests

```bash
cargo test -p paimon-datafusion --test incremental_query
cargo fmt --all -- --check
cargo clippy -p paimon-datafusion --all-targets -- -D warnings
```

### API and Format

- New SQL table function only; storage format unchanged
- Relies on the core incremental APIs already merged into `main`

### Documentation

`docs/src/sql.md` documents registration, arguments, modes, and examples.

### DIFF
--- a/docs/src/sql.md
+++ b/docs/src/sql.md
@@ -73,9 +73,9 @@ the catalog-independent `path_to_descriptor` and `descriptor_to_string` scalar
 functions pre-registered. Use `register_catalog(...).await` to add one or more
 Paimon catalogs; registering a catalog also registers the built-in scalar
 function `blob_view` (alias `sys.blob_view`) and the built-in table-valued
-functions (`vector_search`, `hybrid_search`, and `full_text_search` when the
-`fulltext` feature is enabled) against it. It also manages session-scoped
-dynamic options internally for `SET`/`RESET` support.
+functions (`paimon_incremental_query`, `vector_search`, `hybrid_search`, and
+`full_text_search` when the `fulltext` feature is enabled) against it. It also
+manages session-scoped dynamic options internally for `SET`/`RESET` support.
 
 ### REST Catalog Views and SQL Functions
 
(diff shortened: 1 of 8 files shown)


JingsongLi:
The Diff problem of inheriting '# 510'; The non audit mode will first push the SQL projection down to the core, directly triggering the issue of "only comparing projection columns/missing PK". In addition, the catalog section of 'catalog. db. table' is discarded after parsing, and the TVF global registration will be overwritten by the last registered catalog. Multiple catalog scenarios may result in reading the wrong catalog.

Pandas886:
Rebased on the latest main and fixed catalog/database resolution. Added coverage for multiple catalogs and USE database. Could you take another look?

docs/src/sql.md
```diff
@@ -1169,6 +1169,72 @@ CROSS JOIN LATERAL vector_search(
...
+
+When you use a `SQLContext`, `paimon_incremental_query` is registered
+automatically for every catalog you register — no extra setup is needed.
+
+With a raw DataFusion `SessionContext`, register it explicitly:
+
+```rust
+use paimon_datafusion::register_incremental_query;
+
+register_incremental_query(&ctx, catalog.clone(), "default");
```
JingsongLi, line 1189: register by default? so user don't need to register this again.

JingsongLi:
Close now, feel free to create new one.

PR #518, opened 2026-07-13
fix(datafusion): skip view lookup for table functions

shyjsarah:
### Purpose

Linked issue: N/A (hotfix for a regression introduced by #494)

DataFusion preloads relation names before planning. For a query such as `FROM vector_search(...)`, it may first ask the current schema for a table named `vector_search`. After REST view support was added, a missing table was reinterpreted as a view, so an error from `get_view` could abort planning before DataFusion resolved the registered table function.

This change prevents registered table functions from being incorrectly resolved as REST catalog views.

### Brief change log

- Skip REST view fallback when a missing relation name matches a registered table function.
- Keep ordinary table and REST view resolution unchanged.
- Add a regression test using the built-in `vector_search` table function and a generic REST catalog.

### Tests

- `cargo test -p paimon-datafusion --lib sql_context::tests::registered_table_function_skips_view_lookup_during_relation_preload -- --exact --nocapture`
- `cargo test -p paimon-datafusion --lib rest_catalog_view` — 26 passed
- `cargo clippy -p paimon-datafusion --lib -- -D warnings`

The full `paimon-datafusion` library test run with `fulltext` reached 274 passing tests. Seven existing scan tests could not load their expected table fixtures in this local environment.

### API and Format

### Documentation

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -447,6 +447,16 @@ impl SchemaProvider for PaimonSchemaProvider {
                     if branch.is_some() {
                         return Ok(None);
                     }
+                    // DataFusion preloads every relation name before planning, including
+                    // registered table functions. Do not reinterpret a missing UDTF name as
+                    // a REST view; the planner will resolve it through the UDTF registry.
+                    if session_state
+                        .as_ref()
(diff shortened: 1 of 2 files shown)


jerry-024 (approved):
+1

PR #519, opened 2026-07-13
feat(datafusion): expose hybrid search scores

QuakeWang:
### Purpose

DataFusion hybrid search currently discards the fused scores returned by `execute_scored`, so queries cannot access the Spark-compatible `__paimon_search_score` metadata column.

### Brief change log

  - Expose `__paimon_search_score` from hybrid-search table functions.
  - Match scanned rows through `_ROW_ID` and append scores without exposing the internal row ID.
  - Preserve projection, count, limit, empty-result, and row-tracking-only behavior.

### Tests

  - `cargo test -p paimon-datafusion physical_plan::search_score::tests`
  - `cargo test -p paimon-datafusion hybrid_search_tests`
  - `cargo test -p paimon-datafusion --features fulltext hybrid_search_tests`

### API and Format

Adds the nullable `Float32` metadata column `__paimon_search_score` to hybrid-search results. No storage-format change.

### Documentation

Updated the SQL documentation.

### DIFF
--- a/crates/integrations/datafusion/src/hybrid_search.rs
+++ b/crates/integrations/datafusion/src/hybrid_search.rs
@@ -33,7 +33,9 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use datafusion::arrow::array::Array;
-use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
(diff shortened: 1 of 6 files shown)


JingsongLi (approved):
+1

PR #521, opened 2026-07-14
fix(blob): support unknown-length BlobDescriptor ranges

QuakeWang:
### Purpose

`BlobDescriptor.length == -1` means reading from `offset` to EOF. The dedicated BLOB writer cast this value to `u64`, while the `blob-descriptor-field` resolver rejected it as negative, causing inconsistent behavior across table configurations.

### Brief change log

  - Centralize BlobDescriptor range validation and unknown-length handling.
  - Resolve `length == -1` in both dedicated BLOB writes and inline descriptor reads.
  - Reject invalid ranges and short reads instead of producing inconsistent blob data.
  - Add end-to-end coverage for both paths.

### Tests

  - `cargo test -p paimon-datafusion --test blob_tests`

### API and Format

No API or storage format change. This implements the existing BlobDescriptor protocol semantics.

### Documentation

Documented the `length == -1` behavior and range constraints.

### DIFF
--- a/crates/integrations/datafusion/tests/blob_tests.rs
+++ b/crates/integrations/datafusion/tests/blob_tests.rs
@@ -22,7 +22,7 @@
 mod common;
 
 use arrow_array::{Array, BinaryArray, Int32Array, RecordBatch, StringArray};
-use common::{create_sql_context, create_test_env, exec};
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #522, opened 2026-07-14
feat(c): add write/commit C FFI bindings

### LINKED ISSUE #520: [Feature] Add write/commit C FFI bindings for bindings/c/
## Background

[Apache Doris](https://github.com/apache/doris) (an MPP OLAP database) is adding Paimon write support. We currently have a working JNI-based write path that bridges to the Paimon Java SDK, but we'd like to add a Rust FFI backend for better performance (zero-copy Arrow, async I/O, no JVM overhead).

The existing `bindings/c/` already provides a complete read-side C FFI (catalog, table, scan, plan, predicate, record batch streaming via Arrow C Data Interface). We'd like to contribute write-side C FFI bindings following the same patterns.

## Current state

- `bindings/c/` — read-side C FFI ✅ (catalog, table, read builder, scan, plan, predicate, record batch reader)
- `bindings/python/` — Python bindings with full write support ✅ (`PyWriteBuilder`, `PyTableWrite`, `PyTableCommit`)
- `bindings/c/` — write-side C FFI ❌ (nothing yet)

## Proposed API surface

Following the existing naming conventions and opaque handle patterns from `bindings/c/`:

```
// === WriteBuilder ===
paimon_write_builder *paimon_table_new_write_builder(paimon_table *table);
void paimon_write_builder_free(paimon_write_builder *wb);
void paimon_write_builder_with_overwrite(paimon_write_builder *wb);

// === TableWrite ===
paimon_table_write *paimon_write_builder_new_write(paimon_write_builder *wb, paimon_error **err);
void paimon_table_write_free(paimon_table_write *tw);

// write a single Arrow record batch via C Data Interface (zero-copy)
paimon_error *paimon_table_write_write_batch(
    paimon_table_write *tw,
    FFI_ArrowArray *array,
    FFI_ArrowSchema *schema);

// close file writers and produce commit messages
// returns paimon_commit_messages* on success, NULL + err on failure
paimon_commit_messages *paimon_table_write_prepare_commit(
    paimon_table_write *tw, paimon_error **err);

// === TableCommit ===
paimon_table_commit *paimon_write_builder_new_commit(paimon_write_builder *wb);
void paimon_table_commit_free(paimon_table_commit *tc);

paimon_error *paimon_table_commit_commit(
    paimon_table_commit *tc, paimon_commit_messages *msgs);
paimon_error *paimon_table_commit_overwrite(
    paimon_table_commit *tc, paimon_commit_messages *msgs,
    ... static_partitions ...);
paimon_error *paimon_table_commit_truncate_table(paimon_table_commit *tc);
paimon_error *paimon_table_commit_abort(
    paimon_table_commit *tc, paimon_commit_messages *msgs);

// === CommitMessage serialization ===
// For cross-process transfer, commit messages need serialization.
// Proposal: use protobuf for schema-safe, language-agnostic serialization,
// matching Paimon's existing DataFileMeta / CommitMessage struct fields.
paimon_bytes paimon_commit_messages_serialize(paimon_commit_messages *msgs);
paimon_commit_messages *paimon_commit_messages_deserialize(
    paimon_bytes data, paimon_error **err);
void paimon_commit_messages_free(paimon_commit_messages *msgs);
```

## CommitMessage serialization — discussion needed

This is the key open question. Paimon's `CommitMessage` is a Rust struct with nested `Vec<DataFileMeta>`, `Vec<IndexFileMeta>`, etc. For C FFI consumers (like Doris) that need to transfer messages across a process boundary and eventually to a remote coordinator, serialization is required.

Options:

| Format | Pros | Cons |
|--------|------|------|
| **Protobuf** | Schema-safe, cross-language, Paimon already uses protobuf internally | Adds `prost-build` dependency to `bindings/c` |
| **JSON** | Simple, no schema needed | Bulky for DataFileMeta, slower |
| **FlatBuffers** | Zero-copy deserialization | More complex, less community familiarity |
| **Paimon binary format** | Reuses existing `CommitMessageSerializer` logic | Java-specific, needs porting to Rust |

We lean toward **Protobuf** — it's the standard approach, Paimon already uses it, and it generates bindings for all languages. But we'd like community input before committing.

## Implementation plan

1. Add `bindings/c/src/write.rs` with the API surface above
2. Add CommitMessage protobuf schema and serialization
3. Add tests under `bindings/c/tests/` (following Python test patterns)
4. Update `bindings/c/src/lib.rs` to register the new module

## Questions for the community

1. Does this API surface look reasonable? Any naming or pattern preferences?
2. **Protobuf vs JSON vs other formats for CommitMessage serialization?**
3. Should `write_arrow_batch` also accept a batch of batches (`&[RecordBatch]`) or keep it single-batch per call?
4. Any concerns about adding a protobuf build dependency to `bindings/c/`?

---

We're happy to implement this and contribute back. Looking forward to your feedback!

suxiaogang223:
## Summary

Add write-side C FFI bindings (`bindings/c/src/write.rs`) following the same opaque handle / `extern "C"` patterns as the existing read-side bindings. Closes #520.

## API surface (14 functions)

```
// WriteBuilder
paimon_table_new_write_builder(table)        → paimon_result_write_builder
paimon_write_builder_free(wb)                → void
paimon_write_builder_with_overwrite(wb)      → *mut paimon_error
paimon_write_builder_new_write(wb)           → paimon_result_table_write
paimon_write_builder_new_commit(wb)          → paimon_result_table_commit

// TableWrite
paimon_table_write_free(tw)                  → void
paimon_table_write_write_arrow_batch(tw, array, schema) → *mut paimon_error
paimon_table_write_prepare_commit(tw)        → paimon_result_prepare_commit

// TableCommit
paimon_table_commit_free(tc)                 → void
paimon_table_commit_commit(tc, msgs)         → *mut paimon_error
paimon_table_commit_overwrite(tc, msgs)      → *mut paimon_error
paimon_table_commit_truncate_table(tc)       → *mut paimon_error
paimon_table_commit_abort(tc, msgs)          → *mut paimon_error

// CommitMessages
paimon_commit_messages_free(msgs)            → void
```

## Design decisions

- **In-process transfer**: CommitMessages are passed as opaque handles (no serialization), matching Python bindings
- **Shared commit_user**: WriteBuilder generates a UUID shared by writer and committer
- **Arrow C Data Interface**: Batch import via `FFI_ArrowArray`/`FFI_ArrowSchema` raw pointers

## Tests (15 cases, all passing)

- Read path (3), Predicate (3), Write path (7), Error handling (1), Integration (1)

Run: `cargo +1.91.0 test -p paimon-c`

### DIFF
--- a/bindings/c/src/write.rs
+++ b/bindings/c/src/write.rs
@@ -1,0 +1,506 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::ffi::c_void;
+use std::ptr;
+
+use arrow_array::ffi::{from_ffi, FFI_ArrowArray, FFI_ArrowSchema};
+use arrow_array::{RecordBatch, StructArray};
+use paimon::table::{CommitMessage, Table, TableCommit, TableWrite};
+
+use crate::error::{check_non_null, paimon_error, PaimonErrorCode};
+use crate::result::{
+    paimon_result_prepare_commit, paimon_result_table_commit, paimon_result_table_write,
+    paimon_result_write_builder,
+};
+use crate::runtime;
+use crate::types::*;
+
+// ======================= WriteBuilder ===============================
+
+/// Create a new WriteBuilder from a Table.
+///
+/// The returned WriteBuilder holds a shared `commit_user` (UUID) that will be
+/// used by both `new_write()` and `new_commit()` for duplicate-commit detection.
+///
+/// # Safety
+/// `table` must be a valid pointer from `paimon_catalog_get_table`, or null (returns error).
+#[no_mangle]
+pub unsafe extern "C" fn paimon_table_new_write_builder(
+    table: *const paimon_table,
+) -> paimon_result_write_builder {
+    if let Err(e) = check_non_null(table, "table") {
+        return paimon_result_write_builder {
+            write_builder: ptr::null_mut(),
+            error: e,
+        };
+    }
+    let table_ref = &*((*table).inner as *const Table);
+    let wb = table_ref.new_write_builder();
+    let commit_user = wb.commit_user().to_string();
+    let state = WriteBuilderState {
+        table: table_ref.clone(),
+        commit_user,
+        overwrite: false,
+    };
+    let inner = Box::into_raw(Box::new(state)) as *mut c_void;
+    paimon_result_write_builder {
+        write_builder: Box::into_raw(Box::new(paimon_write_builder { inner })),
+        error: ptr::null_mut(),
+    }
+}
+
+/// Free a paimon_write_builder.
+///
+/// # Safety
+/// Only call with a write_builder returned from `paimon_table_new_write_builder`.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_write_builder_free(wb: *mut paimon_write_builder) {
+    if !wb.is_null() {
+        let wrapper = Box::from_raw(wb);
+        if !wrapper.inner.is_null() {
+            drop(Box::from_raw(wrapper.inner as *mut WriteBuilderState));
+        }
+    }
+}
+
+/// Enable overwrite mode for the WriteBuilder.
+///
+/// In overwrite mode, a subsequent `paimon_table_commit_overwrite` will replace
+/// the data in the written partitions rather than appending.
+///
+/// # Safety
+/// `wb` must be a valid pointer from `paimon_table_new_write_builder`, or null (returns error).
+#[no_mangle]
+pub unsafe extern "C" fn paimon_write_builder_with_overwrite(
+    wb: *mut paimon_write_builder,
+) -> *mut paimon_error {
+    if let Err(e) = check_non_null(wb, "wb") {
+        return e;
+    }
+    let state = &mut *((*wb).inner as *mut WriteBuilderState);
+    state.overwrite = true;
+    ptr::null_mut()
+}
+
+// ======================= TableWrite ===============================
+
+/// Create a new TableWrite from the WriteBuilder.
+///
+/// The returned TableWrite accumulates Arrow batches until `prepare_commit` is called.
+///
+/// # Safety
+/// `wb` must be a valid pointer from `paimon_table_new_write_builder`, or null (returns error).
+#[no_mangle]
+pub unsafe extern "C" fn paimon_write_builder_new_write(
+    wb: *const paimon_write_builder,
+) -> paimon_result_table_write {
+    if let Err(e) = check_non_null(wb, "wb") {
+        return paimon_result_table_write {
+            write: ptr::null_mut(),
+            error: e,
+        };
+    }
+    let state = &*((*wb).inner as *const WriteBuilderState);
+
+    let mut builder = match state
+        .table
+        .new_write_builder()
+        .with_commit_user(state.commit_user.clone())
+    {
+        Ok(b) => b,
+        Err(e) => {
+            return paimon_result_table_write {
+                write: ptr::null_mut(),
+                error: paimon_error::from_paimon(e),
+            }
+        }
+    };
+
+    if state.overwrite {
+        builder = builder.with_overwrite();
+    }
+
+    let tw = match builder.new_write() {
+        Ok(w) => w,
+        Err(e) => {
+            return paimon_result_table_write {
+                write: ptr::null_mut(),
+                error: paimon_error::from_paimon(e),
+            }
+        }
+    };
+
+    let inner = Box::into_raw(Box::new(tw)) as *mut c_void;
+    paimon_result_table_write {
+        write: Box::into_raw(Box::new(paimon_table_write { inner })),
+        error: ptr::null_mut(),
+    }
+}
+
+/// Free a paimon_table_write.
+///
+/// Dropping a TableWrite before calling `prepare_commit` discards any
+/// uncommitted data.
+///
+/// # Safety
+/// Only call with a write returned from `paimon_write_builder_new_write`.
+#[no_mangle]
+pub unsafe extern "C" fn paimon_table_write_free(tw: *mut paimon_table_write) {
+    if !tw.is_null() {
+        let wrapper = Box::from_raw(tw);
+        if !wrapper.inner.is_null() {
+            drop(Box::from_raw(wrapper.inner as *mut TableWrite));
+        }
+    }
+}
+
+/// Write a single Arrow record batch into the table's writers.
+///
+/// The Arrow data is imported via the Arrow C Data Interface. `array` and
+/// `schema` must point to valid `ArrowArray` and `ArrowSchema` structs
+/// filled by the caller. Ownership is transferred — the caller must not
+/// release the structs after this call.
+///
+/// # Safety
+/// `tw` must be a valid pointer from `paimon_write_builder_new_write`, or null (returns error).
+/// `array` and `schema` must be valid pointers to initialized ArrowArray /
+/// ArrowSchema structs, or null (returns error).
+#[no_mangle]
+pub unsafe extern "C" fn paimon_table_write_write_arrow_batch(
+    tw: *mut paimon_table_write,
+    array: *mut c_void,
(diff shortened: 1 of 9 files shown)


bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+    if let Err(e) = check_non_null(array, "array") {
+        return e;
+    }
+    if let Err(e) = check_non_null(schema, "schema") {
+        return e;
+    }
+
+    let table_write = &mut *((*tw).inner as *mut TableWrite);
+
+    let ffi_array = ptr::read(array as *const FFI_ArrowArray);
```
QuakeWang: `ptr::read` does not clear the caller's `release` callback. The imported copy owns the same resources while the original struct still appears valid, which can result in a double-free.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+    let ffi_schema = ptr::read(schema as *const FFI_ArrowSchema);
+
+    let batch = match unsafe { from_ffi(ffi_array, &ffi_schema) } {
```
QuakeWang: Non-Struct or nullable root arrays can panic here. The imported schema is also not checked against the table schema, so an incompatible schema may be written directly into data files.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+) -> paimon_result_prepare_commit {
+    if let Err(e) = check_non_null(tw, "tw") {
+        return paimon_result_prepare_commit {
+            messages: ptr::null_mut(),
+            error: e,
+        };
+    }
+    let table_write = &mut *((*tw).inner as *mut TableWrite);
+
+    match runtime().block_on(table_write.prepare_commit()) {
```
QuakeWang: The messages do not retain their source table or `commit_user`. Messages produced for one table can therefore be committed to another table, creating a snapshot that references files under the wrong location.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+        return e;
+    }
+    if let Err(e) = check_non_null(msgs, "msgs") {
+        return e;
+    }
+
+    let table_commit = &*((*tc).inner as *const TableCommit);
+    let messages = Box::from_raw((*msgs).inner as *mut Vec<CommitMessage>);
+    // Release the outer wrapper to avoid double-free when caller frees msgs
+    drop(Box::from_raw(msgs));
```
QuakeWang: This destroys the messages handle even though the API documentation requires callers to free it explicitly. Freeing it after commit becomes a use-after-free, and a failed commit cannot be retried or aborted.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+
+    let table_commit = &*((*tc).inner as *const TableCommit);
+    let messages = Box::from_raw((*msgs).inner as *mut Vec<CommitMessage>);
+    drop(Box::from_raw(msgs));
+
+    if messages.is_empty() {
+        return ptr::null_mut();
+    }
+
+    match runtime().block_on(table_commit.abort(&messages)) {
```
QuakeWang: `prepare_commit` can produce `new_index_files`, but the current abort implementation only deletes data and changelog files. Dynamic-bucket aborts leave index files behind.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,506 @@
...
+    table: *const paimon_table,
+) -> paimon_result_write_builder {
+    if let Err(e) = check_non_null(table, "table") {
+        return paimon_result_write_builder {
+            write_builder: ptr::null_mut(),
+            error: e,
+        };
+    }
+    let table_ref = &*((*table).inner as *const Table);
+    let wb = table_ref.new_write_builder();
```
QuakeWang: The generated `commit_user` cannot be supplied by the caller, and the API does not accept a `commit_identifier`. Separate writers cannot share commit identity, and uncertain commit results cannot be retried idempotently.

suxiaogang223:
Thanks for the detailed review, @QuakeWang! I added regression tests to reproduce each of the issues you pointed out and confirmed the failures before making the fixes. The new tests now cover Arrow C Data ownership, invalid and schema-mismatched batches, commit message provenance and lifetime, index cleanup on abort, and caller-provided commit identity with idempotent retries. Everything is passing now, including the full paimon-c test suite and the targeted core tests. The fixes are in 27a9d90. When you get a chance, could you please take another look? Thanks!

JingsongLi:
No issue from me.

crates/paimon/src/table/table_commit.rs
```diff
@@ -594,7 +599,13 @@ impl TableCommit {
 
         loop {
             let latest_snapshot = self.snapshot_manager.get_latest_snapshot().await?;
-            if let Some(start_snapshot_id) = duplicate_check_start_snapshot_id {
+            // Explicit identifiers are stable commit identities and must also
+            // deduplicate a fresh invocation after an uncertain result. The
+            // batch identifier is intentionally excluded because callers may
+            // perform multiple independent batch commits with the same user.
+            let duplicate_check_start = duplicate_check_start_snapshot_id
+                .or((commit_identifier != BATCH_COMMIT_IDENTIFIER).then_some(1));
```
QuakeWang: Every commit with an explicit identifier scans from snapshot 1 to the latest snapshot, resulting in O(N²) snapshot reads over successive commits. If an older snapshot has expired, retrying its identifier may also commit the data again, breaking idempotency.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,707 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn paimon_table_new_write_builder(
+    table: *const paimon_table,
+) -> paimon_result_write_builder {
+    new_write_builder(table, None)
+}
+
+/// Create a WriteBuilder with a caller-provided stable commit identity.
+///
+/// Distributed writers for one commit should use the same `commit_user`.
```
QuakeWang: This states that distributed writers can share a `commit_user`, but opaque commit messages cannot be merged. If two writers commit separately with the same identifier, the second call succeeds without adding its files to the snapshot, silently dropping data.

bindings/c/src/write.rs
```diff
@@ -0,0 +1,707 @@
...
+    input: &ArrowSchema,
+    target: &ArrowSchema,
+) -> Result<(), *mut paimon_error> {
+    let matches = input.fields().len() == target.fields().len()
+        && input
+            .fields()
+            .iter()
+            .zip(target.fields().iter())
+            .all(|(input, target)| {
+                input.name() == target.name() && input.data_type() == target.data_type()
```
QuakeWang: This ignores the target field's nullability. A nullable Arrow column containing nulls is accepted for a Paimon `NOT NULL` field, allowing data that violates the table schema to be written.

QuakeWang (changes requested):

crates/paimon/src/table/table_commit.rs
```diff
@@ -262,6 +291,7 @@ impl TableCommit {
             },
             None,
             commit_identifier,
+            false,
```
QuakeWang: Only APPEND has a filtered retry path. OVERWRITE and TRUNCATE always disable filtering, so retrying an uncertain result after an intervening commit can re-execute the destructive operation and remove data written in between.

crates/paimon/src/table/table_commit.rs
```diff
@@ -1065,6 +1113,35 @@ impl TableCommit {
...
+        let earliest_snapshot_id = self
+            .snapshot_manager
+            .earliest_snapshot_id()
+            .await?
+            .unwrap_or(latest.id());
+        for snapshot_id in (earliest_snapshot_id..=latest.id()).rev() {
+            let snapshot = if snapshot_id == latest.id() {
+                Some(latest.clone())
+            } else {
+                self.snapshot_manager.get_snapshot(snapshot_id).await.ok()
```
QuakeWang: This suppresses all snapshot read errors, including I/O and deserialization failures. If the latest snapshot for this commit user cannot be read, filtering may fall back to an older identifier and commit the same data again.

JingsongLi:
1. The explicit identifier API for `overwrite/truncate` sets `filter_committed=false` by default. It is unclear whether retries following a successful operation might execute destructive operations again, potentially causing the accidental deletion of new data committed in the meantime.
  2. `is_committed_identifier` uses `.await.ok()` when reading historical snapshots, which swallows I/O and deserialization exceptions. This may fail to detect committed identifiers, leading to duplicate commits.

crates/paimon/src/table/table_commit.rs
```diff
@@ -511,12 +579,23 @@ impl TableCommit {
...
+        self.truncate_table_impl(BATCH_COMMIT_IDENTIFIER, false)
             .await
     }
 
     /// Truncate the entire table with a caller-provided commit identifier.
+    ///
+    /// A previously committed identifier is filtered so retrying an uncertain
+    /// result cannot delete data committed in between.
     pub async fn truncate_table_with_identifier(&self, commit_identifier: i64) -> Result<()> {
+        self.truncate_table_impl(commit_identifier, true).await
```
QuakeWang, line 591: When the first identified truncate is a no-op on an already-empty table, `try_commit` exits without writing a snapshot, so the identifier is never recorded. Retrying the same identifier after another writer appends data is not filtered and deletes the intervening data.

JingsongLi:
No issue from me.

QuakeWang (approved):
LGTM

PR #523, opened 2026-07-15
feat(c): expose string and range predicates through the C binding

JunRuiLee:
### Purpose

The C binding (`bindings/c`) exposed only comparison, null, and IN/NOT IN predicates. String-pattern and range operators — already implemented in the core `PredicateBuilder` — were unreachable from C, so C consumers (e.g. engine integrations) could not push those filters down.

### Brief change log

- Add C FFI functions for `starts_with` / `ends_with` / `contains` / `like` / `between` / `not_between`, each with a default (case-sensitive) and a `_with_case_sensitive` variant.
- Reuse the existing `build_leaf_predicate_datum` / `build_leaf_predicate_datums` helpers; `like` maps its escape byte via a small `escape_char_from_c` helper (`0` → core default `\`) and passes it through a closure, so no error-handling path is duplicated. Non-`\` escapes are rejected by core.
- Pin the six new base symbols with compile-time signature assertions, matching the existing convention that guards C-linked predicate symbols against accidental ABI changes.
- Binding-only: no changes to the core crate.

### Tests

- `cargo test -p paimon-c` — added unit tests covering the happy path, a rejected non-`\` escape, a non-string datum error, the `low > high` short-circuit, and a smoke test touching every `_with_case_sensitive` variant.

### API and Format

- Adds new C API symbols; no change to existing symbols or to any storage format.

### Documentation

- Doc comments on each new function; no separate docs needed.

### DIFF
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -641,6 +641,20 @@ unsafe fn datum_from_c(d: &paimon_datum) -> Result<Datum, *mut paimon_error> {
     }
 }
 
+/// Map a C escape byte to the core's `Option<char>` escape parameter.
+///
+/// `0` (NUL) means "use the core default escape" (`None`; the core resolves it
+/// to `\`). Any other byte is the escape character. `c_char` is signed on the
+/// target platforms, so the byte is taken through `u8` before `char` to avoid
+/// sign extension.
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #524, opened 2026-07-15
feat(c): expose read builder scan options and time-travel through the C binding

JunRuiLee:
### Purpose

The C binding's `paimon_table_new_read_builder` took only the table — there was no channel to pass a per-read scan selector, so time-travel reads (by snapshot-id / tag / timestamp / version) were unreachable from C.

### Brief change log

- Add `paimon_table_new_read_builder_with_options(table, options, options_len)` taking a `paimon_option` array. The existing `paimon_table_new_read_builder` is kept and now delegates to it with an empty map (one shared implementation).
- Resolve time-travel at builder construction: reject more than one time-travel selector up front; run the core's `copy_with_time_travel`; a selector that is set but does not resolve to a snapshot is an error — never a silent read-of-latest. Unsupported scan options (`scan.watermark`, incremental) surface as the core's typed error.
- Pin both read-builder symbols with compile-time ABI signature guards.
- Binding-only: no changes to the core crate.

### Tests

- `cargo test -p paimon-c` — empty options build like the plain builder, a non-selector option builds, >1 selector is rejected, an unsupported scan option is rejected, a malformed selector value fails closed (asserting the unified "did not resolve" error), and a null-pointer-with-length guard. Error-path tests assert the error identity, not just the shape.

### API and Format

- Adds one new C API symbol; no change to existing symbols or to any storage format.

### Documentation

- Doc comments on the new function.

### DIFF
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -1723,6 +1723,18 @@ const _: unsafe extern "C" fn(
     paimon_datum,
 ) -> paimon_result_predicate = paimon_predicate_not_between;
 
+// Read builder ABI signature guards. These pin the C-linked read-builder
+// constructors so an accidental signature change fails to compile rather than
+// silently breaking header consumers. To add behavior, introduce a new
+// `paimon_table_new_read_builder_*` symbol instead of changing one of these.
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #528, opened 2026-07-16
fix(catalog): handle filesystem catalog existence on object stores

### LINKED ISSUE #527: Filesystem catalog misdetects databases and tables on object stores
### Bug

The filesystem catalog checks database and table paths with `FileIO::exists()`. On S3-compatible storage this becomes a HEAD request for the exact key, but an existing Paimon warehouse often contains only objects below that prefix and no directory marker.

This can produce inconsistent results:

- `list_databases()` finds a database prefix, while `get_database()` or `list_tables()` returns `DatabaseNotExist`.
- A table with valid schema files can be listed but `get_table()` returns `TableNotExist`.
- A directory without any schema file can be exposed by `list_tables()` as if it were a table.

### Reproduction

Create or copy a Paimon warehouse to an S3-compatible store without directory marker objects, then open it with `FileSystemCatalog`. A layout such as:

```text
warehouse/db.db/table/schema/schema-0
warehouse/db.db/table/data/data-file
```

has no objects named `warehouse/db.db/` or `warehouse/db.db/table/`. Listing sees the prefixes, but an exact existence check does not.

### Expected behavior

Database paths should use directory semantics so markerless prefixes are recognized. A table should exist only when it has a valid `schema-*` file, and `list_tables()` should filter incomplete directories. This matches the Java filesystem catalog behavior.

### Proposed fix

Add a directory-aware existence check in `FileIO`, use it for databases, and use schema files to determine table existence.

I am willing to submit a PR.

Pandas886:
### Purpose

Linked issue: close #527

The filesystem catalog could discover a database or table while listing an object-store prefix, then reject the same entry when checking the exact path. Existing Paimon warehouses often have objects below a prefix without a directory marker, so an exact HEAD request is not enough to determine whether a database exists.

Table existence also needs to be based on schema files rather than the table directory itself. This is the behavior used by the Java filesystem catalog.

### Brief change log

- Add a directory-aware existence check to `FileIO`.
- Use directory semantics when checking filesystem catalog databases.
- Check `schema-0` first, then fall back to other schema IDs when checking tables.
- Filter directories without a valid schema from `list_tables()`.
- Cover empty directories, markerless prefixes, non-zero schema IDs, and incomplete table directories.

### Tests

- `cargo check -p paimon --features storage-s3`
- Tested against a local MinIO container with an empty database, a markerless database and table, a table starting at `schema-1`, and an incomplete table directory.

### API and Format

This adds `FileIO::exists_dir()`. There is no storage format change.

### Documentation

No documentation changes are needed.

### DIFF
--- a/crates/paimon/src/catalog/filesystem.rs
+++ b/crates/paimon/src/catalog/filesystem.rs
@@ -190,12 +190,19 @@ impl FileSystemCatalog {
 
     /// Check if a database exists.
     async fn database_exists(&self, name: &str) -> Result<bool> {
-        self.file_io.exists(&self.database_path(name)).await
+        self.file_io.exists_dir(&self.database_path(name)).await
     }
 
     /// Check if a table exists.
     async fn table_exists(&self, identifier: &Identifier) -> Result<bool> {
-        self.file_io.exists(&self.table_path(identifier)).await
+        let table_path = self.table_path(identifier);
+        let schema_manager = SchemaManager::new(self.file_io.clone(), table_path);
+
+        if self.file_io.exists(&schema_manager.schema_path(0)).await? {
+            return Ok(true);
+        }
+
+        Ok(!schema_manager.list_all_ids().await?.is_empty())
     }
 }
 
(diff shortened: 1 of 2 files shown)


JingsongLi:
PR changes “table existence” to “schema must exist,” and `list_tables()` will also filter out orphaned directories without a schema; however, `drop_database(cascade=false)` still checks the original subdirectories directly. As a result, the Catalog shows that the database has no tables, but it returns `DatabaseNotEmpty` due to an incomplete or orphaned directory. The Java implementation uses `listTables(name)` for this check; the semantics should be standardized.

Pandas886:
Good catch. I updated `drop_database()` to use `list_tables(name)` for the non-cascade emptiness check, matching the Java catalog semantics. I also added a regression test covering a database that contains only an incomplete table directory: it is not listed as a table and the database can be dropped without cascade.

JingsongLi (approved):
+1

PR #529, opened 2026-07-16
fix(datafusion): remove unneeded wildcard pattern for clippy 1.97.0

hechao-ustc:
### Purpose

Fix the `unneeded_wildcard_pattern` clippy warning introduced in Rust 1.97.0. 

### Brief change log

Remove the redundant `if_exists: _` field in the `AlterTableOperation::DropColumn` match arm, as it is already covered by the `..` pattern

### Tests

No new tests needed — behavior is unchanged.

### API and Format

No changes.

### Documentation

No documentation update needed.


QuakeWang (approved):
+1

PR #531, opened 2026-07-16
[core] Align primary-key vector source metadata with Java

JunRuiLee:
## Summary
- parse Java's current primary-key vector source metadata frame with dataLevel
- filter current ANN segments by exact level source coverage, matching Java PkVectorBucketIndexState
- update PK-vector source-meta fixtures and tests

## Motivation
This follows the Java-side PK-vector source metadata changes in apache/paimon#8672.

Java now writes `dataLevel` into `GlobalIndexMeta.sourceMeta` between the frame version and source-file count. Without this update, Rust reads `dataLevel` as the source-file count and cannot correctly consume Java-written PK-vector metadata.

Java also selects current ANN payloads by exact level coverage: active compacted files are grouped by data level, each segment's `sourceMeta.dataLevel()` must match one level, and the segment's source files must exactly equal that level's active files. Mirroring that rule keeps Rust from searching stale or partially-overlapping ANN segments after compaction rewrites files.

## Tests
- cargo fmt --check
- cargo test -p paimon pk_vector_source --lib
- cargo test -p paimon pk_vector_scan --lib
- cargo test -p paimon vindex::pkvector::bucket --lib
- cargo test -p paimon vindex::pkvector::ann --lib
- cargo test -p paimon pk_vector_orchestrator --lib
- cargo test -p paimon --test pk_vector_baseline_test
- cargo clippy -p paimon --all-targets -- -D warnings

### DIFF
--- a/crates/paimon/src/spec/pk_vector_source.rs
+++ b/crates/paimon/src/spec/pk_vector_source.rs
@@ -76,15 +76,28 @@ impl PkVectorSourceFile {
 /// Mirrors Java `org.apache.paimon.index.pkvector.PkVectorSourceMeta`.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PkVectorSourceMeta {
+    data_level: i32,
     source_files: Vec<PkVectorSourceFile>,
 }
 
 impl PkVectorSourceMeta {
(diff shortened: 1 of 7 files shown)


JingsongLi (approved):
+1

PR #532, opened 2026-07-16
fix(catalog): align FileSystemCatalog with Java for object store comp…

mlboy:
…atibility

- database_exists: append trailing slash so OpenDAL correctly identifies directory paths on OSS/S3 (Hadoop does this via getFileStatus fallback)
- table_exists: check schema file presence instead of bare directory existence, mirroring Java's AbstractCatalog.tableExistsInFileSystem
- list_tables: filter directories by schema validity and sort results, matching Java's listTablesInFileSystem behavior

### Purpose

Linked issue: close #xxx


crates/paimon/src/catalog/filesystem.rs
```diff
@@ -189,13 +189,47 @@ impl FileSystemCatalog {
...
+    /// the list operation may error; this is treated as "table does not exist".
+    async fn table_exists_in_filesystem(&self, table_path: &str) -> Result<bool> {
+        let schema_0_path = self.schema_file_path(table_path, 0);
+        if self.file_io.exists(&schema_0_path).await? {
+            return Ok(true);
+        }
+        let manager = SchemaManager::new(self.file_io.clone(), table_path.to_string());
+        match manager.list_all_ids().await {
+            Ok(ids) => Ok(!ids.is_empty()),
+            Err(_) => Ok(false),
```
QuakeWang: This turns every `list_all_ids` failure, including permission and transient storage errors, into “table does not exist.” If `schema-0` is absent but later schema files exist, `create_table` can treat the existing table as new and write a new `schema-0`. I reproduced this with a writable but non-listable schema directory: the call returned `Ok(())` and created `schema-0`.

QuakeWang (changes requested):

crates/paimon/src/catalog/filesystem.rs
mlboy: Sorry about that, I missed this edge case. It's been fixed now — only NotFound is treated as non-existence, other errors are properly propagated.

crates/paimon/src/catalog/filesystem.rs
```diff
@@ -189,13 +189,53 @@ impl FileSystemCatalog {
...
+    /// Verify a table path contains valid schema metadata.
+    ///
+    /// Mirrors Java's `AbstractCatalog.tableExistsInFileSystem`:
+    /// 1. Fast path — check if `schema/schema-0` exists.
+    /// 2. Slow path — list schema directory for any `schema-{N}` file.
+    ///
+    /// Only a `NotFound` error from the schema directory is treated as
+    /// "table does not exist." All other errors (permission, network, etc.)
+    /// are propagated to avoid masking real failures.
+    async fn table_exists_in_filesystem(&self, table_path: &str) -> Result<bool> {
```
QuakeWang, line 222: The behavior introduced here has no regression coverage. Existing tests use normal local directories and pass before this change, so markerless prefixes, schema fallback, invalid-directory filtering, and non-`NotFound` error propagation remain unverified.

mlboy: Would you mind reviewing this one more time? thanks

JingsongLi:
CI failed

JingsongLi:
CI failed

PR #534, opened 2026-07-17
perf(blob): parallelize descriptor range reads

JingsongLi:
## What changed

- Resolve `BlobDescriptor` columns and URI range groups concurrently.
- Share a reader-lifetime limiter across metadata and range I/O.
- Cap active I/O at 8 requests and range admission at 64 MiB.
- Preserve row and column order, with deterministic concurrency tests.

## Why

External `BlobDescriptor` materialization awaited descriptor columns, URI groups, and merged ranges serially. On high-latency object storage, those round trips accumulated across a batch. This change adds bounded asynchronous concurrency while retaining the existing per-URI range coalescing.

## Impact

This improves latency and throughput for batches containing multiple non-coalescible external descriptors while bounding request pressure and admitted range bytes. The ordinary `.blob` reader and public configuration remain unchanged.

## Validation

- `cargo test -p paimon`
- `git diff --check`


leaves12138 (approved):
The latest update resolves the Clippy failure. I rechecked the bounded parallel descriptor-read implementation and the new concurrency coverage, and found no correctness issues. The targeted overlap test and formatting checks pass locally.

PR #536, opened 2026-07-17
[core] Support MAP shared-shredding in Parquet

JingsongLi:
## What changed

- add Java-compatible MAP shared-shredding write and read plans for `MAP<STRING, T>`
- infer the number of shared physical columns from the first buffered row
- persist and restore the field dictionary, column allocation, overflow set, and layout metadata in Parquet
- integrate MAP and Variant shredding through shared read/write plan abstractions
- assemble logical MAP values before residual predicate evaluation
- support `none`, `lz4`, and `zstd` field-dictionary compression
- normalize nested Arrow value types and gather only mapped values with `interleave`
- reject malformed or oversized shredding metadata before allocation or schema construction

## Why

Paimon Java supports the PIP-43 MAP shared-shredding layout, but the Rust implementation previously supported only Variant shredding. Rust therefore could not write or reconstruct the Java-compatible MAP physical layout.

The read path also needs to account for Java-generated nested Arrow field metadata and untrusted footer dimensions without materializing every shared physical value column.

## Impact

Parquet readers and writers can use:

```
fields.<column>.map.storage-layout=shared-shredding
fields.<column>.map.shared-shredding.max-columns=<n>
```

The implementation preserves null and empty MAP semantics, overflow entries, nested ARRAY/MAP values, and residual-filter ordering. Unsupported simultaneous Variant and MAP shredding plans are rejected.

## Validation

- `cargo clippy -p paimon --lib --tests --no-deps -- -D warnings`
- `cargo test -p paimon --lib` — 1625 passed, 1 ignored

### DIFF
--- a/crates/paimon/src/arrow/shredding/map.rs
+++ b/crates/paimon/src/arrow/shredding/map.rs
@@ -261,18 +261,19 @@ fn build_physical_struct_type(
     value_type: &DataType,
     num_columns: usize,
     include_overflow: bool,
-) -> DataType {
+) -> Result<DataType> {
     let mut fields = Vec::with_capacity(num_columns + 2);
     fields.push(DataField::new(
         0,
         FIELD_MAPPING_NAME.to_string(),
         DataType::Array(ArrayType::new(DataType::Int(IntType::new()))),
     ));
+    let direct_value_type = value_type.copy_with_nullable(true)?;
     for i in 0..num_columns {
         fields.push(DataField::new(
             (i + 1) as i32,
             physical_column_name(i),
-            value_type.clone(),
+            direct_value_type.clone(),
         ));
     }
     if include_overflow {
@@ -285,7 +286,7 @@ fn build_physical_struct_type(
             )),
         ));
     }
-    DataType::Row(RowType::new(fields))
+    Ok(DataType::Row(RowType::new(fields)))
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


crates/paimon/src/arrow/shredding/map.rs
```diff
@@ -0,0 +1,2762 @@
...
+    fields.push(DataField::new(
+        0,
+        FIELD_MAPPING_NAME.to_string(),
+        DataType::Array(ArrayType::new(DataType::Int(IntType::new()))),
+    ));
+    for i in 0..num_columns {
+        fields.push(DataField::new(
+            (i + 1) as i32,
+            physical_column_name(i),
+            value_type.clone(),
```
leaves12138: `value_type.clone()` preserves `NOT NULL` on every direct physical column, but a direct column is absent (and therefore null) whenever a row contains fewer than `K` map entries. For example, `MAP<STRING, BIGINT NOT NULL>` with rows `{a: 10, b: 20}` and `{a: 30}` fails in `to_physical_batch` with `Found unmasked nulls for non-nullable StructArray field "__col_1"`. Please make the direct shared columns nullable (or provide non-null placeholders for unmapped slots) and add a sparse-row roundtrip test for a non-nullable value type.

leaves12138 (changes requested):
I found one correctness blocker in the MAP shared-shredding write path: sparse rows fail for `MAP<STRING, T NOT NULL>` because direct physical columns inherit the non-nullable value type while unmapped slots are represented as nulls. The existing CI and targeted MAP tests pass, but a local reproducer fails in `to_physical_batch`; details are in the inline comment.

crates/paimon/src/arrow/shredding/map.rs
leaves12138: Fixed in `0e3d0eb`. I re-ran the original sparse-row reproduction, the MAP shredding test suite, and an additional end-to-end Parquet roundtrip with `MAP<STRING, BIGINT NOT NULL>`; all pass.

leaves12138 (approved):
The nullable direct-column fix addresses the reported sparse-row failure without weakening the logical MAP value nullability. I re-ran the original reproduction, all MAP shredding unit tests, and an additional end-to-end Parquet roundtrip for `MAP<STRING, BIGINT NOT NULL>`; all pass. CI is also green.

PR #537, opened 2026-07-18
[c] Support vector search materialized read in C FFI bindings

JunRuiLee:
### Purpose

Part of #514. Expose vector search to C/C++/Go consumers (e.g. Doris) via `paimon-c`. Primary-key vector search produces materialized rows (physical `(file, position)` coordinates, no global row ids), so the unified user-facing shape across storage types is **materialized rows + a score column** — matching Java Spark, where both primary-key and data-evolution reads surface rows plus a `__paimon_search_score` metadata column. This PR wires that through: `execute_read()` gains a data-evolution branch, and a single C entry returns the existing Arrow-stream reader.

### Brief change log

- `feat(table)`: `VectorSearchBuilder::execute_read()` now materializes rows for **both** primary-key and data-evolution vector search. Unified user-visible score column `__paimon_search_score` (was PK-only `_PKEY_VECTOR_SCORE`, kept as a rejected projection alias). Data-evolution rows are joined to their scores by global `_ROW_ID`, reordered to the search's best-first rank (not re-sorted by score), then `_ROW_ID` is stripped; any missing / null / wrong-type / unknown id, or a row count that disagrees with the scored result, fails loud (never a NaN score or a silently dropped row). A filter on the data-evolution path fails loud (parity is a follow-up); the primary-key filter path is unchanged.
- `feat(c)`: new `bindings/c/src/vector_search.rs` — a builder handle over owned state, setters (`with_vector_column` / `with_query_vector` / `with_limit` / `with_options` / `with_filter`), and a single terminal `paimon_vector_search_builder_execute_read` returning `paimon_result_record_batch_reader` (reuses the existing Arrow C Data Interface reader; consumers drain it with the existing `paimon_record_batch_reader_next` / `paimon_arrow_batch_free` / `paimon_record_batch_reader_free`). The entry is storage-agnostic — C does no table-type routing; the core decides. Errors surface via the result's `error` field; an empty result is a normal end-of-stream reader. Compile-time ABI signature guards pin every exported symbol.

### Tests

- Core: data-evolution materialization end to end; rank-order fidelity (proven with a score tie); row-id alignment across batches; fail-loud branches (missing / null / wrong-type / unknown `_ROW_ID`, count mismatch); reserved-projection rejection.
- C: 6 integration tests comparing the C Arrow stream against an independent core `execute_read()` reference across primary-key and data-evolution tables; a primary-key filter that excludes a neighbor; the data-evolution filter `InvalidInput` error; setter validation; and the empty end-of-stream case.
- `cargo test -p paimon` and `cargo test -p paimon-c` green; `cargo build -p paimon-datafusion` clean (cross-crate `Send` gate); `cargo clippy -D warnings` and `cargo fmt --check` clean.

### API and Format

No on-disk format change. New C ABI symbols only (add-only, each guarded by a compile-time signature assertion). This replaces the earlier scored (row-ids + scores) C entry, which was never released.

### Documentation

Doc comments on the new FFI functions; no user-facing docs change.

### DIFF
--- a/crates/paimon/src/table/vector_search_builder.rs
+++ b/crates/paimon/src/table/vector_search_builder.rs
@@ -296,6 +296,47 @@ impl<'a> VectorSearchBuilder<'a> {
     /// is always hidden. A filter is unsupported here and fails loud inside
     /// `execute_scored`.
     async fn execute_de_vector_read(&self) -> crate::Result<ArrowRecordBatchStream> {
+        // Validate the target column exists and is a vector-bearing type before any
+        // work. The data-evolution search returns an empty result for an unknown
+        // field (its scored-path behavior), which would make a typo'd or scalar
+        // column look like a normal empty read here — violating `execute_read`'s
+        // fail-loud contract (a C/Doris caller would see EOF, not an input error).
+        // Reject it up front instead.
+        let vector_column =
+            self.vector_column
+                .as_deref()
+                .ok_or_else(|| crate::Error::ConfigInvalid {
+                    message: "Vector column must be set via with_vector_column()".to_string(),
+                })?;
+        let field = self
+            .table
+            .schema()
+            .fields()
+            .iter()
+            .find(|f| f.name() == vector_column)
+            .ok_or_else(|| crate::Error::DataInvalid {
+                message: format!("vector search column '{vector_column}' does not exist"),
+                source: None,
+            })?;
+        // Require a FLOAT-element vector column: `ARRAY<FLOAT>` or `VECTOR<FLOAT>`,
+        // matching the element type the vector index/search operates on. An
+        // `ARRAY<INT>` (or any non-float element) is not a searchable vector column.
+        let is_float_vector = match field.data_type() {
+            DataType::Vector(t) => matches!(t.element_type(), DataType::Float(_)),
+            DataType::Array(t) => matches!(t.element_type(), DataType::Float(_)),
+            _ => false,
+        };
+        if !is_float_vector {
+            return Err(crate::Error::DataInvalid {
+                message: format!(
+                    "vector search column '{vector_column}' must be a FLOAT vector column \
+                     (ARRAY<FLOAT> or VECTOR<FLOAT>), got {:?}",
+                    field.data_type()
+                ),
+                source: None,
+            });
+        }
+
         let sr = self.execute_scored().await?;
 
         // Resolve the projected user columns up front so an invalid projection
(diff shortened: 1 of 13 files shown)


crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -588,8 +635,9 @@ impl<'a> VectorSearchBuilder<'a> {
             Some(names) => {
                 for name in names {
                     if name == PKEY_VECTOR_POSITION_COLUMN
-                        || name == PKEY_VECTOR_SCORE_COLUMN
+                        || name == SEARCH_SCORE_COLUMN
```
leaves12138: This reserved-name validation only runs for `Some(names)`. With the default projection, a valid table column named `__paimon_search_score` is retained, and `attach_scores_by_row_id` then appends another field with the same name. I reproduced an output schema containing two `__paimon_search_score` columns. The legacy `_PKEY_VECTOR_SCORE` alias is likewise rejected only for explicit projections. Please validate the resolved/default field list as well, and add default-projection collision coverage for both PK and DE materialization.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -283,10 +282,58 @@ impl<'a> VectorSearchBuilder<'a> {
...
         }
 
-        Err(crate::Error::DataInvalid {
-            message: "vector search read is only supported for primary-key vector indexes".into(),
-            source: None,
-        })
+        // Data-evolution (global-index) vector search: materialize rows from the
+        // scored global row-ids and attach the unified score column. A non-vector
+        // column or a set filter fails loud inside execute_scored below.
+        self.execute_de_vector_read().await
```
leaves12138, line 288: This fallback can turn an invalid vector-column name into a successful empty reader. `evaluate_batch_vector_search` returns an empty result when field lookup fails, and the new test explicitly expects `with_vector_column("other")` on the empty table to succeed. Through the C API, a typo therefore looks like normal EOF instead of an input error, contrary to the `execute_read` fail-loud contract. Please validate that the DE target field exists and has a supported vector type before returning an empty result, with core/C coverage for unknown and scalar columns.

leaves12138 (changes requested):
The focused core and C FFI tests pass locally, but I found two correctness/API issues in the new data-evolution materialized-read path: default projection can produce duplicate reserved score columns, and invalid vector-column names can silently return an empty reader. Please address these before merge.

crates/paimon/src/table/vector_search_builder.rs
JunRuiLee: Good catch — fixed in e3dbea0. `resolve_materialize_read_type` now runs the reserved-name check on the **resolved** field list (both the explicit-projection and default cases), so a user column literally named `__paimon_search_score` / `_PKEY_VECTOR_SCORE` is rejected before it can collide with the appended score column. The raw-request check is kept for the explicit-projection path so a requested reserved name still gets the clear "reserved column" error rather than a confusing "not found" from field resolution. Added default-projection collision coverage for both the primary-key and data-evolution materialization paths.

crates/paimon/src/table/vector_search_builder.rs
JunRuiLee: Fixed in e3dbea0. `execute_read` on the data-evolution path now validates up front that the target column exists and is a FLOAT vector column (`ARRAY<FLOAT>` or `VECTOR<FLOAT>`) before doing any search, so an unknown, scalar, or non-float (e.g. `ARRAY<INT>`) column fails loud with `InvalidInput` instead of surfacing as an empty EOF reader through the C API. `execute_scored`'s existing empty-on-unknown-field behavior is left unchanged. Added core coverage (unknown / scalar / non-float-array) and C FFI coverage (unknown and scalar columns mapping to `InvalidInput`).

leaves12138 (approved):
Re-reviewed the follow-up fixes. Reserved-column validation now covers the resolved default projection, invalid data-evolution vector targets fail loudly with appropriate input errors, and the focused core/C FFI regression tests pass locally. Looks good to me.

PR #538, opened 2026-07-18
docs(python): add Python Integration documentation page

hechao-ustc:
### Purpose

Add a new Python Integration page to the Paimon Rust documentation, following the same format as the existing Go Integration page.

### Brief change log

- Add docs/src/python-binding.md — Python Integration documentation                                                                 
- Add Python Integration entry to docs/mkdocs.yml navigation   

### Tests

N/A — documentation-only change.    

### API and Format

N/A — no API or format changes.  

### Documentation

This PR is the documentation. The new page covers all Python binding APIs available at bindings/python/:                            

  - Catalog creation (filesystem, OSS, REST)
  - SQL Context (SQLContext) with DDL/DML support
  - Reading tables (scan-then-read via ReadBuilder)
  - Writing tables (write-then-commit via WriteBuilder)
  - Column projection, limit, case sensitivity
  - Filter push-down (lightweight dict predicate format, compound predicates, literal type mapping)
  - Time travel (snapshot ID, timestamp, version, tag)
  - Table inspection (snapshots, tags, partition stats)
  - Python UDF registration
  - Complete end-to-end example

### DIFF
--- a/docs/src/python-binding.md
+++ b/docs/src/python-binding.md
@@ -1,0 +1,441 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Python Integration
+
+The Python integration is a binding built on top of Apache Paimon Rust, allowing you to access Paimon tables from Python programs. It uses [PyArrow](https://arrow.apache.org/docs/python/) for zero-copy data transfer via the [Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html).
+
+## Prerequisites
+
+- Python 3.10 or later
+- Supported platforms: Linux (amd64, arm64), macOS (amd64, arm64)
+
+## Installation
+
+```bash
+pip install pypaimon-rust
+```
+
+The pre-built native library is embedded in the package and automatically loaded at runtime — no manual build step is needed.
+
+## Creating a Catalog
+
+Use `PaimonCatalog` with a dictionary of options to create a catalog. The catalog type is determined by the `metastore` option (default: `filesystem`).
+
+```python
+from pypaimon_rust.datafusion import PaimonCatalog
+
+# Local filesystem
+catalog = PaimonCatalog({"warehouse": "/path/to/warehouse"})
+
+# List databases and tables
+print(catalog.list_databases())
+print(catalog.list_tables("default"))
+
+# Get a table handle
+table = catalog.get_table("default.my_table")
+```
+
+### Alibaba Cloud OSS
+
+```python
+catalog = PaimonCatalog({
+    "warehouse": "oss://bucket/warehouse",
+    "fs.oss.accessKeyId": "your-access-key-id",
+    "fs.oss.accessKeySecret": "your-access-key-secret",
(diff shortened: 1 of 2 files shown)


docs/src/python-binding.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Prerequisites
+
+- Python 3.10 or later
+- Supported platforms: Linux (amd64, arm64), macOS (amd64, arm64)
+
+## Installation
+
+```bash
+pip install pypaimon-rust
```
QuakeWang: `pypaimon-rust` does not declare `pyarrow` as a runtime dependency. In a clean environment, following this installation command causes subsequent queries to fail with `ModuleNotFoundError: No module named pyarrow`.

docs/src/python-binding.md
```diff
@@ -0,0 +1,441 @@
...
+```python
+rb = table.new_read_builder()
+rb.with_projection(["id", "name"])
+
+# Continue with scan-then-read as above...
+```
+
+## Limit
+
+Use `with_limit` to cap the number of rows returned. A limit of `0` returns zero rows.
```
QuakeWang: `with_limit` is only a scan-planning hint and does not guarantee that at most the specified number of rows is returned. For example, `with_limit(1)` still returns all 3 rows when they belong to a single split.

docs/src/python-binding.md
```diff
@@ -0,0 +1,441 @@
...
+
+ctx = SQLContext()
+
+def add_one(args):
+    # args is a tuple of PyArrow Arrays
+    arr = args[0]
+    return pa.array([v.as_py() + 1 for v in arr], type=pa.int32())
+
+ctx.register_udf(
+    ctx.udf(add_one, [pa.int32()], pa.int32(), "immutable", name="add_one")
```
QuakeWang: This UDF example does not run: `SQLContext` has no `udf` method, and a single-argument UDF receives a PyArrow Array rather than a tuple. Consequently, `args[0]` produces a Scalar and triggers a runtime error.

docs/src/python-binding.md
```diff
@@ -0,0 +1,441 @@
...
+Use `with_limit` to cap the number of rows returned. A limit of `0` returns zero rows.
+
+```python
+rb = table.new_read_builder()
+rb.with_limit(100)
+```
+
+## Case Sensitivity
+
+Use `with_case_sensitive` to control whether column-name matching in projections and predicates is case-sensitive. Defaults to `True` (exact match). Set to `False` for case-insensitive matching (ASCII case-folding).
```
QuakeWang, line 210: The filter predicate is constructed using the current case-sensitivity setting when `with_filter` is called. Calling `with_case_sensitive(False)` afterward does not affect an existing predicate, but this ordering constraint is not documented.

hechao-ustc:
@QuakeWang All changes have been made as requested, PTAL

docs/src/python-binding.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Filter Push-Down
+
+Filter push-down prunes data at two levels:
+
+1. **Scan planning** — skips partitions, buckets, and data files based on file-level statistics (min/max).
+2. **Read-side** — applies row-level filtering via Parquet native row filters for leaf predicates.
+
+!!! warning
+    Filter push-down is a **best-effort** optimization. The returned results may still contain rows that do not satisfy the filter condition. Callers should always apply residual filtering on the returned records to ensure correctness.
```
QuakeWang: The filter semantics described here do not match the current implementation. Scan pruning can be conservative, but the read path applies exact residual filtering and does not return rows that fail the predicate.

docs/src/python-binding.md
```diff
@@ -0,0 +1,455 @@
...
+-->
+
+# Python Integration
+
+The Python integration is a binding built on top of Apache Paimon Rust, allowing you to access Paimon tables from Python programs. It uses [PyArrow](https://arrow.apache.org/docs/python/) for zero-copy data transfer via the [Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html).
+
+## Prerequisites
+
+- Python 3.10 or later
+- Supported platforms: Linux (amd64, arm64), macOS (amd64, arm64)
```
QuakeWang: The supported-platform list does not match the published artifacts: the release workflow builds a Windows wheel, and a win_amd64 wheel is available on PyPI.

hechao-ustc:
@QuakeWang Addressed comments, PTAL.

JingsongLi (approved):
+1

PR #539, opened 2026-07-18
perf(blob): finalize blob bunch metadata once

QuakeWang:
### Purpose

`BlobBunch::add` previously rebuilt and merged all logical ranges for every file. Fallback reads then regrouped, resorted, and cloned the same metadata, causing unnecessary planning overhead as rolling blob files accumulated.

### Brief change log

  - Finalize each blob bunch once after all files are collected.
  - Cache logical ranges and sequence group boundaries.
  - Reuse finalized groups during fallback reads to avoid repeated merges, sorting, and metadata clones.

### Tests


JingsongLi (approved):
+1

PR #540, opened 2026-07-18
fix(core): resolve storage config aliases to canonical keys

liwuhen:
### Purpose

`normalize_storage_config` currently applies mirrored key pairs from a single
snapshot of the normalized properties. This cannot resolve aliases that require
more than one mirror step.

For Azure, underscore aliases are declared through a dotted intermediate key:

- `azure.account_name` → `azure.account.name`
- `azure.account.name` → `azure.account-name`

However, `AzdlsConfig` reads the hyphenated canonical key
`azure.account-name`. Because newly mirrored keys are not processed again,
`azure.account_name` is silently ignored. The same issue affects the other
declared Azure underscore aliases.

This change replaces symmetric mirroring with direct, ordered
`alias → canonical` mappings. It also defines deterministic precedence when
multiple forms are supplied:

1. An explicitly supplied canonical key
2. The first matching alias in declaration order

For Azure, dotted aliases are declared before underscore aliases, resulting in:

`canonical > dotted alias > underscore alias`

### Brief change log

- Change `normalize_storage_config` from symmetric one-pass mirroring to direct
  `alias → canonical` resolution.
- Preserve an explicitly supplied canonical value.
- Resolve multiple aliases in declaration order.
- Map every supported Azure dotted and underscore alias directly to the
  hyphenated key read by `AzdlsConfig`:
  - `account_name`
  - `account_key`
  - `sas_token`
  - `client_id`
  - `client_secret`
  - `tenant_id`
  - `authority_host`
- Update the S3, COS, OBS, and GCS alias tables to follow the same directional
  mapping contract.

### Tests

- Unit test `test_azdls_config_parse_underscore_aliases`
  (`storage_azdls.rs`): data-driven coverage for every declared Azure
  underscore alias and verification that each value reaches the corresponding
  `AzdlsConfig` field. **Fails on the pre-fix code.**
- Unit test `test_azdls_config_alias_priority` (`storage_azdls.rs`): verifies
  canonical keys take precedence over aliases and dotted aliases take
  precedence over underscore aliases.
- Existing storage configuration tests verify the directional alias conversion
  does not regress S3, COS, OBS, or GCS configuration parsing.
- Commands run locally:
  - `CARGO_INCREMENTAL=0 cargo test -p paimon --features storage-azdls --lib storage_azdls`
  - `CARGO_INCREMENTAL=0 cargo test -p paimon --no-default-features --features storage-s3,storage-cos,storage-azdls,storage-obs,storage-gcs --lib io::storage_`
  - `git diff --check`

### API and Format

Supported external storage configuration aliases now resolve directly to the
canonical keys consumed by each backend. When multiple forms of the same
setting are supplied, precedence is deterministic: the canonical key wins,
followed by aliases in declaration order.

### Documentation

### DIFF
--- a/crates/paimon/src/io/storage_azdls.rs
+++ b/crates/paimon/src/io/storage_azdls.rs
@@ -32,21 +32,23 @@ const AZURE_ACCOUNT_KEY: &str = "azure.account-key";
 const AZURE_SAS_TOKEN: &str = "azure.sas-token";
 
 const CONFIG_PREFIXES: &[&str] = &["fs.azure.", "fs.abfs.", "abfs.", "abfss.", "azure."];
-const MIRRORED_KEYS: &[(&str, &str)] = &[
-    ("azure.account-name", "azure.account.name"),
-    ("azure.account_name", "azure.account.name"),
-    ("azure.account-key", "azure.account.key"),
-    ("azure.account_key", "azure.account.key"),
-    ("azure.sas-token", "azure.sas.token"),
-    ("azure.sas_token", "azure.sas.token"),
-    ("azure.client-id", "azure.client.id"),
-    ("azure.client_id", "azure.client.id"),
-    ("azure.client-secret", "azure.client.secret"),
-    ("azure.client_secret", "azure.client.secret"),
-    ("azure.tenant-id", "azure.tenant.id"),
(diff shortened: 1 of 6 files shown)


QuakeWang (approved):
LGTM

PR #541, opened 2026-07-18
fix(index): deduplicate vector search top-k results

QuakeWang:
### Purpose

Vector search merged results by concatenating row IDs. When index sources overlap, duplicate rows could consume multiple top-k slots and return fewer unique rows than requested.

### Brief change log

  - Deduplicate results by row ID before top-k selection.
  - Keep the highest score for duplicate rows.
  - Preserve existing ordering and tie-breaking behavior.
  - Add regression tests for overlapping results.

### Tests


JingsongLi (approved):
+1

PR #545, opened 2026-07-19
fix(row): guard timestamp conversion against overflow from corrupt data

hechao-ustc:
### Purpose

`read_timestamp_value` performs unchecked multiplication `(* 1_000 / * 1_000_000)` on the millis value read from `.row` files when converting to microsecond or nanosecond precision. A corrupt or malicious file can supply an arbitrary i64 (e.g. `i64::MAX`), causing a panic in debug builds or silent wrapping in release builds. This PR replaces bare arithmetic with checked_mul + checked_add, returning `Error::DataInvalid` on overflow.

### Brief change log
- Microsecond and Nanosecond arms of read_timestamp_value: replaced bare operators with `checked_mul` + `checked_add` + `ok_or_else`,  returning `DataInvalid` with millis and nanos context on overflow 

### Tests

### DIFF
--- a/crates/paimon/src/arrow/format/row.rs
+++ b/crates/paimon/src/arrow/format/row.rs
@@ -1777,11 +1777,27 @@ fn read_timestamp_value(input: &mut BlockInput<'_>, unit: TimeUnit) -> crate::Re
         TimeUnit::Millisecond => millis,
         TimeUnit::Microsecond => {
             let nanos = i64::from(input.read_var_u32()?);
-            millis * 1_000 + nanos / 1_000
+            millis
+                .checked_mul(1_000)
+                .and_then(|v| v.checked_add(nanos / 1_000))
+                .ok_or_else(|| Error::DataInvalid {
+                    message: format!(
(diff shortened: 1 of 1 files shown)


crates/paimon/src/arrow/format/row.rs
```diff
@@ -1777,11 +1777,27 @@ fn read_timestamp_value(input: &mut BlockInput<'_>, unit: TimeUnit) -> crate::Re
         TimeUnit::Millisecond => millis,
         TimeUnit::Microsecond => {
             let nanos = i64::from(input.read_var_u32()?);
-            millis * 1_000 + nanos / 1_000
+            millis
+                .checked_mul(1_000)
```
QuakeWang: The separate `checked_mul` incorrectly rejects valid timestamps at the negative boundary. The existing writer can encode `i64::MIN`, and the final reconstructed value still fits in `i64`, but this returns `DataInvalid` before adding the nanos component, making valid `.row` data produced by the writer unreadable.

QuakeWang (approved):
+1

PR #546, opened 2026-07-19
feat(python): add BETWEEN/NOT BETWEEN predicate support

hechao-ustc:
### Purpose

The Python dict-format predicate parser was missing `BETWEEN` and `NOT BETWEEN` operators, which are standard SQL comparison operators already supported in the C bindings (`paimon_predicate_between` / `paimon_predicate_not_between`) and the core `PredicateBuilder`. This PR adds them to close the gap.

### Brief change log

- Add `between` and `notBetween` match arms in `leaf_to_predicate()`, each requiring exactly 2 literals (low, high), delegating to `PredicateBuilder::between` / `not_between`
- Add 5 unit tests covering successful conversion (int and string fields), wrong literal count rejection, and literal value
  verification

### Tests

- `cargo test -p pypaimon_rust -- predicate`


QuakeWang (approved):
+1

PR #548, opened 2026-07-19
[datafusion] Align Paimon scan statistics with Parquet

JingsongLi:
## Summary

Align Paimon scan statistics with DataFusion native Parquet semantics. This exposes conservative manifest-derived column statistics while avoiding compressed full-file bytes that do not represent projected scan bytes.

## Changes

- decode per-file value statistics using the current schema and dense column mappings
- aggregate manifest min/max bounds and publish exact null counts only for row-preserving scans
- fail open on malformed, inconsistent, schema-mismatched, or ordering-incompatible statistics
- suppress unsafe statistics for aggregation merge engines, filters, limits, row ranges, and deletion files
- report `total_byte_size` as unknown because Paimon only stores compressed full-file sizes
- add regression coverage for aggregation, unsafe null-count paths, malformed rows, and binary ordering

## Testing

- `cargo test -p paimon value_stats_for_field`
- `cargo test -p paimon-datafusion --lib test_partition_statistics`
- `cargo test -p paimon-datafusion --lib test_binary_manifest_bounds`
- `git diff --check`

Runtime filter support remains a separate follow-up.

### DIFF
--- a/crates/integrations/datafusion/src/physical_plan/scan.rs
+++ b/crates/integrations/datafusion/src/physical_plan/scan.rs
@@ -18,10 +18,12 @@
 use std::sync::Arc;
 
 use datafusion::arrow::compute::cast;
-use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
+use datafusion::arrow::datatypes::{
+    DataType as ArrowDataType, SchemaRef as ArrowSchemaRef, TimeUnit,
+};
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
Looks good to me. The manifest-derived bounds are conservative, exact null counts are restricted to row-preserving scans, and malformed or schema-mismatched statistics safely fall back to unknown.

PR #549, opened 2026-07-19
[datafusion] Support runtime filters for Paimon scans

JingsongLi:
## Summary

Support DataFusion runtime-filter pushdown for Paimon scans while separating conservative pruning from exact row filtering.

This PR contains the complete runtime-filter work: Core reader support, DataFusion physical runtime-filter handling, and the TPC-DS benchmark coverage used to compare Paimon with DataFusion's Parquet reader.

## Changes

- use one predicate channel throughout Core readers and add a boolean that controls whether predicates may remove individual rows
- always make supported predicates available for conservative file, row-group, stripe, and page pruning
- apply Parquet Arrow row filters and format residual filters only when row filtering is enabled
- keep predicates attached when row filtering is disabled, while limiting pre-merge pruning to semantics-safe paths: primary-key conjuncts for key-value merges and raw-convertible data-evolution files
- reject `_ROW_ID` projection whenever a predicate could skip physical positions
- retain only DataFusion physical filters that can be evaluated against the scan output schema
- add the independent DataFusion session option `paimon.read.row_filter`, defaulting to `false`
- return `PushedDown::No` by default so the parent DataFusion operator performs exact filtering; return `PushedDown::Yes` and filter inside the scan only when `paimon.read.row_filter=true`
- preserve the existing `TableProvider` contract: pure partition predicates are `Exact`, while supported data predicates are `Inexact`
- translate completed column/literal comparisons and `IN` lists into Paimon predicates and combine them with the existing pushed predicate
- use a cost-aware, bounded wait for dynamic filters on larger scans without delaying small scans
- downgrade row-count and null-count precision when runtime filters are present
- keep `--parquet-pushdown-filters` scoped to DataFusion's Parquet reader in the TPC-DS-derived benchmark; Paimon keeps row filtering disabled while still using predicates for pruning

## Testing

- `cargo test -p paimon-datafusion --lib config::tests`
- `cargo test -p paimon-datafusion --lib paimon_read_option`
- `cargo test -p paimon-datafusion --lib physical_plan::scan::tests`
- `cargo test -p paimon-datafusion --test blob_tests`
- `cargo test -p paimon-tpcds-bench`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `git diff --check`

The Core regression verifies that a predicate can skip a Parquet row group without removing non-matching rows inside a retained row group when row filtering is disabled. DataFusion tests cover schema compatibility, completed and incomplete dynamic filters, bounded waiting, comparison and `IN` translation, pruning-only behavior by default, explicit scan-side row filtering, SQL `SET`/`RESET`, and conservative statistics.


leaves12138 (approved):
Looks good to me. The pruning-only channel stays separate from exact row filtering, is used only for conservative Parquet statistics pruning, and is kept out of PK merge and data-evolution pre-merge reads. The _ROW_ID guard also covers pruning that can skip physical positions.

crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -64,6 +90,238 @@ fn to_datafusion_batch(batch: RecordBatch, schema: &ArrowSchemaRef) -> DFResult<
...
+    })?;
+
+    let (op, field, datum) = comparison;
+    match op {
+        Operator::Eq => predicate_builder.equal(field.name(), datum).ok(),
+        Operator::NotEq => predicate_builder.not_equal(field.name(), datum).ok(),
+        Operator::Lt => predicate_builder.less_than(field.name(), datum).ok(),
+        Operator::LtEq => predicate_builder.less_or_equal(field.name(), datum).ok(),
+        Operator::Gt => predicate_builder.greater_than(field.name(), datum).ok(),
+        Operator::GtEq => predicate_builder.greater_or_equal(field.name(), datum).ok(),
```
QuakeWang, line 261: Binary/VARBINARY ordering filters are unsafe here. DataFusion/Arrow uses unsigned lexicographic byte ordering, while Paimon evaluates `Datum::Bytes` using Java signed-byte ordering (`0xff < 0x00`). Translating `<`, `<=`, `>`, or `>=` can therefore prune rows that DataFusion would keep.

crates/paimon/src/table/table_read.rs
```diff
@@ -204,6 +222,19 @@ impl<'a> PaimonTableRead<'a> {
...
+    fn with_row_filter(mut self, row_filter: bool) -> Self {
+        self.row_filter = row_filter;
+        self
+    }
+
+    fn merge_predicates(&self) -> &[Predicate] {
+        if self.row_filter {
+            &self.data_predicates
+        } else {
+            &[]
```
QuakeWang: With the default `row_filter = false`, existing pushed predicates are removed from data-evolution reads. This moves filtering after Blob materialization, so a descriptor from a logically filtered-out row may still be resolved and fail the query. This changes the behavior previously covered by the default Blob regression test.

crates/integrations/datafusion/src/physical_plan/scan.rs
```diff
@@ -420,23 +735,60 @@ impl ExecutionPlan for PaimonTableScan {
         let read_type = self.read_type.clone();
         let pushed_predicate = self.pushed_predicate.clone();
         let case_sensitive = self.case_sensitive;
+        let runtime_filters = self.runtime_filters.clone();
+        let apply_row_filter = scan_applies_row_filter(context.session_config().options());
```
QuakeWang: Planning may return `PushedDown::Yes` and remove the parent `FilterExec`, but execution independently reads the row-filter setting from `TaskContext`. If the two contexts differ, this scan can emit rows that the removed parent filter would have rejected.

crates/paimon/src/table/data_evolution_reader.rs
```diff
@@ -448,6 +470,9 @@ impl DataEvolutionReader {
         resolve_blob_views: bool,
         descriptor_fields: &HashSet<String>,
     ) -> bool {
+        if !self.apply_row_filter {
```
QuakeWang, line 473: `apply_row_filter` defaults to `false` in DataFusion. This skips filtering before BLOB resolution, so rows excluded by `WHERE` may still resolve their descriptors. A corrupted or inaccessible descriptor can therefore fail the query before the parent `FilterExec` runs. The current test avoids the default path by explicitly enabling `paimon.read.row_filter`, leaving this blocking behavioral regression uncovered.

JingsongLi: We do not support cases involving damage, as this does not meet the requirements.

QuakeWang: Got it, others look good to me : )

QuakeWang (approved):
+1

PR #552, opened 2026-07-20
fix(datafusion): enforce blob reader location boundaries

QuakeWang:
### Purpose

`BlobReaderRegistry` used raw string prefix matching, allowing locations such as `.../foo` to match `.../foobar` and `s3://bucket` to match `s3://bucket-other`. This could route BLOB reads through the wrong registered `FileIO`.

### Brief change log

  - Match only the exact registered location or descendant path boundaries.
  - Preserve longest-match and registration semantics.
  - Cover slash, Windows, authority, and sibling-location boundaries with regression tests.

### Tests

  - `cargo test -p paimon-datafusion --lib blob_reader`
  - `cargo check -p paimon-datafusion --all-targets`

### DIFF
--- a/crates/integrations/datafusion/src/blob_reader.rs
+++ b/crates/integrations/datafusion/src/blob_reader.rs
@@ -31,6 +31,25 @@ pub struct BlobReaderRegistry {
     readers: Arc<RwLock<Vec<BlobFileIO>>>,
 }
 
+fn allows_backslash_boundary(location: &str) -> bool {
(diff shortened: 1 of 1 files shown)


leaves12138 (approved):
Reviewed the location-boundary matching, longest-match behavior, registration semantics, and Windows path handling. The focused unit tests, formatting check, and all-targets check pass locally. Looks good to me.

PR #553, opened 2026-07-20
docs: add TPC-DS SF100 performance report

JingsongLi:
## Summary

Publish an auditable TPC-DS SF100 read-performance report comparing Paimon Rust with DataFusion's native Parquet reader. Across the 96 queries that complete under the shared 32 GiB runtime limit, Paimon finishes the workload within 9.3% of native Parquet; the gap is 3.1% for queries whose native-Parquet median is at least five seconds.

The source Parquet files use SNAPPY, while the Paimon rewrite uses ZSTD and occupies 23.25% less physical data-file storage. The timing comparison therefore includes the higher-compression codec's additional decompression CPU cost as well as its storage and I/O savings.

## Changes

- replace the standalone HTML article with an English Markdown page integrated into the MkDocs theme, navigation, and documentation homepage
- summarize the validated measurements and protocol without committing the large per-iteration JSON reports
- document the exact benchmark revision, binary checksum, query revision, dataset layout, machine, cache policy, alternating source order, commands, and aggregation method
- document the SNAPPY-to-ZSTD compression change: Paimon data files use 27,975,615,232 bytes versus 36,449,466,908 bytes for the source, a 23.25% reduction, with more CPU-intensive decompression
- describe the StarRocks, Apache Doris, and Milvus integration directions, including the materialized vector-search C FFI path for Milvus
- scope the conclusion to the tested unpartitioned append-only Parquet-backed read path and describe other Paimon capabilities without presenting them as measured

## Testing

- [x] `cargo clean`
- [x] `cargo build --release -p paimon-tpcds-bench`
- [x] run all 96 comparable queries for both sources with one warmup and three measured iterations
- [x] run Q67, Q78, and Q97 separately for both sources and retain their resource-limit errors
- [x] verify query IDs, iteration counts, zero normal-query errors, zero warmup failures, zero normal-query spills, and all 288 output-row-count comparisons
- [x] verify all 24 source files use SNAPPY and all 125 Paimon data files use ZSTD from their Parquet metadata
- [x] recalculate the published workload medians from the local benchmark reports
- [x] `python3 scripts/validate_asf_yaml.py`
- [x] `cd docs && mkdocs build --strict`

## Notes

This is a TPC-DS-derived non-TPC benchmark, not an official TPC result. Q67, Q78, and Q97 are excluded from the aggregate because both sources exhaust the same 32 GiB DataFusion memory pool.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -53,3 +53,16 @@ mkdocs build
 ```
 
 The output will be in the `docs/site/` directory.
+
+## Performance Report
+
+The TPC-DS SF100 performance report is maintained as
+`docs/src/benchmark.md` and is included in the MkDocs navigation. Preview it
+as part of the documentation site:
+
+```bash
+cd docs
+mkdocs serve
+```
+
+Then open <http://127.0.0.1:8000/benchmark/>.
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -49,6 +49,7 @@ nav:
   - Home: index.md
   - Getting Started: getting-started.md
   - SQL Integration: sql.md
+  - Performance: benchmark.md
   - Go Integration: go-binding.md
   - Python Integration: python-binding.md
   - Architecture: architecture.md
--- a/docs/src/benchmark.md
+++ b/docs/src/benchmark.md
@@ -1,0 +1,305 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements. See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership. The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied. See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Paimon Rust at Native Speed
+
+## TPC-DS SF100 read performance with DataFusion
+
+!!! warning "Benchmark disclosure"
+
+    This is a TPC-DS-derived non-TPC benchmark. It is not an official TPC
+    result and must not be compared with official TPC results.
+
+This experiment compares Paimon Rust with DataFusion's native Parquet reader
+using the same DataFusion 54 SQL, planner, execution operators, query files,
+runtime limits, and local SSD. It measures 24 unpartitioned, append-only
+Paimon tables imported from the SF100 Parquet source.
+
+## Result
+
+Each query received one warmup followed by three measured iterations. The
+table reports the median of three workload totals, where each workload total
+is the sum of `execution_ms` for the same measured-iteration index across the
+96 queries that completed in every measured iteration for both sources.
+
+| Source | Median workload execution time | Three-run range |
+| --- | ---: | ---: |
+| Paimon Rust | 317.61 s | 314.64–317.76 s |
+| Native Parquet | 290.53 s | 289.69–290.75 s |
+
+Paimon Rust completes the workload within 9.3% of DataFusion's native Parquet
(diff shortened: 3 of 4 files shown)


docs/src/html/benchmark/index.html
```diff
@@ -0,0 +1,434 @@
...
+<dt>Memory pool</dt>
+<dd>32 GiB</dd>
+</div>
+<div>
+<dt>Parallelism</dt>
+<dd>12 target partitions</dd>
+</div>
+<div>
+<dt>Run configuration</dt>
+<dd>0 warmup · 1 measured iteration</dd>
```
leaves12138: Blocking: this result uses 0 warmups and a single measured iteration, while `benchmarks/tpcds/README.md`'s published procedure uses `--warmup 1 --iterations 3`. The headline gap is only 1.39 seconds over roughly 515 seconds (0.27%), which can easily be smaller than run-to-run or cache noise; one observation cannot establish parity. Please rerun both sources for multiple measured iterations under the same cache policy (ideally alternating source order), publish an aggregate such as the median plus dispersion, and derive the headline numbers from that dataset.

docs/src/html/benchmark/index.html
```diff
@@ -0,0 +1,434 @@
...
+</div>
+<p>A high-performance lakehouse table-format core for multiple engines.</p>
+</div>
+<div class="references">
+<a href="https://github.com/apache/paimon-rust" target="_blank" rel="noreferrer">Paimon Rust Project<!-- --> ↗</a>
+<a href="https://github.com/apache/paimon-rust/pull/549" target="_blank" rel="noreferrer">Runtime Filters and End-to-End Performance #549<!-- --> ↗</a>
+<a href="https://github.com/apache/doris/pull/65381" target="_blank" rel="noreferrer">Doris Paimon Write Architecture and Rust FFI Hook #65381<!-- --> ↗</a>
+<a href="https://github.com/StarRocks/starrocks/pull/73048" target="_blank" rel="noreferrer">StarRocks Paimon Vector Search #73048<!-- --> ↗</a>
+</div>
+<small>Community engineering update · July 20, 2026 · Results from a locally reproducible experiment</small>
```
leaves12138: Blocking: the page labels this a locally reproducible experiment, but the PR contains neither raw benchmark JSON nor enough metadata to reproduce or verify the numbers. The benchmark README explicitly requires the exact paimon-rust commit, generated-data scale and generator version, file counts and physical bytes, OS/CPU/memory/storage model, and exact commands for every published report. Please add those details and link or commit the two raw result files so reviewers can verify the totals, failed queries, and the 30/96 count.

docs/src/html/benchmark/index.html
```diff
@@ -0,0 +1,434 @@
...
+<article id="top">
+<section class="hero section-shell">
+<div class="hero-glow" aria-hidden="true">
+</div>
+<div class="eyebrow">
+<span>
+</span> PERFORMANCE MILESTONE · JUL 2026</div>
+<h1><strong>Paimon Rust</strong> <em>At Native Speed</em>
+</h1>
+<p class="hero-lead">In a TPC-DS SF100 workload on DataFusion 54, Paimon Rust finishes the 96 comparable queries <strong>within 0.27% of native Parquet</strong>. For the queries that dominate execution time, Paimon is 12.9% faster in aggregate. Table semantics, snapshots, schema evolution, and lakehouse capabilities no longer need to come with a material analytical-performance penalty.</p>
```
leaves12138: Blocking accuracy issue: this benchmark harness imports the data into unpartitioned append-only Paimon tables (`benchmarks/tpcds/README.md`), so the run does not exercise primary-key merging, schema evolution, deletion vectors, row lineage, or multiple file formats. This sentence currently presents the result as evidence that those semantics carry no material performance penalty. Please scope the conclusion to the append-only SF100 read path tested here, and describe the broader features as capabilities rather than benchmarked outcomes.

leaves12138 (changes requested):
The standalone page builds correctly, but the public performance conclusion needs a stronger evidence chain before publication. I found three blocking issues around measurement stability, reproducibility, and benchmark scope.

leaves12138 (approved):
The previous blocking concerns have been addressed. The benchmark protocol, reproducibility details, and scope are now sufficiently documented. The uncommitted raw JSON is acceptable given its size; a compact summary artifact would be a non-blocking future improvement.

PR #555, opened 2026-07-20
feat(table): support the Lumina/DiskANN backend in primary-key vector read

JunRuiLee:
### Purpose

Part of #514.

Java primary-key (PK) vector search builds its ANN reader through the generic `GlobalIndexer.create(segment.indexType(), ...)` registry (`PkVectorAnnSegmentSearcher`), so a PK table may use either the native ivf-* (vindex) engine or the Lumina engine (DiskANN as its internal index). The Rust PK read path, however, hardcoded the vindex reader in two places — the ANN scorer and the per-segment metric verification — so a Java-written PK table with `fields.<col>.pk-vector.index.type = lumina` was unreadable (the vindex reader failed to open a Lumina segment). The Rust data-evolution/append vector path already dispatches on the backend; this change wires the same dispatch onto the PK path.

### Brief change log

- Add `VectorSearchMetric::from_lumina`, symmetric to the existing `from_vindex`.
- Resolve one `VectorIndexBackend` per PK search from the configured index type (a PK table uses a single index type — enforced on the Java side and already filtered on the Rust side), failing loud on an unrecognized type.
- `verify_pk_vector_segment_metrics`: read the Lumina metric from the segment's index metadata and the vindex metric from the segment file bytes.
- ANN scorer: build `LuminaVectorGlobalIndexReader` or `VindexVectorGlobalIndexReader` based on the resolved backend.

### Tests

- Unit tests: `from_lumina` variant mapping; the Lumina branch of the segment metric check (matching + mismatch, over a synthetic Lumina index-metadata blob); backend classification (`lumina`/`lumina-vector-ann` -> Lumina, ivf-* -> vindex).
- `cargo test -p paimon` passes (lib + integration).
- Note: an end-to-end Lumina PK read requires the Lumina native library, which is unavailable in this environment; a cross-language Lumina fixture is a follow-up.

### API and Format

No API or on-disk format change. Read-only; the data-evolution/append path is unchanged.

### Documentation

No documentation changes.

### DIFF
--- a/crates/paimon/src/table/pk_vector_scan.rs
+++ b/crates/paimon/src/table/pk_vector_scan.rs
@@ -330,9 +330,9 @@ fn plan_from_inputs(
                 source_meta,
                 path,
                 file_size,
-                // Not consumed on the search path: the vindex reader loads its
-                // metadata from the index file bytes and ignores this field, so an
-                // absent value defaulting to an empty vec is acceptable.
+                // The Lumina reader consumes this as its serialized index
+                // metadata; the vindex reader ignores it and loads metadata from
+                // the segment file bytes. Absent value defaults to an empty vec.
                 index_meta: gim.index_meta.clone().unwrap_or_default(),
             });
     }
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #556, opened 2026-07-20
[core] Search primary-key vector buckets and files concurrently

JunRuiLee:
### Purpose

Part of #514, and the last of three PRs split out of the original combined change on the primary-key vector `execute_read` path (after #559 streaming exact fallback and #560 batch multi-query, both now on `main`). This PR adds **intra-search parallelism**: a configurable concurrency limit fans the per-bucket and per-exact-file searches out, without changing results.

### Brief change log

- `perf(table)`: a `global-index.thread-num` option (default 32, rejected when non-positive) threads a concurrency limit through the search. The per-bucket loop (orchestrator) and per-exact-file loop (bucket search) fan out with `buffer_unordered` up to that limit; a limit of 1 takes a plain sequential loop so visit order stays deterministic. Results are accumulated by fixed query and split index and merged through the existing order-independent bounded heaps and global top-k, so output does not depend on completion order. The ANN segment scan stays sequential (synchronous CPU work).
- `refactor(table)`: the exclusion predicate threaded through the search (residual filter ∩ deletion vectors) is marked `Send + Sync` so the search future is `Send` and can run on a multi-threaded runtime.

### Tests

- Sequential call order asserted at both layers for concurrency 1; parallel (concurrency > 1) with tied/NaN/equal-score candidates completing out of order produces the deterministic ranking equal to the serial result.
- `global-index.thread-num` default/explicit/non-positive.
- `cargo test -p paimon --lib` green (1793 passed); `cargo clippy -p paimon --lib --tests` and `cargo fmt --check` clean.

### API and Format

No on-disk format change, no new result columns. The concurrency limit defaults to a value that does not change results, only scheduling. New public surface: the `global-index.thread-num` option (matches Java Paimon's option name, semantics, and default of 32).

### Documentation

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -24,6 +24,7 @@ const DATA_EVOLUTION_ENABLED_OPTION: &str = "data-evolution.enabled";
 const GLOBAL_INDEX_ENABLED_OPTION: &str = "global-index.enabled";
 const GLOBAL_INDEX_SEARCH_MODE_OPTION: &str = "global-index.search-mode";
 const GLOBAL_INDEX_ROW_COUNT_PER_SHARD_OPTION: &str = "global-index.row-count-per-shard";
+const GLOBAL_INDEX_THREAD_NUM_OPTION: &str = "global-index.thread-num";
 const GLOBAL_INDEX_COLUMN_UPDATE_ACTION_OPTION: &str = "global-index.column-update-action";
 const SORTED_INDEX_RECORDS_PER_RANGE_OPTION: &str = "sorted-index.records-per-range";
 const BTREE_INDEX_FALLBACK_SCAN_MAX_SIZE_OPTION: &str = "btree-index.fallback-scan-max-size";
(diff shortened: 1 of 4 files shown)


JingsongLi:
`global-index.thread-num` is supposed to be the maximum concurrency for global-index I/O, but the same value is applied independently at both levels: up to N bucket futures here, and each bucket starts up to N exact-file futures in `bucket_search_batch`. Therefore, with the default N=32, a single search can issue up to 1024 exact-file searches concurrently, and concurrent queries multiply this further. This differs from Java's shared `GlobalIndexReadThreadPool`, which caps the whole search at N. I reproduced this with two buckets, two files per bucket, and `concurrency=2`; four file searches were active simultaneously. Could we use one shared semaphore/concurrency budget across both levels, or only fan out at one level, and add a test that verifies the peak concurrency?

JunRuiLee:
Thanks @JingsongLi, you're exactly right — the limit was applied independently at each level, so a single search could reach N×N concurrent exact-file reads (and concurrent queries multiplied it further).

Fixed in 5293431. Instead of two independent `buffer_unordered(N)` limits, the search now threads a single shared `Arc<Semaphore>` of N permits through the whole search, mirroring Java's single shared `GlobalIndexReadThreadPool`. A permit is acquired only around leaf exact-file I/O; the per-bucket orchestration never holds one, so it can't starve leaf work — the async analogue of the "start from the caller" note in `PrimaryKeyVectorRead.searchBuckets`. Total in-flight exact-file searches are now capped at N overall rather than per bucket, and the `concurrency == 1` path stays strictly sequential.

I also added `search_candidates_peak_concurrency_capped_across_buckets`, which reproduces your scenario (two buckets, two files each, `concurrency = 2`): it instruments the exact-file search to record the peak number of simultaneously active searches and asserts it stays ≤ N. I confirmed the test observes peak = 4 and fails without the shared budget, and peak = 2 (passing) with it.

JingsongLi (approved):
+1

PR #557, opened 2026-07-20
build(python): give RC artifacts unique versions

XiaoHongbo-Hope:
## Summary

- Derive a PEP 440 Python version from the release tag.
- Build v0.3.0-rcN as 0.3.0rcN without changing the Cargo workspace version.
- Validate that the tag base matches workspace.package.version.
- Use the requested release tag for checkout, publication gating, and PyPI index selection.
- Pin release actions and keep Windows version setup on Bash.

## Why

RC1 and RC2 currently both build pypaimon-rust 0.3.0. Because TestPyPI publishing uses skip-existing, RC2 can silently reuse RC1 artifacts. Unique RC versions keep retries idempotent while ensuring different RCs publish different files.

## Verification

- v0.3.0 keeps dynamic version 0.3.0.
- v0.3.0-rc1 and v0.3.0-rc2 become 0.3.0rc1 and 0.3.0rc2.
- Mistagged and malformed versions fail.
- Manual dispatch checks out and publishes the requested tag.
- Maturin 1.14.1 produced pypaimon_rust-0.3.0rc2.tar.gz with matching PKG-INFO and pyproject metadata.
- Ruff, Python byte compilation, YAML parsing, actionlint, and git diff --check passed.

### DIFF
--- a/.github/workflows/release_python_binding.yml
+++ b/.github/workflows/release_python_binding.yml
@@ -56,7 +56,7 @@ jobs:
           --tag "${RELEASE_TAG}"
           --pyproject bindings/python/pyproject.toml
 
-      - uses: PyO3/maturin-action@v1
+      - uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1.51.0
         with:
           working-directory: bindings/python
           command: sdist
@@ -84,6 +84,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Set Python release version
+        shell: bash
         env:
           RELEASE_TAG: ${{ inputs.tag || github.ref_name }}
         run: >
@@ -91,7 +92,7 @@ jobs:
           --tag "${RELEASE_TAG}"
           --pyproject bindings/python/pyproject.toml
 
-      - uses: PyO3/maturin-action@v1
+      - uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1.51.0
         with:
           working-directory: bindings/python
           target: ${{ matrix.target }}
@@ -174,7 +175,7 @@ jobs:
 
       - name: Publish to TestPyPI
         if: contains(github.ref, '-')
-        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
         with:
           repository-url: https://test.pypi.org/legacy/
           skip-existing: true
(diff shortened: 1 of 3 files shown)


.github/workflows/release_python_binding.yml
```diff
@@ -70,6 +83,14 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      - name: Set Python release version
+        env:
+          RELEASE_TAG: ${{ inputs.tag || github.ref_name }}
+        run: >
+          python3 scripts/python_release_version.py
+          --tag "${RELEASE_TAG}"
```
QuakeWang, line 95: The `wheels` matrix includes `windows-latest`, where `run` uses PowerShell by default. `${RELEASE_TAG}` does not read the environment variable in PowerShell, so the Windows wheel job receives an empty tag and fails.

scripts/python_release_version.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+import tomllib
+from pathlib import Path
+
+
+TAG_PATTERN = re.compile(r"^v(?P<base>\d+\.\d+\.\d+)(?:-rc(?P<rc>[1-9]\d*))?$")
```
QuakeWang, line 29: The workflow tag filter accepts `vX.Y.Z-rc0`, but this pattern rejects it. Such a tag will trigger the release workflow and then fail during version validation.

XiaoHongbo-Hope: The trigger is intentionally broad. Semantic validation rejects RC0; the release guide starts candidates at RC1.

.github/workflows/release_python_binding.yml
XiaoHongbo-Hope: Fixed by running the version step with Bash on every matrix OS.

JingsongLi:
- The tag provided as input to `workflow_dispatch` is used only for this release; the checkout does not use this tag, and the build actually runs on the manually selected branch.
   - The `release` job is still restricted by `startsWith(github.ref, ‘refs/tags/’)`, so manual triggers will never result in a release; TestPyPI/PyPI checks also continue to use `github.ref` rather than the input tag.
   - It is recommended to uniformly generate `RELEASE_TAG`, check out `ref: RELEASE_TAG`, allow dispatch to proceed to release, and use `RELEASE_TAG` to determine RC status.

XiaoHongbo-Hope:
Fixed in 16c2d1a: manual dispatch now uses RELEASE_TAG for checkout, release gating, and TestPyPI/PyPI selection.

XiaoHongbo-Hope:
> Fixed in [16c2d1a](https://github.com/apache/paimon-rust/commit/16c2d1a3694b87532114a338b3ddd8ae66b6e6c4): manual dispatch now uses RELEASE_TAG for checkout, release gating, and TestPyPI/PyPI selection.

QuakeWang (approved):
+1

PR #560, opened 2026-07-20
[core] Batch primary-key vector search over a shared plan

JunRuiLee:
### Purpose

Part of #514, and the second of three PRs split out of the original combined change on the primary-key vector `execute_read` path. This PR adds **batch multi-query search**: N query vectors share one snapshot/manifest plan, segment preload, and opened readers. Single-query output stays byte-identical (single == batch-of-one).

> **Stacked PR.** Its parent — streaming exact fallback (#559) — has merged, so this branch is now rebased onto `main` and the diff is just the two batch commits. Full stack:
> 1. streaming exact fallback — #559 (merged)
> 2. **this PR** — batch multi-query
> 3. parallel search — #556 (stacked on this)

### Brief change log

- `feat(vindex)`: the bucket-local ANN search gains a batch entry so N queries share one pass over each ANN segment and each uncovered file. The ANN searcher builds the live-row mask once (query-independent), opens one reader per segment, and drives `visit_batch_vector_search`; the exact searcher scores all queries in one stream pass. A single query short-circuits to the existing single-query path so its result is unchanged. The batch ANN scorer dispatches Lumina vs. vindex readers per segment backend, matching the single-query path.
- `feat(table)`: `BatchVectorSearchBuilder` gains `with_filter`/`with_projection` and an `execute_read` returning one Arrow stream per query in input order (empty streams preserve arity; any query error fails the whole call). The builder factors the query-independent plan (snapshot, segment preload, residual allow-list, backend, ANN scorer) into a shared step reused across queries. The single-query `plan_and_search_pk_candidates` becomes a thin batch-of-one wrapper whose output is byte-identical. The scored `execute()` stays fail-loud on the primary-key path, and a data-evolution `execute_read` validates the target column exists and is a FLOAT vector column (`ARRAY`/`VECTOR`) up front so an unknown/scalar/non-float column fails loud instead of returning an empty stream.

### Tests

- Single == batch-of-one at every layer; per-query heaps independent under a shared residual/DV; arity preserved on an empty snapshot; malformed query, zero limit, and reserved/unknown/scalar/non-float columns fail loud (core + C FFI coverage).
- `cargo test -p paimon` green; `cargo build -p paimon-datafusion` clean; `cargo clippy -p paimon --all-targets -D warnings` and `cargo fmt --check` clean.

### API and Format

No on-disk format change, no new result columns. Single-query output is byte-identical. New public surface: `BatchVectorSearchBuilder::{with_filter, with_projection, execute_read}`.

### Documentation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4426,6 +4426,7 @@ dependencies = [
  "hmac 0.12.1",
  "indexmap 2.14.0",
  "libloading 0.9.0",
+ "log",
  "lz4_flex 0.11.6",
  "md-5 0.10.6",
  "opendal-core",
--- a/crates/paimon/src/table/pk_vector_data_file_reader.rs
+++ b/crates/paimon/src/table/pk_vector_data_file_reader.rs
@@ -519,6 +519,60 @@ mod integration_tests {
         assert_eq!(streamed[0][1].row_position, 3);
     }
 
+    /// A multi-query `search_file` returns independent per-query Top-K lists: the
+    /// slot for a query in a batch is identical to that query searched alone (no
+    /// cross-query bleed), and a shared `is_excluded` applies to every query.
+    #[tokio::test]
+    async fn search_file_multi_query_returns_independent_per_query_top_k() {
+        let rows = vec![
+            Some(vec![0.0, 0.0]),
+            Some(vec![1.0, 0.0]),
+            Some(vec![2.0, 0.0]),
+            Some(vec![3.0, 0.0]),
+        ];
+        let (factory, file_name) =
+            build_factory(&rows, rows.len() as i64, "memory:/pkvdfr_multiquery").await;
+        let active = BucketActiveFile {
+            file_name,
+            row_count: rows.len() as i64,
+        };
+        // Exclude physical position 1 for all queries (shared predicate).
+        let is_excluded = |pos: i64| pos == 1;
+        let q0 = [0.0f32, 0.0]; // nearest is pos 0
+        let q1 = [3.0f32, 0.0]; // nearest is pos 3
+
+        let batch = factory
+            .search_file(
+                &active,
+                &[&q0, &q1],
+                VectorSearchMetric::L2,
+                2,
+                &is_excluded,
+            )
+            .await
+            .unwrap();
+        assert_eq!(batch.len(), 2, "one result list per query");
+
+        // Each query searched alone must equal its slot in the batch.
+        let only_q0 = factory
+            .search_file(&active, &[&q0], VectorSearchMetric::L2, 2, &is_excluded)
+            .await
+            .unwrap();
+        let only_q1 = factory
+            .search_file(&active, &[&q1], VectorSearchMetric::L2, 2, &is_excluded)
+            .await
+            .unwrap();
+        assert_eq!(batch[0], only_q0[0]);
+        assert_eq!(batch[1], only_q1[0]);
+
+        // Sanity: distinct nearest neighbours, and the excluded position is absent.
+        assert_eq!(batch[0][0].row_position, 0);
+        assert_eq!(batch[1][0].row_position, 3);
+        assert!(batch
+            .iter()
+            .all(|list| list.iter().all(|r| r.row_position != 1)));
+    }
+
     /// A `DataFileMeta.row_count` larger than the file's real row count means the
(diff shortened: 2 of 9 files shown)


JingsongLi:
- `ARRAY<FLOAT>` is a valid PK-vector column, but batch searches only perform query validation in advance for `VECTOR<FLOAT>`.
   - For empty tables or empty execution plans, queries on `ARRAY<FLOAT>` that contain NaN or Inf will be returned as normal empty results rather than reporting an invalid input. The minimal regression I added consistently failed.

JunRuiLee:
> * `ARRAY<FLOAT>` is a valid PK-vector column, but batch searches only perform query validation in advance for `VECTOR<FLOAT>`.
> * For empty tables or empty execution plans, queries on `ARRAY<FLOAT>` that contain NaN or Inf will be returned as normal empty results rather than reporting an invalid input. The minimal regression I added consistently failed.

Thanks, fixed. Batch PK-vector query validation now also covers `ARRAY<FLOAT>` before planning, using the configured index dimension, so NaN/Inf queries fail loud even for empty plans. Added a regression test for the empty `ARRAY<FLOAT>` table case.

JingsongLi:
**1. Missing authorization check**

 `BatchVectorSearchBuilder::execute()` no longer calls `CoreOptions::ensure_read_authorized()`, although the previous implementation performed this check before validation and empty-table fast paths. This method returns data-derived row IDs and scores outside `TableScan` / `TableRead`, so a table with `query-auth.enabled=true` can bypass the fail-closed read authorization through the batch scored-search API. Please restore the authorization check at the beginning of `execute()` and add a batch-specific regression test.

**2. `ARRAY<FLOAT>` with Lumina is incorrectly handled as vindex**

 `pk_vector_query_dimension()` unconditionally uses `VindexVectorIndexOptions` for every `ARRAY<FLOAT>` column. For a valid PK-vector configuration using `lumina` or `lumina-vector-ann`, this fails with `Unsupported vindex index type` before planning, including on an empty table. Please resolve the dimension according to the configured backend: use the effective `lumina.index.dimension` for Lumina and `VindexVectorIndexOptions` only for vindex backends. A regression test covering `ARRAY<FLOAT>` with Lumina would also be useful.

**3. Default projection no longer rejects reserved-column conflicts**

 The previous implementation validated the resolved field list after applying the projection, which also caught reserved-name conflicts in the default projection. The new resolver only validates explicitly requested projection names. As a result, a table containing a user column such as `__paimon_search_score`, `_PKEY_VECTOR_POSITION`, or `_ROW_ID` can silently pass on an empty result and may produce duplicate or incorrectly interpreted metadata columns for non-empty results. Please restore the post-resolution reserved-field validation for both the single-query and batch resolvers, along with the removed default-projection regression tests.

JunRuiLee:
Addressed in fd72a9a (fixes) and 5fd4157 (cleanup):

**1. Missing authorization check** — `BatchVectorSearchBuilder::execute()` now calls `ensure_read_authorized()` at the very top, before config validation and any fast path, matching `execute_read` and the single-query builder. A `query-auth.enabled=true` table previously returned empty results on an empty snapshot and bypassed the fail-closed guard; it now fails closed. Added a batch-specific regression test.

**2. `ARRAY<FLOAT>` with Lumina** — `pk_vector_query_dimension()` now resolves the dimension per the configured backend: Lumina via `LuminaVectorIndexOptions` (`lumina.index.dimension`), vindex/IVF via `VindexVectorIndexOptions`. A valid Lumina `ARRAY<FLOAT>` PK-vector column is no longer rejected as `Unsupported vindex index type` before planning (including on an empty table). Added a regression test covering `ARRAY<FLOAT>` + Lumina.

**3. Default projection reserved-column conflicts** — both `resolve_materialize_read_type` resolvers (single-query and batch) now validate the *resolved* field list against the reserved metadata names, so the default all-columns projection also rejects a colliding user column, not only names in an explicit projection. The explicit pre-check is kept so a requested reserved name that is not a real column still fails as a reserved-column error rather than "column not found". Restored the default-projection regression tests for both resolvers.

Separately, dropped the stale `_PKEY_VECTOR_SCORE` reserved name (5fd4157): it was the score column's pre-rename name and nothing emits it anymore — the injected metadata columns are `__paimon_search_score`, `_PKEY_VECTOR_POSITION`, and `_ROW_ID`.

JingsongLi (approved):
+1

PR #561, opened 2026-07-20
build(bindings): package complete legal metadata

XiaoHongbo-Hope:
## Summary

- Generate target-specific Python and Go legal metadata from the locked dependency graph.
- Package and byte-verify LICENSE, NOTICE, and third-party license reports.
- Verify report freshness before publishing Python and Go artifacts.
- Pin the ARM64 OpenSSL snapshot and validate its SHA256 checksums.
- Validate every Python wheel member against WHEEL RECORD.

## Coverage

- Python: five targets, including XZ Utils 5.8.3 and target-specific Linux OpenSSL.
- Go: four native targets with matching third-party reports.
- Apache Avro, Arrow, DataFusion, and Object Store notices.
- AWS-LC composite license text on the final dependency tree.

## Verification

- python3.11 scripts/release_licenses.py --check
- python3.11 scripts/dependencies.py verify
- cargo metadata --locked
- Python sdist build and artifact verification
- WHEEL RECORD tamper test
- actionlint and git diff --check


JingsongLi (changes requested):
The target-specific third-party license reports are required in the binary wheels and Go artifacts, but I do not think all generated outputs should be committed to the main source tree. This PR adds nine largely duplicated reports, totaling about 99,000 lines and 5.7 MB.

More importantly, `scripts/release.sh` creates the ASF source archive with `git archive HEAD`, and these files are not marked `export-ignore`. The source release will therefore contain target-specific `LICENSE`, `NOTICE`, and third-party reports that explicitly describe native binaries which are not present in that source distribution. ASF guidance requires legal metadata to accurately reflect the contents of each distribution and explicitly notes that source and binary legal files may differ.

Could we keep the cargo-about configuration, templates, curated license inputs, generation script, and artifact verifiers in the repository, but generate each target-specific report in its wheel/Go build job and package it directly with the corresponding binary artifact? For the Go module, the reports can be added only to the release-tag commit together with the native libraries. If the generated reports must remain tracked, they should at least be excluded from the ASF source archive with `export-ignore`.

XiaoHongbo-Hope:
Thanks, addressed in ab816a3. Target-specific Python/Go legal files are now generated in the corresponding binary jobs and removed from the main tree; Go files are added only to the release-tag commit. The Python source-only report remains in the ASF source archive. Also added archive-path collision and dependency-derived NOTICE checks.

QuakeWang (approved):
LGTM

JingsongLi (approved):
+1

PR #562, opened 2026-07-20
build: harden Rust crate publishing

XiaoHongbo-Hope:
## Summary

- Package LICENSE, NOTICE, test helpers, and testdata in both published crates.
- Validate release tags and package inventories with locked dependencies.
- Publish paimon, wait for the exact crates.io version, then dry-run and publish paimon-datafusion.
- Skip versions that already exist so failed GA runs can be retried safely.

## Why

The current matrix uses max-parallel: 1, which limits concurrency but does not guarantee paimon runs before paimon-datafusion. The packaged crates also omit legal files and reference test helpers outside their package roots.

RC validation does not dry-run both crates together: it packages both crates and dry-runs paimon. The paimon-datafusion dry-run runs only after paimon is visible on crates.io during GA publication.

## Verification

- cargo package --locked -p paimon -p paimon-datafusion --all-features
- Package inventories contain LICENSE, NOTICE, dependency reports, test helpers, and testdata
- Rust formatting, YAML parsing, actionlint, and diff checks
- Static workflow assertions confirm the explicit job dependency chain and separate dry-runs


jianguotian:
+1

JingsongLi (approved):
+1

PR #566, opened 2026-07-21
feat(blob): support ARRAY<BLOB> reads

QuakeWang:
### Purpose

Rust BLOB reads only recognized scalar BLOB fields. As a result, Java-written `ARRAY<BLOB>` columns stored in dedicated `.blob` files were not decoded or routed through data-evolution fallback.

### Brief change log

  - Decode Java `ARRAY<BLOB>` payloads in inline and descriptor modes.
  - Route `ARRAY<BLOB>` fields through dedicated blob sources and preserve placeholder fallback across sequence groups.
  - Add a Java-compatible fixture and end-to-end coverage.

### Tests

  - `cargo build --features fulltext,vortex`

### API and Format

No API or storage-format changes. This adds read compatibility for the existing Java `ARRAY<BLOB>` format.


JingsongLi (approved):
+1

PR #573, opened 2026-07-21
feat(partial-update): support ignore-delete semantics

### LINKED ISSUE #572: partial-update: support ignore-delete semantics
### Search before asking

- [x] I searched the existing issues and pull requests and found nothing similar.

### Please describe the bug

Paimon Rust currently rejects `ignore-delete` options on primary-key tables using
`merge-engine=partial-update`, even though Java Paimon supports this mode.

For example, creating or opening a table with either option fails validation:

```text
merge-engine=partial-update
ignore-delete=true
```

or:

```text
merge-engine=partial-update
partial-update.ignore-delete=true
```

The read-side and flush-time partial-update merge functions also reject
`DELETE` and `UPDATE_BEFORE` rows unconditionally. This prevents Rust readers
from opening Java-created tables and historical files that use the supported
ignore-delete semantics.

Expected behavior:

- Both `ignore-delete` and `partial-update.ignore-delete` are accepted.
- With an effective value of `true`, `DELETE` and `UPDATE_BEFORE` are ignored.
- The default and an explicit `false` still reject retract rows.
- Ignored retract rows do not consume writer buffer, sequence numbers, or enter
  input changelog files.
- Once enabled, the effective value cannot be changed from `true` to `false`,
  matching Java Paimon compatibility safeguards.
- Other advanced partial-update options remain unsupported.

### Proposed solution

- Narrowly allowlist the two supported option keys in partial-update validation.
- Apply ignore-delete semantics in both read-side merge and write-side filtering.
- Filter direct `_VALUE_KIND` batches before writer buffering.
- Add schema, writer, reader, core integration, and DataFusion integration tests.
- Document the supported behavior and remaining limitations.

shyjsarah:
### Purpose

Linked issue: close #572

Paimon Rust currently rejects `ignore-delete` options for `merge-engine=partial-update` tables and unconditionally fails on `DELETE` / `UPDATE_BEFORE` rows. This prevents compatibility with Java Paimon tables and historical files that use the supported ignore-delete semantics.

This change aligns the basic Rust partial-update implementation with Java for this option while keeping advanced partial-update modes unsupported.

### Brief change log

- Accept exact `ignore-delete` and `partial-update.ignore-delete` options for partial-update tables, including explicit `false`.
- Ignore `DELETE` and `UPDATE_BEFORE` during read-side partial-update merging when enabled, including retract-only keys.
- Filter direct `_VALUE_KIND` retract rows before writer buffering, sequence allocation, data-file merge, and input changelog generation.
- Reject changing the effective option from `true` back to `false`, including alias removal and canonical-key override cases.
- Keep sequence groups, partial aggregation, remove-record options, and deletion-vector combinations fail-fast.

### Tests

- [x] `cargo test -p paimon --test incremental_batch_scan_test`
- [x] `cargo test -p paimon-datafusion --test pk_tables`
- [x] `cargo clippy -p paimon --lib --tests -- -D warnings`
- [x] `cargo clippy -p paimon-datafusion --test pk_tables -- -D warnings`

Coverage includes:

- canonical and alias option parsing;
- explicit `false`;
- `DELETE` and `UPDATE_BEFORE`;
- retract-only keys;
- writer buffer, sequence, data, and input changelog behavior;
- ALTER attempts from effective `true` to `false`;
- core table and DataFusion SQL end-to-end reads.

### API and Format

No public API or storage-format change. The change only enables an existing Java-compatible table option and filters retract rows according to that option.

### Documentation

Updated the SQL table-option documentation with the supported keys, behavior, ALTER restriction, and remaining partial-update limitations.

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -453,6 +453,15 @@ impl TableSchema {
         new_schema.highest_field_id =
             highest_field_id.max(Self::current_highest_field_id(&new_schema.fields));
 
+        if PartialUpdateConfig::new(&new_schema.options).is_enabled()
+            && self.core_options().ignore_delete()
+            && !CoreOptions::new(&new_schema.options).ignore_delete()
+        {
+            return Err(crate::Error::Unsupported {
+                message: "Cannot change ignore-delete from true to false.".to_string(),
+            });
+        }
+
         // Re-run create-time validations on the final schema, mirroring Java
         // `SchemaValidation.validateTableSchema` after applying changes.
         Schema::validate_key_field_types(
@@ -2099,7 +2108,7 @@ mod tests {
     #[test]
     fn test_partial_update_schema_validation_rejects_unsupported_options() {
         for (key, value) in [
-            ("ignore-delete", "true"),
+            ("fields.value.ignore-delete", "true"),
             ("fields.value.sequence-group", "g1"),
             ("fields.default-aggregate-function", "last_non_null"),
         ] {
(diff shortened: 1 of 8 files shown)


crates/paimon/src/spec/schema.rs
```diff
@@ -453,6 +453,15 @@ impl TableSchema {
         new_schema.highest_field_id =
             highest_field_id.max(Self::current_highest_field_id(&new_schema.fields));
 
+        if PartialUpdateConfig::new(&new_schema.options).is_enabled()
```
jerry-024, line 456: **[major] Preserve the one-way `ignore-delete` invariant across merge-engine changes**

This guard only runs when the resulting schema is still `partial-update`, so the invariant can be bypassed across schema versions:

1. start with `merge-engine=partial-update` and `ignore-delete=true`
2. set `merge-engine=deduplicate` (this guard is disabled)
3. set `ignore-delete=false` or remove it (this guard is still disabled)
4. set `merge-engine=partial-update` again (the previous effective value is already false)

All calls to `apply_changes` succeed, and historical `DELETE` / `UPDATE_BEFORE` rows can become unreadable again. The Java `SchemaManager.checkAlterTableOption` enforces the `true -> false` restriction on the option itself, independent of merge engine.

Please preserve the effective true state across engine changes, or reject disabling/removing it whenever the current schema has it enabled. A regression test covering the four-step transition above would catch this.

JingsongLi (approved):
+1

PR #574, opened 2026-07-21
perf(datafusion): push filters into Paimon Parquet readers

JingsongLi:
## Summary

Make pushed predicates exact by default in Paimon readers and allow DataFusion runtime filters to run during Parquet decoding without exposing Parquet-specific types from Paimon Core. Prefer native Paimon decoder predicates for supported static filters while reserving the format-neutral RowFilter path for dynamic and otherwise uncovered expressions.

## Changes

- apply supported Paimon predicates exactly by default, with residual filtering for compound or unsupported decoder predicates
- add format-neutral Arrow `RowFilter`, `RowFilterContext`, and `RowFilterFactory` APIs based on projected `RecordBatch` evaluation
- adapt DataFusion physical expressions per file schema, split conjunctions into independent row filters, and keep live dynamic-filter updates visible
- recognize static physical comparisons, NULL checks, IN, BETWEEN, boolean composition, LIKE, and string predicates already covered by the pushed Paimon predicate
- use the native Paimon Parquet row filter for covered static predicates to avoid duplicate decoder evaluation; keep dynamic and uncovered expressions in the generic decoder path
- adapt generic row filters to Parquet `ArrowPredicate`s inside the Parquet reader
- preserve Parquet predicate cost ordering using compressed column sizes and fall back safely when a filter cannot be adapted
- retain exact scan-level filtering because decoder filtering is an optimization and may be unavailable for a file or read path
- keep primary-key merge, schema evolution, `_ROW_ID`, and position-based reads on correctness-preserving paths
- remove the internal row-filter toggle and benchmark switch so exact filtering is the default behavior
- align the TPC-DS DataFusion and Paimon reader batch size through `--batch-size`

## Testing

- [x] `cargo test -p paimon --lib` — 1705 passed, 1 ignored
- [x] `cargo test -p paimon-datafusion --lib physical_plan::scan::tests` — 26 passed
- [x] `cargo test -p paimon-datafusion --test read_tables test_temporal_filter_pushdown_via_datafusion_scan -- --exact` — passed
- [x] `cargo test -p paimon --features vortex --lib arrow::format::vortex::tests` — 16 passed
- [x] `cargo check -p paimon-datafusion -p paimon-tpcds-bench`
- [x] `cargo clippy -p paimon -p paimon-datafusion --all-targets -- -D warnings`

## Notes

The row-filter contract is intentionally format-neutral so another reader such as Vortex can implement an adapter later. This PR only connects it to Parquet; Vortex behavior is unchanged.


leaves12138 (approved):
Reviewed the filter pushdown lifecycle, Parquet reader integration, dynamic filter handling, primary-key merge semantics, schema evolution, and fallback paths. Targeted Parquet and DataFusion tests pass. No blocking issues found.

PR #576, opened 2026-07-21
fix(datafusion): avoid loading tables for SHOW TABLES

### LINKED ISSUE #575: SHOW TABLES loads full table providers
### Search before asking

### Please describe the bug

In the DataFusion integration, `SHOW TABLES` is evaluated through `information_schema.tables`. DataFusion asks each schema provider for the type of every listed table. The default `SchemaProvider::table_type` implementation calls `table(name)`, which makes the Paimon schema provider load the full table provider and initialize its FileIO.

As a result, metadata listing can fail when a catalog can list a table but the table's storage is not accessible from the current client. `SHOW TABLES` should only require catalog metadata; an actual query against that table should continue to load the provider and report any storage error.

A minimal reproduction is a catalog implementation where:

- `list_tables("default")` returns `metadata_only`
- `get_table(...)` returns an error indicating that table loading is unavailable
- executing `SHOW TABLES` currently propagates that error

Expected behavior: `SHOW TABLES` lists `metadata_only` without calling `get_table`.

### Solution

Override `PaimonSchemaProvider::table_type` and reuse table/view types collected by `table_names`. Temporary objects should retain their provider-reported type, catalog views should remain `VIEW`, and direct table reads should keep the existing lazy-loading behavior.

shyjsarah:
### Purpose

Linked issue: close #575

`SHOW TABLES` should use catalog metadata without opening every listed table. The default DataFusion `SchemaProvider::table_type` implementation calls `table(name)`, which can initialize FileIO and turn a metadata-only operation into a table access.

### Brief change log

- collect base-table and view types while listing catalog objects
- override `PaimonSchemaProvider::table_type` to use the collected metadata instead of loading table providers
- preserve temporary table/view precedence and catalog view classification
- keep direct table reads on the existing lazy-loading path
- add regression coverage for unavailable table loading and catalog view types

### Tests

- `cargo test -p paimon-datafusion --test sql_context_tests`
- `cargo clippy -p paimon-datafusion --all-targets -- -D warnings`

### API and Format

### Documentation

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -25,7 +25,7 @@ use std::sync::RwLock;
 use async_trait::async_trait;
 use datafusion::catalog::{CatalogProvider, MemorySchemaProvider, SchemaProvider};
 use datafusion::common::{plan_datafusion_err, Column};
-use datafusion::datasource::TableProvider;
+use datafusion::datasource::{TableProvider, TableType};
 use datafusion::error::Result as DFResult;
 use datafusion::execution::SessionState;
 use datafusion::logical_expr::{expr_fn::cast, Expr, LogicalPlan, LogicalPlanBuilder};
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #578, opened 2026-07-21
fix(table): reject invalid global index metadata

### LINKED ISSUE #577: Missing global index metadata can silently prune matching rows
### Search before asking

### Please describe the bug

On the latest main commit a9a8f6b, GlobalIndexScanner converts missing or unparseable sorted global-index metadata into BTreeIndexMeta(None, None, false). BTreeIndexMeta then interprets that value as an only-null index.

For equality and range predicates, file-level pruning reports no match without opening the physical index file. The scanner returns an indexed result such as Some([]) instead of None, so rows covered by that index shard can be silently omitted in fast, full, and detail search modes.

A minimal reproduction uses two BTree shards for the same field: one with valid metadata and one with missing metadata. Querying a key present in both shards returns only the valid shard result instead of falling back to the normal table scan.

Malformed key lengths are related: BTreeIndexMeta::deserialize currently slices with unchecked lengths and can panic before the scanner can handle the parse error.

### Solution

Preserve missing or unparseable metadata as unavailable and return None when evaluating that field, allowing the caller to use the normal table scan. Validate serialized key lengths before slicing so malformed metadata returns InvalidData instead of panicking.

- [x] I am willing to submit a PR.

XiaoHongbo-Hope:
## Summary

Fixes #577.

Missing or malformed metadata for a declared sorted global-index entry could previously be converted into an empty `BTreeIndexMeta`. That metadata is interpreted as an only-null index, so file pruning could silently omit matching rows. Malformed key lengths could also panic during deserialization, and a truncated null-flags trailer could be interpreted as legacy metadata.

Because the manifest already declares these entries as BTree or Bitmap indexes, missing or corrupt metadata is invalid table data. This change surfaces `DataInvalid` instead of silently falling back to a normal scan.

## Changes

- Change `GlobalIndexScanner::create` to return `Result<Option<Self>>`, reserving `Ok(None)` for cases with no applicable sorted index.
- Reject missing outer metadata, missing serialized index metadata, and deserialization failures for declared BTree/Bitmap entries; preserve the underlying parse error as the source.
- Validate serialized first-key and last-key lengths before slicing, returning `InvalidData` for negative or truncated lengths.
- Reject a one-byte, truncated versioned null-flags trailer instead of treating it as legacy metadata.
- Add regressions for all three scanner metadata errors, invalid key lengths, and truncated trailers.

## Testing

- [x] `cargo clippy -p paimon --all-targets -- -D warnings`
- [x] `cargo test -p paimon global_index_scanner` (37 passed)
- [x] `cargo test -p paimon btree::meta` (6 passed)
- [x] `cargo test -p paimon --all-targets` (1727 passed, 1 ignored)

## Notes

There is no public API or storage-format change. Valid legacy and versioned BTree metadata remain supported.

### DIFF
--- a/crates/paimon/src/btree/meta.rs
+++ b/crates/paimon/src/btree/meta.rs
@@ -196,7 +196,14 @@ impl BTreeIndexMeta {
             == 1;
         pos += 1;
 
-        if data.len().saturating_sub(pos) >= 2 {
+        let trailer_len = data.len().saturating_sub(pos);
+        if trailer_len == 1 {
+            return Err(invalid_meta(
+                "BTreeIndexMeta null flags trailer is truncated",
+            ));
+        }
+
+        if trailer_len >= 2 {
             let format_version = data[pos];
             pos += 1;
             if format_version == FORMAT_VERSION_WITH_NULL_FLAGS {
@@ -288,5 +295,15 @@ mod tests {
             let error = BTreeIndexMeta::deserialize(&encoded).unwrap_err();
             assert_eq!(error.kind(), io::ErrorKind::InvalidData);
         }
+    }
+
+    #[test]
+    fn test_meta_rejects_truncated_null_flags_trailer() {
+        let meta = BTreeIndexMeta::new(Some(Vec::new()), Some(Vec::new()), true);
+        let mut encoded = meta.serialize();
+        assert_eq!(encoded.pop(), Some(0));
+
+        let error = BTreeIndexMeta::deserialize(&encoded).unwrap_err();
+        assert_eq!(error.kind(), io::ErrorKind::InvalidData);
     }
 }
--- a/crates/paimon/src/btree/meta.rs
+++ b/crates/paimon/src/btree/meta.rs
@@ -32,6 +32,28 @@ const FORMAT_VERSION_WITH_NULL_FLAGS: u8 = 1;
 const FIRST_KEY_IS_NULL: u8 = 1;
 const LAST_KEY_IS_NULL: u8 = 1 << 1;
 
+fn invalid_meta(message: &'static str) -> io::Error {
+    io::Error::new(io::ErrorKind::InvalidData, message)
+}
+
+fn read_key(data: &[u8], pos: &mut usize) -> io::Result<Vec<u8>> {
+    let remaining = data
+        .get(*pos..)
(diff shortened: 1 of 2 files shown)


JingsongLi (changes requested):
Blocking: malformed global-index metadata is currently treated as an unavailable optimization and silently falls back to a normal table scan.

`BTreeIndexMeta::deserialize(bytes).ok()` discards the `InvalidData` error, and `evaluate_leaf` later returns `Ok(None)` when any shard has unavailable metadata. This makes a corrupted or incomplete declared BTree/Bitmap index indistinguishable from an unsupported predicate or an absent index. It also hides storage corruption and may unexpectedly turn an indexed query into a full scan without any warning.

The bounds checks added to `BTreeIndexMeta::deserialize` are correct: malformed external data must return a structured error instead of panicking. However, that error should be propagated as `Error::DataInvalid`, consistent with failures while opening or querying the physical index file.

Please consider changing `GlobalIndexScanner::create` to return `Result<Option<Self>>`, reserving `Ok(None)` for cases where no applicable index exists or the predicate cannot be evaluated. Missing required metadata and deserialization failures for a declared sorted index should return `Err`. The regression tests should assert an error rather than a successful fallback.

XiaoHongbo-Hope:
> Blocking: malformed global-index metadata is currently treated as an unavailable optimization and silently falls back to a normal table scan.
> 
> `BTreeIndexMeta::deserialize(bytes).ok()` discards the `InvalidData` error, and `evaluate_leaf` later returns `Ok(None)` when any shard has unavailable metadata. This makes a corrupted or incomplete declared BTree/Bitmap index indistinguishable from an unsupported predicate or an absent index. It also hides storage corruption and may unexpectedly turn an indexed query into a full scan without any warning.
> 
> The bounds checks added to `BTreeIndexMeta::deserialize` are correct: malformed external data must return a structured error instead of panicking. However, that error should be propagated as `Error::DataInvalid`, consistent with failures while opening or querying the physical index file.
> 
> Please consider changing `GlobalIndexScanner::create` to return `Result<Option<Self>>`, reserving `Ok(None)` for cases where no applicable index exists or the predicate cannot be evaluated. Missing required metadata and deserialization failures for a declared sorted index should return `Err`. The regression tests should assert an error rather than a successful fallback.

Fixed

JingsongLi (approved):
+1

JingsongLi (approved):
+1

PR #580, opened 2026-07-21
feat(table): read materialized deletion-vector PK splits

### LINKED ISSUE #579: Support materialized partial-update and aggregation reads with deletion vectors
### Motivation

Paimon Rust can plan deletion vectors and apply them in `DataFileReader`, but primary-key tables using `partial-update` or `aggregation` are always routed to `KeyValueFileReader`. That reader rejects splits carrying deletion vectors, so otherwise readable compacted plans fail with `KeyValueFileReader does not support deletion vectors`.

The requested scope is read-only compatibility for files whose merge result is already fully materialized.

### Solution

Route the whole plan through `DataFileReader` only when deletion vectors are enabled, `deletion-vectors.merge-on-read` is disabled, every split is raw-convertible, every file is above level 0, and every file has a known zero `delete_row_count`.

Keep failing closed for level-0 files, ambiguous metadata, mixed plans that still need per-key merging, merge-on-read, and deletion-vector writes.

### Anything else?

This is independent of #572 and #573.

### Willingness to contribute

- [x] I am willing to submit a PR!

shyjsarah:
### Purpose

Linked issue: close #579

Allow read-only access to fully materialized `partial-update` and `aggregation` primary-key plans with deletion vectors. This change is independent of #573.

### Brief change log

- Route eligible plans through `DataFileReader` so deletion vectors filter deleted rows.
- Require raw-convertible compacted files with known zero `delete_row_count` and merge-on-read disabled.
- Fail closed for level-0 files, ambiguous metadata, and plans that still require per-key merging.

### Tests

- Partial-update and aggregation compacted DV reads.
- Level-0, non-raw-convertible, unknown/nonzero delete-row count, and merge-on-read rejection.
- `cargo test -p paimon-datafusion --test pk_tables`

### API and Format

### Documentation

Updated the SQL merge-engine documentation with the supported read-only scope and limitations.


JingsongLi (approved):
+1

PR #582, opened 2026-07-22
fix(blob): validate entry integrity on reads

QuakeWang:
### Purpose

BLOB entries contain a magic number, embedded length, and CRC32, but inline reads previously fetched only payload bytes. Corrupted entries could therefore be returned as valid data.

### Brief change log

  - Validate entry magic, length, and CRC32 for scalar and inline `ARRAY<BLOB>` reads.
  - Reject short or corrupted entries, including empty BLOB values.
  - Preserve descriptor-mode range-only behavior and add corruption regression tests.

### Tests


JingsongLi (approved):
+1

PR #584, opened 2026-07-22
partial_update: support sequence-group reads

### LINKED ISSUE #583: feat(partial-update): support sequence-group reads
### Motivation

Java Paimon supports `fields.<sequence-field>.sequence-group=<protected-fields>` for `merge-engine=partial-update` tables, but Paimon Rust currently only supports basic partial-update mode and rejects every `fields.*.sequence-group` option at table-open time.

This prevents Paimon Rust / DataFusion from reading existing Java/Flink-created tables that use independent field-group sequencing. Falling back to the current basic latest-non-null merge would be incorrect because an older version from one update stream could overwrite fields protected by another sequence group.

I searched the existing issues and did not find an issue covering partial-update sequence-group support.

### Proposed scope

Add read-side sequence-group support for primary-key tables:

- parse single and composite sequence fields and their protected fields;
- validate referenced fields, duplicate protection, and primary-key conflicts;
- widen projected reads with the sequence fields required by projected protected fields;
- compare sequence tuples with Java-compatible null/ordering semantics;
- update multiple independent groups without cross-group overwrites;
- skip a group when all of its sequence fields are null;
- add Java-written / Rust-read compatibility tests where practical.

### Out of scope

To keep the contribution reviewable and fail-closed:

- field aggregation inside partial-update sequence groups;
- DELETE / UPDATE_BEFORE and remove-record semantics;
- Rust writer support for sequence-group tables.

The writer and Rust CREATE/ALTER paths should continue to reject sequence-group options until their semantics are implemented separately.

### Willingness to contribute

I'm willing to submit a PR for this scoped read-side support.

shyjsarah:
### Purpose

Linked issue: close #583

Paimon Rust currently rejects existing `merge-engine=partial-update` tables that use `fields.<sequence-fields>.sequence-group`. This change adds read-side support so Rust and DataFusion can read Java/Flink-created sequence-group tables without falling back to incorrect basic latest-non-null semantics.

### Brief change log

- Parse and validate single or composite sequence fields, protected fields, multiple independent groups, missing fields, primary-key conflicts, and repeated protected fields.
- Widen projected reads with the sequence fields required to merge projected or predicate-referenced protected fields, then remove those internal dependencies from the final projection.
- Merge each sequence group independently using Java-compatible ascending lexicographic comparison with nulls first.
  - Skip a group when its entire sequence tuple is null.
  - Let a newer or equal sequence tuple update the group.
  - Update sequence fields atomically and allow accepted protected fields to be set to null.
- Keep Rust CREATE/ALTER and writer paths fail-closed for sequence-group options.
- Continue rejecting sequence-group aggregation and remove-record/delete semantics, which are outside this PR.

### Tests

- `cargo test -p paimon --lib` — 1797 passed, 1 ignored.
- `cargo clippy -p paimon --lib --tests -- -D warnings`.
- `cargo test -p paimon-datafusion --test pk_tables` — 52 passed.
- Added coverage for independent groups, composite tuples, all-null tuple skipping, partial-null ordering, null protected-field replacement, projection dependency widening, malformed options, and end-to-end projected reads.

### API and Format

No public API or storage-format changes. This only expands read compatibility for existing partial-update tables. Rust table creation and writes remain unsupported for sequence groups.

### Documentation

Updated the SQL documentation with the supported read-side behavior and explicit unsupported scope.

### Review notes

The comparison behavior follows Java `UserDefinedSeqComparator.create(..., true)`: ascending field order and nulls first. Equal sequence tuples use the later row in the existing merge order. The implementation intentionally does not add ownership restrictions beyond primary-key and repeated protected-field validation used by the Java path.

### DIFF
--- a/crates/paimon/src/spec/partial_update.rs
+++ b/crates/paimon/src/spec/partial_update.rs
@@ -138,7 +138,7 @@ impl<'a> PartialUpdateConfig<'a> {
             .iter()
             .filter(|(key, _)| is_fields_option_with_suffix(key, SEQUENCE_GROUP_SUFFIX))
             .collect();
-        options.sort_by(|(lhs, _), (rhs, _)| lhs.cmp(rhs));
+        options.sort_by_key(|(key, _)| *key);
 
         options
             .into_iter()
--- a/crates/paimon/src/spec/partial_update.rs
+++ b/crates/paimon/src/spec/partial_update.rs
@@ -15,7 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+
+use crate::spec::DataField;
 
 const MERGE_ENGINE_OPTION: &str = "merge-engine";
 const PARTIAL_UPDATE_ENGINE: &str = "partial-update";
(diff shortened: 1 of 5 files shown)


JingsongLi (approved):
+1

PR #586, opened 2026-07-22
fix: use i64 for index file size

### LINKED ISSUE #585: IndexFileMeta truncates file sizes above 2 GiB
### Search before asking

### Please describe the bug

Rust declares IndexFileMeta.file_size as i32 even though the index-manifest Avro schema declares _FILE_SIZE as long and Java Paimon uses long. The fast Avro decoder reads the complete long and then casts it to i32.

As a result, Rust builders reject an individual index file larger than i32::MAX, while a Java-written manifest containing a larger value is silently truncated when Rust reads it. For example, 2,147,483,648 cannot be represented correctly.

Expected behavior: Rust should preserve the full signed 64-bit file size used by the storage schema and Java implementation.

### Solution

Use i64 for IndexFileMeta.file_size, remove the narrowing cast in the fast decoder, and use checked i64 conversions in all index-file writers. Keep the existing Avro long schema unchanged and add a boundary round-trip regression test.

- [x] I am willing to submit a PR.

XiaoHongbo-Hope:
### Purpose

Linked issue: close #585

Rust declared IndexFileMeta.file_size as i32 even though the index-manifest Avro schema and Java Paimon both use a 64-bit long. Rust builders therefore rejected individual index files larger than 2 GiB, while the fast decoder silently narrowed Java-written values.

This change preserves the full signed 64-bit file size across Rust read and write paths.

### Brief change log

- Widen IndexFileMeta.file_size and all index-file construction paths from i32 to i64.
- Remove the narrowing cast from the fast index-manifest decoder and clean up redundant downstream conversions.
- Add a regression that round-trips i32::MAX + 1 through Avro OCF and the fast decoder.
- Keep checked conversions from u64 and usize so values above i64::MAX still fail explicitly.

### Tests

- [x] cargo fmt --all -- --check
- [x] cargo test --locked -p paimon --lib — 1800 passed, 1 ignored
- [x] cargo test --locked -p paimon-datafusion --test system_tables test_table_indexes_system_table -- --exact
- [x] cargo clippy --locked -p paimon -p paimon-c -p paimon-datafusion --all-targets -- -D warnings

### API and Format

This changes the public Rust type of IndexFileMeta.file_size from i32 to i64, so downstream source code with explicit i32 construction must be updated.

The storage format does not change: _FILE_SIZE remains an Avro long. Existing manifests remain compatible, and Java can read values written by this change. Older Rust readers still cannot correctly read values above i32::MAX and should be upgraded.

_ROW_COUNT remains unchanged in this focused PR because widening it affects a broader set of row-count and vector-index APIs.

### Documentation

No documentation update is required.

### DIFF
--- a/bindings/c/src/tests.rs
+++ b/bindings/c/src/tests.rs
@@ -1764,7 +1764,7 @@ fn build_pk_vector_table(path: &str, vectors: &[[f32; PK_DIM]]) -> Table {
         let index_file = IndexFileMeta {
             index_type: INDEX_TYPE.to_string(),
             file_name: index_file_name,
-            file_size: i32::try_from(index_file_size).unwrap(),
+            file_size: i64::try_from(index_file_size).unwrap(),
             row_count: i32::try_from(row_count).unwrap(),
             deletion_vectors_ranges: None,
             global_index_meta: Some(GlobalIndexMeta {
--- a/crates/integrations/datafusion/src/system_tables/table_indexes.rs
+++ b/crates/integrations/datafusion/src/system_tables/table_indexes.rs
@@ -131,7 +131,7 @@ impl TableProvider for TableIndexesTable {
             buckets.push(entry.bucket);
             index_types.push(index_file.index_type.as_str());
             file_names.push(index_file.file_name.as_str());
-            file_sizes.push(i64::from(index_file.file_size));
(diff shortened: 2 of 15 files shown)


JingsongLi (approved):
+1

PR #587, opened 2026-07-22
fix(table): enable global index by default

XiaoHongbo-Hope:
## Summary

Align the Rust `global-index.enabled` default with Java Paimon and PyPaimon. When the option is absent, Rust scans can now use existing global indexes; an explicit `false` still disables them.

## Changes

- Default `global-index.enabled` to `true` when it is not configured.
- Cover the default and explicit `true` / `false` overrides at the option layer.
- Exercise the regular table-scan path without an explicit override and verify that it reads the committed BTree index row ranges.
- Update the SQL option reference with the corrected default.

## Testing

- `cargo test --locked -p paimon --lib` (1800 passed, 1 ignored)
- `cargo test --locked -p paimon --lib table::btree_global_index_build_builder::tests` (23 passed)
- Parent/head regression check: parent returns no row ranges; this PR returns `[0,0]` and `[2,2]`
- `cargo clippy --locked -p paimon --all-targets -- -D warnings`

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -540,7 +540,7 @@ impl<'a> CoreOptions<'a> {
         self.options
             .get(GLOBAL_INDEX_ENABLED_OPTION)
             .map(|value| value.eq_ignore_ascii_case("true"))
-            .unwrap_or(false)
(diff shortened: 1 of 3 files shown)


QuakeWang (approved):
+1

PR #590, opened 2026-07-22
perf(table): push row-range pruning into manifest scan

XiaoHongbo-Hope:
## Summary

Push row-range pruning into the data-manifest scan for data-evolution tables. Fast and Full global-index modes now determine selected row ranges before reading data manifests, so unrelated manifest files and entries can be skipped before merge, grouping, and statistics pruning.

## Changes

- Evaluate Fast and Full global-index lookups before data-manifest I/O; keep Detail mode's post-manifest fallback for unindexed ranges.
- Prune whole manifests by row-id bounds, then prune decoded entries before merge and data-evolution grouping.
- Apply the same row-range path to snapshot, delta, and changelog planning while preserving deletion-vector handling.
- Fail open when row-id metadata is missing or invalid, and retain downstream pruning as a correctness safeguard.
- Add scan-trace counters for manifest files and entries pruned by row ranges.
- Cover manifest-level pruning, entry-level pruning, Detail fallback, invalid metadata, and rolled BLOB/vector scan-to-read behavior.

## Testing

- [x] `cargo clippy --locked --release -p paimon --lib --tests --no-deps -- -D warnings`
- [x] `cargo test --locked --release -p paimon --lib` (1842 passed, 1 ignored)
- [x] Same-snapshot A/B on a real data-evolution table: median planning time improved from 2.203 s to 1.342 s with identical planned files and read results.

## Notes

Selected-range reads for rolled dedicated BLOB/vector files rely on the reader support added by #592.


JingsongLi:
Any filtering that can be done in advance should be done so, and should even be moved down to the manifest reading phase.

QuakeWang (approved):
LGTM

crates/paimon/src/table/table_scan.rs
```diff
@@ -1030,6 +1086,111 @@ impl<'a> PaimonTableScan<'a> {
...
+        let path = format!("{table_path}/{MANIFEST_DIR}/{index_manifest_name}");
+        Ok(Some(
+            IndexManifest::read(self.table.file_io(), &path).await?,
+        ))
+    }
+
+    async fn evaluate_global_index_row_ranges(
+        &self,
+        snapshot: &Snapshot,
+        index_entries: &[IndexManifestEntry],
```
JingsongLi, line 1342: Just for deletion-vectors? If it is. It needs to be in DV mode to be read. And it should be filtered so that only DV remains.

XiaoHongbo-Hope: > Just for deletion-vectors? If it is. It needs to be in DV mode to be read. And it should be filtered so that only DV remains.

Updated.

JingsongLi:
+1

crates/paimon/src/table/table_scan.rs
```diff
@@ -1224,21 +1657,67 @@ impl<'a> PaimonTableScan<'a> {
...
+        let index_entries = self
+            .read_index_manifest_entries(
+                &snapshot,
+                global_index_settings.is_some(),
+                core_options.deletion_vectors_enabled(),
+            )
+            .await?;
+        let manifest_row_ranges = self
+            .manifest_row_ranges(&snapshot, index_entries.as_deref(), global_index_settings)
+            .await?;
```
leaves12138, line 1679: Could we short-circuit an empty `manifest_row_ranges` here? When the global index returns `Some(vec![])`, a manifest/file with missing or invalid row-id metadata makes both row-range pruning helpers fail open. The file then reaches `split_row_ranges_for_files`, which returns `DataInvalid` because `row_id_range()` is `None`. As a result, a valid zero-hit Fast-mode lookup on a legacy data-evolution table fails instead of returning an empty plan. I reproduced this by planning with `Some(Vec::new())` against a committed file whose `first_row_id` is `None`; `plan()` returns `Cannot apply selected row ranges...`. Please return an empty plan before reading/planning data manifests when the selected range set is empty, and add a regression test.

leaves12138 (commented):
Found one correctness edge case in the empty row-range path.

PR #591, opened 2026-07-22
[umbrella] feat(datafusion): support REST format table partitions

sundapeng:
### Purpose

A Format Table loaded from REST Catalog can have its partitions managed by the catalog instead of discovered from the directory layout. This draft PR shows the complete Rust support for such tables in one place. **It is not meant to be merged**: every change reaches `main` through the child PRs below, and this branch follows the top of the child stack so the whole feature can be read and tried together.

### Child PRs

| # | PR | Scope | Status |
|---|----|-------|--------|
| 1 | #811 | Keep Format Table partition columns out of Parquet decoder filters | Merged in `7cb6f8a` |
| 2 | #813 | Skip committer staging files and list partition directories concurrently | Merged in `0496577` |
| 3 | #812 | REST partition registration, lookup and statistics APIs | Merged in `2bb2abb` |
| 4 | #814 | Read the registered partitions of a catalog-managed Format Table | Merged in `6824487` |
| 5 | #816 | `SHOW PARTITIONS`, `ALTER TABLE ... ADD / DROP PARTITION`, `Catalog::drop_partitions` | Merged in `cecd3dc` |
| 6 | #817 | `MSCK REPAIR TABLE`, partition directory discovery | Open, based on `main` |
| 7 | #815 | `ANALYZE TABLE ... COMPUTE STATISTICS [NOSCAN]` | Open, depends on #817 |

Review happens in the child PRs; comments here are welcome but will be carried over to the child that owns the code.

### Brief change log

Merged on `main`:

- `SHOW PARTITIONS table [PARTITION (...)]`, `ALTER TABLE table ADD [IF NOT EXISTS] PARTITION (...)` and `ALTER TABLE table DROP [IF EXISTS] PARTITION (...)`. DROP follows Java `resolveFormatTablePartitionsForDrop`: several specifications, partial specifications expanded to every registered match, values compared as the catalog holds them, registrations removed before directories, and a partition at a custom location only unregistered. Partition literals are read with the column type as Java `TypeUtils.castFromString` reads partition strings. `Catalog::drop_partitions`, which `RESTCatalog` refuses for tables without catalog-managed partitions (#816).
- REST Catalog APIs for paginated partition listing, create/drop registration (1000-entry batches), `list-by-names`, `list-by-filter`, and partition statistics reported with a create (`partitionStatistics` / `replaceStatistics`); a partition listed without statistics decodes them as unknown (-1) (#812).
- Eligible internal Format Tables load immutable REST partition metadata and scan only registered partitions. A scan sends its partition predicate to `list-by-filter` with a partition-name pattern built from leading CHAR/VARCHAR equalities, falls back to the paged listing when the catalog answers 501, and fails on a partition registered at a custom location (#814).
- Partition directories are listed concurrently, bounded by `format-table.scan.list-parallelism`, and entries starting with `.` or `_` below a partition directory, such as `_temporary` and `__magic_*`, are skipped (#813).
- A filter on a Format Table partition column that cannot become a partition predicate is no longer installed as a Parquet decoder filter, where it read the missing file column and returned no rows (#811).

In review:

- #817: `MSCK REPAIR TABLE table [{ADD|DROP|SYNC} PARTITIONS]` reconciles registrations with the directories that exist. Discovery and the catalog listing both finish before any change, DROP and SYNC fail unless the table root is a directory, and a partition at a custom location is never unregistered.
- #815: `ANALYZE TABLE table [PARTITION (...)] COMPUTE STATISTICS [NOSCAN]` measures registered partitions from storage and replaces the statistics the catalog holds. `NOSCAN` reports file count, size and last file creation time; a full run also reads Parquet and ORC footers for row counts. `format-table.statistics.parallelism` (default 8) bounds the work.

Following review on #816, each statement lives in its own module under `crates/integrations/datafusion/src/`: `format_partition_ddl.rs` (#816), `format_partition_repair.rs` (#817) and `format_partition_analyze.rs` (#815). `sql_context.rs` only dispatches them.

### Tests

Each child PR lists its tests. Commands run on the top of the stack (#815's head, the same tree as this branch):

```
cargo fmt --all -- --check
```

fmt and clippy pass. `paimon`: 2967 tests pass. `paimon-rest-server`: 10 tests pass. `paimon-datafusion`: 781 pass, including the 16 tests in `rest_format_partition_sql.rs`, and 39 fail. The failures are the 39 tests that also fail by name on `main` in this environment, because they read fixture tables provisioned by `make docker-up` (two also need the lumina native library).

Before the split, the branch was also validated end to end against a Bennett REST catalog server with a local file warehouse (10,000 registered partitions through `SHOW PARTITIONS`, ADD, MSCK ADD/SYNC, filtered reads, ANALYZE NOSCAN and full, and DROP). That run has not been repeated on the restructured stack.

Listed in each child PR: #816 (SQL literal conversion, the `SHOW PARTITIONS` parser, the `RESTCatalog::drop_partitions` refusal, the `Send` check on `SQLContext::sql`), #817 (repair safeguards) and #815 (footer row counts, in-process measurement).

### API and Format

Changes still in review:

- New `Catalog::drop_partitions`, with a default implementation that returns `Unsupported` (#816).
- `paimon::table::FormatTablePartitionPaths` (with `discover` from #817), `format_partition_value` and `parse_format_partition_value` become public (#816).
- New public `paimon::table::FormatTablePartitionStatsCollector` and option `format-table.statistics.parallelism` (#815).

The API changes that merged with #812 are described there. The storage format is unchanged.

### Documentation

`docs/src/sql.md` gains a Format Table Partitions section covering the `metastore.partitioned-table` prerequisite, `SHOW` / `ADD` / `DROP PARTITION`, `MSCK REPAIR TABLE`, `ANALYZE TABLE`, how partition values are read, and how a partition at a custom location is treated. Each child PR adds its own part.

### Scope and limitations

- Several `DROP PARTITION` specifications are written as repeated clauses, `DROP PARTITION (...), DROP PARTITION (...)`; sqlparser 0.62 cannot parse the Hive form `DROP PARTITION (...), (...)`.
- Reading a partition from its custom location is not implemented yet; such partitions fail closed.
- Partition row counts reported by ANALYZE are not yet used for DataFusion scan statistics (apache/paimon#9351 on the Java side).
- Writing Format Tables from the Rust client is still unsupported; `INSERT INTO` fails before anything is written or registered.
- A TIME partition value must be given as an integer. Partition columns typed DECIMAL, TIMESTAMP, FLOAT, DOUBLE or BINARY remain unsupported on this path.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -37,6 +37,7 @@
 //! - `ALTER TABLE db.t ADD [IF NOT EXISTS] PARTITION (...) [PARTITION (...)]`
 //! - `ALTER TABLE db.t DROP [IF EXISTS] PARTITION (...)`
 //! - `SHOW PARTITIONS db.t [PARTITION (...)]`
+//! - `MSCK REPAIR TABLE db.t [{ADD|DROP|SYNC} PARTITIONS]`
 //! - `CREATE VIEW [IF NOT EXISTS] view [(col, ...)] AS query`
 //! - `DROP VIEW [IF EXISTS] view`
 //! - `CREATE FUNCTION name(args) RETURNS type [LANGUAGE SQL] RETURN expression`
@@ -609,6 +610,7 @@ impl SQLContext {
                 self.handle_truncate_table(truncate, enable_ident_normalization)
                     .await
             }
+            Statement::Msck(msck) => crate::format_partition_repair::execute_msck(self, msck).await,
             Statement::CreateView(create_view) => {
                 if create_view.temporary {
                     // Temporary views are always handled by us (Paimon catalog temp storage)
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -436,7 +436,6 @@ impl SQLContext {
         }
     }
 
-    #[cfg(test)]
     pub(crate) fn dynamic_options(&self) -> &DynamicOptions {
         &self.dynamic_options
     }
@@ -611,6 +610,14 @@ impl SQLContext {
                     .await
             }
             Statement::Msck(msck) => crate::format_partition_repair::execute_msck(self, msck).await,
+            Statement::Analyze(analyze) => {
+                crate::format_partition_analyze::execute_analyze(
+                    self,
+                    analyze,
+                    enable_ident_normalization,
+                )
+                .await
+            }
             Statement::CreateView(create_view) => {
                 if create_view.temporary {
                     // Temporary views are always handled by us (Paimon catalog temp storage)
--- a/crates/paimon/src/table/format_table_scan.rs
+++ b/crates/paimon/src/table/format_table_scan.rs
@@ -107,29 +107,22 @@ impl<'a> FormatTableScan<'a> {
                 let partition_levels_below_root = partition_fields
                     .len()
                     .saturating_sub(root_segments.len().saturating_sub(table_depth));
-                let statuses = self
-                    .list_status_recursive_if_exists(&scan_root.path)
-                    .await?;
+                let files = list_format_table_data_files(
+                    self.table.file_io(),
+                    &scan_root.path,
+                    partition_levels_below_root,
+                    format_extension,
+                )
+                .await?;
                 let mut splits = Vec::new();
-                for status in statuses {
-                    if is_hidden_below_partitions(
-                        &root_segments,
-                        partition_levels_below_root,
-                        &status.path,
-                    ) {
-                        continue;
-                    }
-                    if let Some(split) = self
-                        .status_to_split(
-                            status,
-                            table_path,
-                            format_extension,
-                            schema_id,
-                            partition_fields,
-                            scan_root.partition.clone(),
-                        )
-                        .await?
-                    {
+                for status in files {
+                    if let Some(split) = self.status_to_split(
+                        status,
+                        table_path,
+                        schema_id,
+                        partition_fields,
+                        scan_root.partition.clone(),
+                    )? {
                         splits.push(split);
                     }
                 }
@@ -363,27 +356,11 @@ impl<'a> FormatTableScan<'a> {
         }
     }
 
-    async fn list_status_recursive_if_exists(
-        &self,
-        path: &str,
-    ) -> crate::Result<Vec<crate::io::FileStatus>> {
-        match self.table.file_io().list_status_recursive(path).await {
-            Ok(statuses) => Ok(statuses),
-            Err(err) => {
-                if !self.table.file_io().exists(path).await.unwrap_or(true) {
-                    Ok(Vec::new())
-                } else {
-                    Err(err)
-                }
-            }
-        }
-    }
-
-    async fn status_to_split(
+    /// The split reading one file that [`list_format_table_data_files`] returned.
+    fn status_to_split(
         &self,
         status: crate::io::FileStatus,
         table_path: &str,
-        format_extension: &str,
         schema_id: i64,
         partition_fields: &[DataField],
         known_partition: BinaryRow,
@@ -393,17 +370,6 @@ impl<'a> FormatTableScan<'a> {
         };
         let parent = parent.to_string();
         let file_name = file_name.to_string();
-        if !is_format_table_data_file_name(&file_name) {
-            return Ok(None);
-        }
-        if !file_name.to_ascii_lowercase().ends_with(format_extension) {
-            return Ok(None);
-        }
-        let status = if status.size == 0 {
-            self.table.file_io().get_status(&status.path).await?
-        } else {
-            status
-        };
         let file_size = i64::try_from(status.size).map_err(|_| crate::Error::DataInvalid {
             message: format!(
                 "Format table file '{}' is too large to fit in i64 metadata",
@@ -481,6 +447,51 @@ struct ScanRoot {
 
 fn is_format_table_data_file_name(file_name: &str) -> bool {
     !file_name.is_empty() && !file_name.starts_with('.') && !file_name.starts_with('_')
+}
+
+/// The data files a Format Table scan reads below `root`: files whose own name is not hidden and
+/// ends with the format's extension, outside any entry that [`is_hidden_below_partitions`] skips.
+/// `partition_levels_below_root` is how many partition levels still lie under `root`.
+///
+/// A root that does not exist holds no files. Any other listing failure is returned, since a
+/// partial listing cannot be told apart from a partition that lost files. `ANALYZE TABLE`
+/// measures a partition through this listing, so it counts exactly the files a scan reads.
+pub(crate) async fn list_format_table_data_files(
+    file_io: &crate::io::FileIO,
+    root: &str,
+    partition_levels_below_root: usize,
+    format_extension: &str,
+) -> crate::Result<Vec<crate::io::FileStatus>> {
+    let statuses = match file_io.list_status_recursive(root).await {
+        Ok(statuses) => statuses,
+        Err(error) => {
+            if !file_io.exists(root).await.unwrap_or(true) {
+                return Ok(Vec::new());
+            }
+            return Err(error);
+        }
+    };
+    let root_segments = path_segments(root);
+    let mut files = Vec::with_capacity(statuses.len());
+    for status in statuses {
+        if is_hidden_below_partitions(&root_segments, partition_levels_below_root, &status.path) {
+            continue;
+        }
+        let is_data_file = split_parent_and_file(&status.path).is_some_and(|(_, file_name)| {
+            is_format_table_data_file_name(file_name)
+                && file_name.to_ascii_lowercase().ends_with(format_extension)
+        });
+        if !is_data_file {
+            continue;
+        }
+        let status = if status.size == 0 {
+            file_io.get_status(&status.path).await?
+        } else {
+            status
+        };
+        files.push(status);
+    }
+    Ok(files)
 }
 
 /// Whether a listed file is, or lies inside, an entry whose name starts with `.` or `_` below
@@ -843,7 +854,7 @@ fn supported_format_table_formats() -> Vec<&'static str> {
     ]
 }
 
-fn supported_format_table_extension(format: &str) -> crate::Result<&'static str> {
+pub(crate) fn supported_format_table_extension(format: &str) -> crate::Result<&'static str> {
     match format.to_ascii_lowercase().as_str() {
         "parquet" => Ok(".parquet"),
         "orc" => Ok(".orc"),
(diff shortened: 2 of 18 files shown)


JingsongLi:
1. **High: DROP PARTITION is inconsistent with Java semantics**
   - Rust only supports single, complete partitions: `sql_context.rs:1064` rejects multiple specs, and `sql_context.rs:1734` uses `require_complete=true`.
   - Java supports dropping multiple partitions at once and also supports arbitrary partial specifications such as `dt=...` and `hh=...`, which are expanded to include all matching leaf partitions.
   - Impact: Commands executable in Java, such as `DROP PARTITION (dt=‘20260715’)` and batch DROP operations, fail outright in Rust. Recommendation: Follow Java’s approach—pre-check complete specs using `list-by-names`; expand partial specs after a single catalog traversal.

2. **High: Catalog-managed scans lack partition pruning via the REST endpoint**
   - Rust’s `format_table_scan.rs:216` unconditionally uses `list_partitions` to retrieve all partitions in the table before executing the predicate locally.
   - Java `FormatTableScan.java:230` extracts the leading equality prefix and the full predicate; `CatalogFormatTablePartitionManager.java:68` pushes the pattern/predicate and page size (1000) down to REST.
   - Even when querying a single partition in a table with many partitions, all partition metadata is downloaded and parsed, which may cause significant latency and memory issues.

3. **Note: Boolean partition values are incompatible with Java**
   - Repair preserves and registers the original values in the catalog, but Rust `format_partition.rs:216` uses `str::parse::<bool>()`, which only accepts lowercase `true/false`.
   - Java ignores case and accepts `t/y/yes/1` and `f/n/no/0`.
   - Therefore, if a catalog entry `active=TRUE` is registered via Rust MSCK, subsequent `scan` or `SHOW` operations will report “invalid catalog partition metadata”; Java can read it normally. It is recommended to adopt the Java-compatible rules and perform additional cross-platform testing.

sundapeng:
@JingsongLi Thanks for the review. All three are fixed, and the branch is rebased onto current main (it was conflicting with #600 and #627). Details below, including one place where I could not go all the way and would like your call.

**1. DROP PARTITION semantics**

`DROP PARTITION` now takes several specifications per statement, and a specification that fixes only some of the partition keys expands to every registered partition it matches. The keys need not be a leading prefix, so `DROP PARTITION (hh = '10')` on a `(dt, hh)` table works, matching `PaimonFormatTable.dropFormatTablePartitions` where partial specs are matched by arbitrary key subset. One catalog listing serves the whole statement, however many specifications it carries.

Error semantics follow Java as well: a complete specification names one partition, so a missing one raises unless `IF EXISTS`; a partial one describes a set that is allowed to come out empty, so it is a no-op. A specification that raises leaves the whole statement unapplied.

One thing I could not do. Hive and Spark write several specs as `DROP PARTITION (a), (b)`, and sqlparser 0.62 cannot parse that: `parse_alter_table` comma-separates ALTER *operations*, and a bare second `PARTITION` is read as `RENAME PARTITION`. So multiple specs are written as repeated clauses:

```sql
ALTER TABLE t DROP PARTITION (dt = '20260722'), DROP PARTITION (dt = '20260723');
```

The alternatives are a change in sqlparser, or a hand-rolled pre-parse of the statement like the one `SHOW PARTITIONS` already uses here. I did not add the pre-parse because it duplicates the ALTER TABLE grammar for one syntax variant, but I will add it if you want the Hive spelling accepted.

I also did not add `list-by-names`. Java resolves complete specs through it and asserts in tests that the complete-spec path never does a full traversal, but paimon-rust has no `listPartitionsByNames` anywhere yet (no `Catalog` method, no `RESTApi` method, no resource path, no request type). Adding the whole stack here would grow a PR that is already large. The current code needs the registered spec anyway to resolve the directory, so it costs one listing per statement rather than one per specification. Happy to do it as a follow-up, or here if you prefer.

**2. Partition pruning through the REST endpoint**

A catalog-managed scan now extracts the leading equality prefix from its filter, builds the partition-name prefix pattern with the same contract as `PartitionPathUtils.buildPartitionNamePrefixPattern` (escaped `key=value` joined by `/`, `%` the only wildcard, complete prefix means the exact name, shorter prefix gets `/%`, no pattern when a value is blank or escaping produced a literal `%`), and sends it as `partitionNamePattern` with `maxResults=1000`. The local per-partition match stays, so a catalog that ignores the pattern still produces the same result set.

One Rust-specific detail worth flagging. A straight port of the Java extractor would have pushed nothing in the case that matters most. `PartitionFilter::from_predicate` collapses a filter that pins every partition key into a `PartitionSet` and drops the predicate, so `WHERE dt = 'a' AND hh = '10'` never reaches the scan as a `Predicate`. The extractor therefore also handles `PartitionSet`, taking the longest common leading prefix of its rows. A single-partition set gives the exact name, and `dt = 'a' AND hh IN ('10', '11')` gives `dt=a/%`.

Predicate pushdown to `listPartitionsByFilter` is not included: #500 added REST predicate JSON parsing but not serialization, so there is nothing to encode with yet.

**3. Boolean partition values**

`parse_format_partition_value` now accepts `t/true/y/yes/1` and `f/false/n/no/0` case-insensitively, mirroring `BinaryStringUtils.toBoolean`. There was a unit test asserting that `yes` is rejected, which is exactly the bug, so it is replaced with a table covering every spelling in both directions plus the rejections.

While confirming this I found the same class of problem is wider than booleans: `parse_format_partition_value` covers BOOLEAN, the integer types, CHAR/VARCHAR, DATE and TIME, while `TypeUtils.castFromString` also covers DECIMAL, FLOAT, DOUBLE, TIMESTAMP, TIMESTAMP_LTZ and BINARY. A catalog-managed table partitioned by any of those cannot be scanned. That is not a regression from this PR, and closing it properly means aligning both directions of the conversion, which are currently split across two modules that have already drifted. I would rather do it in its own PR than widen this one. Tell me if you want it here instead.

**Also in this push**

`docs/src/sql.md` documents all four statements, and the PR description now follows the template.

JingsongLi (commented):
Compared this change with the Apache Paimon Java implementation. I found two correctness issues and four Java-parity/performance gaps; details are inline.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1833,6 +1908,262 @@ impl SQLContext {
...
+            // so catalog values that only differ before normalization stay distinguishable.
+            let verbatim = registered.iter().any(|(registered, _)| registered == spec);
+            let mut matches = 0usize;
+            for (registered_spec, registered_normalized) in &registered {
+                let selects = if verbatim {
+                    registered_spec == spec
+                } else {
+                    normalized
+                        .iter()
+                        .all(|(key, value)| registered_normalized.get(key) == Some(value))
```
JingsongLi, line 1968: [P1] Preserve raw partition identity during partial DROP

Partial specifications are matched after converting catalog values to typed Datum values. A table partitioned by (year STRING, month INT) can legitimately contain raw registrations such as {year=2025, month=01} and {year=2026, month=1} after MSCK, because repair preserves directory spelling. DROP PARTITION (month=1) has no verbatim match because the request is partial, so both values normalize to Int(1) and both directories are unregistered and deleted. Java compares partial specs using the raw catalog values, deleting only month=1. Please match partial specs against raw strings and add a mixed-spelling regression test.

crates/paimon/src/table/format_table_scan.rs
```diff
@@ -178,19 +205,81 @@ impl<'a> FormatTableScan<'a> {
...
+        // and the local match below still decides what is actually scanned.
+        let pattern = match &self.partition_filter {
+            Some(filter) => {
+                let leading_values = leading_equality_partition_values(
+                    filter,
+                    partition_fields,
+                    default_partition_name,
+                    core_options.legacy_partition_name(),
+                )?;
+                partition_paths.name_prefix_pattern(&leading_values)
```
JingsongLi, line 232: [P1] Avoid lossy name-pattern pushdown for typed partitions

This formats typed equality literals canonically before building the raw partitionNamePattern. MSCK preserves valid raw spellings such as active=TRUE or month=01, but predicates active = true / month = 1 produce exact patterns active=true / month=1. A catalog that honors the pattern removes those registrations before the local typed check runs, so the query silently loses rows. Please restrict name-pattern pushdown to types with a unique raw spelling (at least CHAR/VARCHAR), or use a typed-filter/list-all fallback, and add raw TRUE/01 scan tests.

crates/paimon/src/catalog/rest/rest_catalog.rs
```diff
@@ -373,16 +375,64 @@ impl Catalog for RESTCatalog {
...
+
+    async fn drop_partitions(
+        &self,
+        identifier: &Identifier,
+        partition_specs: Vec<HashMap<String, String>>,
+    ) -> Result<()> {
+        if partition_specs.is_empty() {
+            return Ok(());
+        }
+        for batch in partition_specs.chunks(PARTITION_BATCH_SIZE) {
```
JingsongLi, line 412: [P2] Preserve ordinary-table DROP semantics

RESTCatalog::drop_partitions always calls the metadata-only REST endpoint. Java only does that for catalog-managed Format Tables; for ordinary Paimon tables it loads the table and commits truncatePartitions. A public Rust Catalog caller can therefore get an error or success without the expected partition-data change. Please branch on has_catalog_managed_partitions(): keep REST unregistering for managed Format Tables and use a table commit for ordinary tables, or explicitly reject non-managed calls until implemented.

crates/paimon/src/table/format_table_scan.rs
```diff
@@ -178,19 +205,81 @@ impl<'a> FormatTableScan<'a> {
...
+            }
+            None => None,
+        };
+        let partitions = rest_env
+            .api()
+            .list_partitions_by_name_pattern(rest_env.identifier(), pattern.as_deref())
```
JingsongLi, line 238: [P2] Push the complete partition predicate to REST

Only a leading-equality name pattern is sent. Filters without such a prefix (for example, hour = 10 for keys (dt, hour), ranges, or starts-with) download the whole registry in 1,000-row pages before local filtering. Java uses paged /partitions/list-by-filter with the serialized predicate and still rechecks locally. Please add the compatible request/path/API with a safe fallback.

crates/paimon/src/table/format_table_scan.rs
```diff
@@ -65,17 +71,25 @@ impl<'a> FormatTableScan<'a> {
...
+            .map(|options| options.table_path.as_str())
+            .or_else(|| core_options.path())
             .unwrap_or_else(|| self.table.location())
             .trim_end_matches('/')
             .to_string();
 
         let partition_fields = self.table.schema().partition_fields();
         let mut splits = Vec::new();
-        for scan_root in self.scan_roots(&core_options, &table_path)? {
+        for scan_root in self.scan_roots(&core_options, &table_path).await? {
```
JingsongLi, line 92: [P2] Bound partition-directory listing concurrency

This awaits one recursive object-store listing per registered partition inside a plain loop. An unfiltered 10k-partition table therefore pays roughly 10k list latencies serially. Java uses configurable bounded concurrency through format-table.scan.list-parallelism (default 64). Please use bounded async concurrency while preserving deterministic final sorting and fail-fast behavior.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -1833,6 +1908,262 @@ impl SQLContext {
...
+            requested.push((
+                spec.len() == partition_key_count,
+                spec,
+                normalized,
+                *ignore_if_not_exists,
+            ));
+        }
+
+        let registered = catalog
+            .list_partitions(identifier)
```
JingsongLi, line 1939: [P2] Use list-by-names for complete DROP specs

Even a complete exact DROP PARTITION calls list_partitions and materializes every registration. Java resolves complete specs through batched /partitions/list-by-names and reserves a full traversal for partial specs. Please implement the corresponding REST/Catalog method and use it for complete requests; this avoids O(all partitions) network and memory for the common exact-drop case.

sundapeng:
@JingsongLi Thanks for the second pass. Rebased onto current main; all six comments are addressed:

- **[P1] Raw identity in DROP**: specs are compared with the values the catalog holds, as Java `resolveFormatTablePartitionsForDrop` does, so `DROP PARTITION (month = 1)` no longer takes `month=01`.
- **[P1] Lossy pattern pushdown**: the name pattern is built only from leading CHAR/VARCHAR equalities. Java still pushes typed values here; happy to port the restriction if you agree.
- **[P2] Ordinary-table DROP**: `RESTCatalog::drop_partitions` now refuses tables without catalog-managed partitions.
- **[P2] Full predicate pushdown**: added `Predicate::to_rest_json` and `list-by-filter`. On 501 the scan retries the paged pattern listing (still the catalog, never the filesystem), where Java throws.
- **[P2] Listing concurrency**: bounded by `format-table.scan.list-parallelism` (default 64, clamped to [1, 1000]).
- **[P2] list-by-names**: used when every DROP spec is complete, batched at 1000. Java master now reads the registry once instead; the result is the same.

Also in this push:

- `fix(datafusion): keep format table partition columns out of decoder filters`: since #574, a filter such as `WHERE active` on a partition column returned no rows. Can split it out if you prefer.
- `fix(datafusion): leave partitions at a custom location where they are`: such partitions fail closed until location support lands.
- `feat(datafusion): support ANALYZE TABLE on catalog-managed format tables`, following the Java command, with docs.

Details and the end-to-end validation are in the PR description.

sundapeng:
@JingsongLi This PR has grown too large to review comfortably, so I am splitting it. The first three PRs each apply to main on their own:

1. a fix that keeps format table partition columns out of Parquet decoder filters;
2. a format table listing fix: skip committer staging directories and list partitions concurrently;
3. the REST partition catalog APIs: list-by-names, list-by-filter, statistics on create, and `Predicate::to_rest_json`.

The catalog-managed scan, the partition commands and ANALYZE follow once those land, and this PR will be narrowed to the partition commands. I will link the new PRs here.

sundapeng:
Moving this back to draft: it stays as the full view of the feature, and the remaining changes land through smaller child PRs (partition DDL, MSCK REPAIR, and #815 for ANALYZE). I will link them here.

sundapeng:
Child PRs: #816 (partition DDL), #817 (MSCK REPAIR), #815 (ANALYZE).

PR #592, opened 2026-07-23
feat(table): support selected ranges for rolled dedicated files

XiaoHongbo-Hope:
## Summary

Allow data-evolution reads with row ranges to consume partially retained rolled BLOB and vector segments. Full reads keep the existing complete-bunch validation.

## Changes

- Relax complete BLOB coverage and vector row-count checks only when a split carries row ranges.
- Allow gaps between selected vector segments and skip non-overlapping vector files before opening them.
- Keep higher-sequence partial overlaps fail-closed to avoid dropping uncovered rows.
- Cover non-first rolled BLOB reads in both value and descriptor modes, sparse vector segments, skipped vector I/O, and overlap validation.

## Testing

- [x] `cargo clippy --locked -p paimon --lib --tests -- -D warnings`
- [x] `cargo test -p paimon --lib` (1816 passed, 1 ignored)

## Notes

This PR is independent of #590. It enables a follow-up change to prune rolled dedicated entries during manifest reading.

### DIFF
--- a/crates/paimon/src/table/data_evolution_reader.rs
+++ b/crates/paimon/src/table/data_evolution_reader.rs
@@ -1135,7 +1135,28 @@ fn open_source_stream(
     blob_as_descriptor: bool,
     anchor_deletion_vector: Option<&DeletionVectorContext>,
 ) -> crate::Result<ArrowRecordBatchStream> {
+    let mut row_ranges = row_ranges;
     if let FieldSource::BlobBunch { bunch, read_fields } = source {
+        let selected_ranges = selected_absolute_row_ranges_for_file(
+            bunch.expected_first_row_id,
+            bunch.expected_row_count,
+            row_ranges.as_deref(),
+            anchor_deletion_vector.map(|context| context.deletion_vector.as_ref()),
+        )?;
+        if let Some(selected_ranges) = selected_ranges.as_deref() {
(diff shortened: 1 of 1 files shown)


crates/paimon/src/table/data_evolution_reader.rs
```diff
@@ -1712,7 +1743,7 @@ fn build_source_plan(
         } = source
         {
             bunch.finalize()?;
-            if !read_fields.is_empty() {
+            if !read_fields.is_empty() && !row_id_pushdown {
                 bunch.validate_logical_range()?;
```
QuakeWang, line 1823: When row ID pushdown is enabled, this skips BLOB coverage validation entirely. For a multi-sequence BLOB bunch, an uncovered selected row remains a placeholder and is ultimately returned as NULL, silently producing an incorrect result for an incomplete or incorrectly pruned split.

XiaoHongbo-Hope: > When row ID pushdown is enabled, this skips BLOB coverage validation entirely. For a multi-sequence BLOB bunch, an uncovered selected row remains a placeholder and is ultimately returned as NULL, silently producing an incorrect result for an incomplete or incorrectly pruned split.

Fixed

QuakeWang (approved):
+1

PR #571, opened 2026-07-21
feat(fulltext): read global indexes via paimon-ftindex-core

shyjsarah:
### Purpose

Migrate the append/data-evolution global full-text search path from the legacy in-repository Tantivy implementation to the shared `paimon-ftindex-core` engine.

This lets Rust and DataFusion read the current `full-text` index archives produced by Java Paimon and PyPaimon while preserving the existing `full_text_search(table_name, column_name, query_text, limit)` SQL API.

This PR intentionally focuses on the non-primary-key/global-index path. The primary-key full-text and hybrid search work is covered separately by #568.

### Brief change log

- Replace the legacy Tantivy archive reader and writer modules with an adapter over `paimon-ftindex-core`.
- Read `full-text` entries from snapshot index manifests through a bounded-concurrency `SeekRead` adapter.
- Support plain-text queries, the Paimon JSON query DSL, Roaring include filters, shard-local/global row-ID translation, and global Top-K merging.
- Align `fast`, `full`, and `detail` search modes with Java and PyPaimon behavior, including raw fallback for uncovered row ranges.
- Filter deleted rows through data-evolution deletion vectors and keep raw fallback scoring in a single BM25 corpus.
- Add core and DataFusion coverage for native index reads, row-range filtering, raw fallback, query normalization, authorization, and branch table queries.

### API and compatibility

- The `full_text_search(table_name, column_name, query_text, limit)` SQL table-function signature is unchanged.
- The feature-gated Rust API changes: `FullTextSearch` and `SearchResult` move from `paimon::tantivy::full_text_search` to `paimon::full_text`, while the legacy `TantivyFullTextReader` and `TantivyFullTextWriter` APIs are removed.
- Plain-text queries remain supported. Callers using advanced Tantivy `QueryParser` syntax must migrate those queries to the Paimon JSON query DSL.
- Ordinary Paimon table data files are unchanged.
- The `fulltext` feature now reads the current Paimon `full-text` archive format. Legacy `tantivy-fulltext` auxiliary indexes are no longer supported and must be rebuilt in the current format.

### Testing

- [x] `cargo check -p paimon --features fulltext`
- [x] `cargo test -p paimon --features fulltext full_text -- --nocapture`
- [x] `cargo check -p paimon-datafusion --features fulltext`
- [x] `cargo test -p paimon-datafusion --features fulltext fulltext_tests -- --nocapture`
- [x] `cargo clippy -p paimon --lib --tests --features fulltext -- -D warnings`
[...]

PR #593, opened 2026-07-23
build: correct Jieba release license metadata

shyjsarah:
### Purpose

Linked issue: N/A (CI hotfix)

Fix release-license generation for Python artifacts after `paimon-ftindex-core` introduced the transitive `jieba-rs` and `jieba-macros` dependencies. Their crates.io packages do not include the workspace license file, so `scripts/release_licenses.py --check` rejects the unresolved generic MIT placeholder.

This extracts the focused legal-metadata fix from #591 and unblocks #571.

### Brief change log

- Add the verified upstream `jieba-rs` workspace MIT license and document its pinned source.
- Add a repository-backed license correction for `jieba-rs` and `jieba-macros`.
- Allow license corrections to cover multiple resolved versions of the same crate while requiring identical license text.
- Add regression coverage for the Jieba correction, repository-backed licenses, duplicate crate versions, and legal-file registration.

### Tests

- [x] `uv run --python 3.11 --no-project python -m unittest discover -s scripts/tests -p "test_*.py"`
- [x] `cargo fetch --locked`
- [x] `python3 scripts/release_licenses.py --check` — validated 21 generated binary legal files
- [x] `git diff --check`

### API and Format

### Documentation

Update `third-party-licenses/README.md` with the pinned upstream source for the Jieba workspace license.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -28,6 +28,7 @@ header:
     - ".github/PULL_REQUEST_TEMPLATE.md"
     - "crates/paimon/tests/**/*.json"
     - "crates/paimon/testdata/**"
+    - "third-party-licenses/jieba-rs-0.10.3.LICENSE"
     - "third-party-licenses/openssl-1.1.1.LICENSE"
     - "**/go.sum"
     - "**/DEPENDENCIES.*.tsv"
--- a/scripts/release_licenses.py
+++ b/scripts/release_licenses.py
@@ -62,6 +62,7 @@ class LicenseCorrection:
     license_name: str
     anchor: str
     license_from_repository: bool = False
+    expected_versions: tuple[tuple[str, str], ...] = ()
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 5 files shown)


JingsongLi (commented):
- `scripts/release_licenses.py:125` refers explicitly to `jieba-rs-0.10.3.LICENSE`, and the README also states that it only covers `jieba-rs/jieba-macros 0.10.3`; however, `correction_html()` in `scripts/release_licenses.py:387` directly uses this file for any version it resolves.
- When I constructed metadata using `9.9.9`, the script still successfully generated and applied the 0.10.3 license. In the future, when dependencies are upgraded, the license check will not require re-verification, potentially resulting in the silent generation of incorrect legal metadata.
- It is recommended to add expected versions for each crate to `LicenseCorrection` and strictly validate them in repository-backed corrections; see the existing `BundledComponent.crate_version` for reference. ([github.com](https://github.com/apache/paimon-rust/pull/593))

shyjsarah:
Thanks for catching this. I’ve added explicit expected versions for both `jieba-rs` and `jieba-macros` to `LicenseCorrection`. Repository-backed corrections now validate every resolved crate version before applying the checked-in license and fail explicitly on any mismatch.

I also added a regression test using version `9.9.9`, while preserving the existing multi-version behavior for corrections that read licenses from resolved crate sources. Could you please take another look?

JingsongLi (approved):
+1

PR #597, opened 2026-07-23
perf(parquet): prune row selection with OffsetIndex

XiaoHongbo-Hope:
### Purpose

A row selection without predicates did not load the Parquet OffsetIndex. Arrow-rs could filter rows, but still fetched unselected data pages, including pages for nested columns.

This is independent of #590: that PR prunes files, while this change prunes pages inside a retained Parquet file. It also improves selected-range reads introduced by #592.

### Changes

- Load ColumnIndex for predicate page pruning.
- Load OffsetIndex for predicates or row selections.
- Add a nested List regression test that verifies sparse selection reduces page I/O.

### Tests

- cargo fmt --all -- --check
- Parquet module tests: 41 passed
- paimon library tests: 1,839 passed, 1 ignored
- clippy with warnings denied

### DIFF
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -310,17 +310,14 @@ impl FormatFileReader for ParquetFormatReader {
         };
         let row_filter_factory = predicates.and_then(|fp| fp.row_filter_factory.as_deref());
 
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
LGTM

PR #598, opened 2026-07-23
partial_update: support field aggregation reads

### LINKED ISSUE #596: feat(partial-update): support field aggregation reads
### Motivation

Paimon Rust can now read `merge-engine=partial-update` tables with sequence groups after #584, but it still rejects existing Java/Flink-created tables that configure:

- `fields.<field>.aggregate-function`
- `fields.default-aggregate-function`

This prevents Rust and DataFusion readers from opening partial-update tables that aggregate protected fields within sequence groups. Treating these tables as basic latest-non-null partial update would be incorrect, especially when an older sequence-group row must be combined in reverse order.

I searched the existing issues and did not find one covering partial-update field aggregation reads. This is a follow-up to #583 / #584.

### Proposed scope

Add read-side field aggregation support for partial-update tables:

- parse and validate per-field and default aggregate-function options;
- reuse the aggregate functions already implemented in Paimon Rust;
- support `last_non_null_value` without a sequence group, matching Java validation;
- require other configured aggregate functions to protect fields in a sequence group;
- apply normal aggregation when an incoming group sequence is newer or equal;
- apply Java-compatible reversed aggregation when an incoming group sequence is older;
- preserve sequence-group projection dependency widening;
- fail closed with a clear error for unknown or type-incompatible aggregate functions;
- add merge-level and end-to-end projected-read tests.

### Out of scope

To keep the contribution reviewable:

- adding Java-only aggregate-function families that are not currently implemented in Rust;
- DELETE / UPDATE_BEFORE retraction and remove-record semantics;
- Rust writer and CREATE/ALTER support for partial-update aggregation.

### Willingness to contribute

I'm willing to submit a PR for this scoped read-side support.

shyjsarah:
## Summary

Add read-side support for partial-update field aggregation, following the sequence-group read support from #584.

The reader now parses and validates `fields.<field>.aggregate-function`, `fields.default-aggregate-function`, and `fields.<field>.list-agg-delimiter`, then applies the configured aggregators while merging rows. Older sequence-group rows use Java-compatible reversed aggregation so order-sensitive functions produce the same result as Paimon Java.

Fixes #596.

## Changes

- validate aggregate fields, functions, types, sequence-group coverage, and supported options before reading;
- support the aggregate functions already registered in Rust, including per-field/default precedence;
- add reversed aggregation semantics for numeric, boolean, value-picking, and `listagg` aggregators;
- preserve all-null and partial-null composite sequence behavior;
- widen internal projections for sequence and residual-filter dependencies without leaking internal columns;
- reset aggregation state between primary keys and keep the non-aggregation fast path;
- fail closed for unsupported functions and advanced modifiers;
- document partial-update field aggregation read behavior and limitations.

This is read-side only. Writer/CREATE support and delete/retract aggregation remain out of scope. There are no public API changes.

## Testing

- `cargo test --locked -p paimon-datafusion --test pk_tables`

Added coverage for validation, normal/reversed aggregation order, equal sequences, all-null and partial-null composite sequences, projection/filter dependency widening, multiple primary keys, state reset, `listagg`, NaN tie behavior, and Decimal precision/backing-integer overflow recovery.

### DIFF
--- a/crates/integrations/datafusion/tests/pk_tables.rs
+++ b/crates/integrations/datafusion/tests/pk_tables.rs
@@ -38,6 +38,8 @@ use datafusion::arrow::datatypes::{
 };
 use paimon::catalog::Identifier;
 use paimon::Catalog;
+use paimon_datafusion::PaimonTableProvider;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 // ======================= Basic PK Write + Read =======================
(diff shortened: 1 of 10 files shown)


JingsongLi (approved):
+1

PR #601, opened 2026-07-23
fix(table): align bitmap index ordering with Java

QuakeWang:
### Purpose

Bitmap lookup mixed serialized byte ordering with logical key ordering. This could select the wrong dictionary block for little-endian numeric keys. Floating-point handling also did not follow Java's canonical NaN contract, making Eq/In candidates and metadata pruning unnecessarily conservative.

### Brief change log

  - Use Java-compatible logical ordering and canonical NaN serialization for Bitmap indexes.
  - Keep Float/Double Eq/In as direct lookups while failing open for residual-sensitive predicates.
  - Restore safe Bitmap metadata pruning without changing existing BTree behavior.
  - Add Java logical-order fixtures and floating-point regression coverage.

### Tests

  - `git diff --check`


JingsongLi (approved):
+1

PR #604, opened 2026-07-23
fix: serialize empty BinaryRow with fixed part

### LINKED ISSUE #603: BinaryRow::new(0) emits truncated bytes in native DataSplit serialization
### Search before asking

### Please describe the bug 🐞

`BinaryRow::new(0).to_serialized_bytes()` currently emits only the 4-byte big-endian arity:

```text
00 00 00 00
```

This is a truncated Paimon `BinaryRow`. The Java reference implementation initializes `BinaryRow.EMPTY_ROW` with its 8-byte fixed part, so `SerializationUtils.serializeBinaryRow(BinaryRow.EMPTY_ROW)` emits 12 bytes:

```text
00 00 00 00 00 00 00 00 00 00 00 00
|-- arity --| |-------- fixed part --------|
```

The problem is exposed by the native `DataSplit` codec added in #565. `DataSplit::serialize()` calls `partition.to_serialized_bytes()`, while `DataSplit::deserialize()` calls the strict `BinaryRow::from_serialized_bytes()` decoder. A non-partitioned split built with the commonly used `BinaryRow::new(0)` therefore cannot deserialize its own output:

```rust
let split = DataSplit::builder()
    .with_snapshot(1)
    .with_partition(BinaryRow::new(0))
    .with_bucket(0)
    .with_bucket_path("file:/tmp/bucket-0".to_string())
    .with_total_buckets(1)
    .with_data_files(vec![])
    .build()
    .unwrap();

let bytes = split.serialize().unwrap();
DataSplit::deserialize(&bytes).unwrap();
```

The last line fails with:

```text
BinaryRow: serialized body too short for arity 0: 0 bytes, need at least 8
```

The existing history confirms that the serializer, rather than the decoder, is wrong:

- #349 established that a Java-compatible empty `BinaryRow` contains the 4-byte arity and the 8-byte null bitmap/fixed part, and introduced `EMPTY_SERIALIZED_ROW`.
- #364 correctly made `BinaryRow::from_serialized_bytes()` reject a truncated body. That validation should not be weakened.
- #565 added native `DataSplit` deserialization and non-empty Java goldens, but does not cover a split whose partition is `BinaryRow::new(0)`.

I also verified the Java side directly with the Maven Central `org.apache.paimon:paimon-bundle:1.4.2` artifact:

- `SerializationUtils.serializeBinaryRow(BinaryRow.EMPTY_ROW)` returns 12 bytes.
- Java `DataSplit#serialize` / `DataSplit#deserialize` round-trips an empty-partition split.
- The Java-generated empty-partition DataSplit is 77 bytes.

Expected behavior: serializing the canonical Rust empty row should produce the same 12-byte representation as Java, and an empty-partition native DataSplit should deserialize and reserialize successfully.

### Solution

Normalize only the canonical in-memory stub (`arity == 0 && data.is_empty()`) in `BinaryRow::to_serialized_bytes()` by returning the existing `EMPTY_SERIALIZED_ROW`.

This keeps the strict decoder from #364 unchanged, preserves normal populated-row serialization, and restores compatibility with Java without changing the DataSplit wire version.

Add regression tests for:

- the exact 12-byte Java wire representation of `BinaryRow::new(0)`;
- an empty-partition DataSplit against bytes generated by Java Paimon 1.4.2;
- deserialize and byte-stable reserialize of that split.

yanbinyang:
### Purpose

Linked issue: close #603

`BinaryRow::new(0)` is an in-memory stub with no backing bytes. Its `to_serialized_bytes()` method consequently emitted only the 4-byte arity, although Paimon's schemaless BinaryRow format also requires the 8-byte fixed part. Native DataSplit serialization therefore produced bytes that its strict decoder correctly rejected for non-partitioned tables.

This is consistent with the existing fixes in #349 and #364: the empty-row writer must emit the Java-compatible 12-byte value, while the decoder must continue rejecting truncated bodies. The missing case became visible after #565 added native DataSplit deserialization.

### Brief change log

- Normalize `arity == 0 && data.is_empty()` in `BinaryRow::to_serialized_bytes()` to the existing `EMPTY_SERIALIZED_ROW`.
- Add a BinaryRow regression test for the exact 12-byte Java wire representation.
- Add an empty-partition DataSplit native round-trip test using bytes generated by Apache Paimon Java 1.4.2.

### Tests

- Independent Java verification with Maven Central `org.apache.paimon:paimon-bundle:1.4.2`:
  - empty BinaryRow serde is byte-stable and 12 bytes;
  - empty-partition DataSplit serde round-trips;
  - generated DataSplit bytes are 77 bytes.
- `cargo +1.92 test -p paimon --lib` — 1899 passed, 0 failed, 1 ignored.
- `cargo +1.92 clippy -p paimon --all-targets -- -D warnings` — passed.
- `cargo +1.92 fmt --all -- --check` — passed.

### API and Format

No API or wire-version change. This fixes the arity-zero serialization path to conform to the existing Java BinaryRow format. Populated rows are unchanged, and the strict decoder remains unchanged.

### Documentation

No documentation update is required. The serializer comment records why the canonical empty row needs normalization.


JingsongLi (approved):
+1

PR #605, opened 2026-07-23
ci: install cargo-about CLI on Intel macOS

XiaoHongbo-Hope:
## What

- keep the prebuilt cargo-about install path on supported runners
- install cargo-about 0.9.1 with the required cli feature on macos-15-intel
- apply the fix to both Python and Go release workflows

## Why

The v0.3.0-rc1 Python release failed because taiki-e/install-action fell back to a source install on x86_64 macOS without enabling cargo-about's cli feature. The step reported success but did not install the cargo-about binary. The Go release workflow uses the same path.

Failed run: https://github.com/apache/paimon-rust/actions/runs/30012257692

## Verification

- parsed both modified workflows as YAML
- git diff --check


QuakeWang (approved):
+1

PR #606, opened 2026-07-23
fix(table): short-circuit empty global index ranges

QuakeWang:
### Purpose

Manifest row-range pruning fails open for legacy files without row-id metadata. An empty global index result can therefore still reach manifest and split planning, causing a planning error instead of returning an empty plan.

### Brief change log

Short-circuit full and manifest-list scans when global index evaluation returns an empty range set. Add coverage for regular, traced, and delta scans in fast and full modes.

### Tests


XiaoHongbo-Hope:
+1

JingsongLi (approved):

PR #608, opened 2026-07-24
feat(c): export vector-search projection over the C FFI

JunRuiLee:
### Purpose

The C FFI vector-search builder (`bindings/c/src/vector_search.rs`) exposed
`with_vector_column` / `query_vector` / `limit` / `options` / `filter` /
`execute_read`, but not projection. C callers therefore could not restrict the
materialized columns and always read every user column (plus the search score),
even though the Rust `VectorSearchBuilder` already supports `with_projection`.

### Changes

- Add `paimon_vector_search_builder_with_projection(builder, columns)`, mirroring
  `paimon_read_builder_with_projection`: `columns` is a null-terminated array of
  null-terminated C strings (pass `null` to clear). Names are stored on the
  builder state and applied in `execute_read` via `with_projection`.
- Column-name validation is deferred to `execute_read` (the core
  `VectorSearchBuilder::with_projection` is infallible and resolves the projection
  when the search runs), so an unknown projected column surfaces as an error from
  `execute_read` rather than from the setter.
- Add a compile-time C ABI signature guard for the new symbol.

Only `execute_read` is affected; the search-only paths are unchanged.

### Tests

- `vector_search_pk_projection_restricts_columns`: projecting `["id"]` through
  `execute_read` materializes only `id` + `__paimon_search_score` (the vector
  column is excluded), matching a Rust reference.
- `vector_search_projection_unknown_column_errors_at_execute_read`: an unknown
  projected column fails loud at `execute_read`.

`cargo test -p paimon-c`, `cargo fmt --all --check`, and
`cargo clippy -p paimon-c --all-targets -- -D warnings` all pass.

### DIFF
--- a/bindings/c/src/tests.rs
+++ b/bindings/c/src/tests.rs
@@ -2474,3 +2474,146 @@ fn plan_from_split_bytes_rejects_garbage() {
     assert!(!r.error.is_null());
     unsafe { paimon_error_free(r.error) };
 }
+
+// --- Vector-search projection through execute_read ------------------------
+
+/// Core Rust reference: sorted output column names materialized by `execute_read`
+/// under an optional projection.
+fn rust_execute_read_column_names(
+    table: &Table,
+    column: &str,
+    query: Vec<f32>,
+    limit: usize,
+    projection: Option<&[&str]>,
+) -> Vec<String> {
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #610, opened 2026-07-24
perf(table): search global index fields concurrently

QuakeWang:
### Purpose

The global index scanner only parallelized shard reads within a single field. For multi-field `AND` predicates, indexed fields were still evaluated sequentially, leaving the configured I/O concurrency underutilized.

### Brief change log

  - Evaluate independent indexed fields concurrently.
  - Share one scan-scoped semaphore across all shard reads so total I/O concurrency stays within `global-index.thread-num`.
  - Add regression coverage for result correctness and the shared concurrency limit.

### Tests


JingsongLi (approved):
+1

PR #613, opened 2026-07-26
fix(datafusion): return vector_search rows in best-first relevance order

XiaoHongbo-Hope:
Makes the DataFusion `vector_search` table function return rows in the index's
best-first relevance order (previously it re-read rows in physical file order, so the
first row was often not the most similar).

**What it does**
- Materializes the scored row-ids (best-first) from `execute_scored`, reads the matching
  rows (projection pushdown + internal `_ROW_ID`), then gathers them back into relevance
  rank and drops `_ROW_ID`. Output schema is unchanged.
- Casts columns to the provider schema (e.g. `Utf8` → DataFusion `Utf8View`) so string
  columns round-trip.
- Runs the search + read + gather in a custom execution-time `VectorSearchExec` (not at
  planning time), so `EXPLAIN` stays cheap and the work runs under the `TaskContext`.
- Keeps the full top-k for the ANN search and applies the outer `LIMIT` by truncating the
  already-ranked result before reading, so recall is unchanged and a large top-k with a
  small outer `LIMIT` only reads what it returns.

**Dependency / known limitation — needs #614**
This PR makes the scan honor the order of `SearchResult.row_ids`, but that order is only
fully correct once #614 is merged. On current `main`, `SearchResult::top_k` does **not**
sort when `candidate_count <= k` (i.e. `limit >= number of matched rows`), so for that
case the index returns rows in scored-map order and this scan faithfully reproduces it.
With #614, the ranking is correct for all limits.

Because of that, the string-column test asserts its hit set order-independently (the cast
is what it verifies); deterministic relevance ordering is covered by
`test_vector_search_orders_by_relevance`, which uses a top-k smaller than the candidate
count (the sorted/truncation path). Merge order: **#614 first, then rebase this PR** for
the full-ordering guarantee across all limits.

### DIFF
--- a/crates/integrations/datafusion/src/vector_search.rs
+++ b/crates/integrations/datafusion/src/vector_search.rs
@@ -24,6 +24,7 @@ use async_trait::async_trait;
 use datafusion::arrow::array::{
     Array, ArrayRef, Int64Array, RecordBatch, RecordBatchOptions, UInt32Array,
 };
+use datafusion::arrow::compute::cast;
 use datafusion::arrow::datatypes::SchemaRef as ArrowSchemaRef;
 use datafusion::catalog::Session;
 use datafusion::catalog::TableFunctionImpl;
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #614, opened 2026-07-26
fix(index): sort search top-k results by relevance when not truncating

XiaoHongbo-Hope:
> **Companion PR:** #613

`SearchResult::top_k` only sorted by score on the truncation path (candidates > k).
When `k >= candidate count`, it returned rows in the unordered scored-map / insertion
order instead of by relevance.

This went unnoticed because the row order was previously discarded downstream (row-range
scans re-read in file order); but consumers that honor the row order — e.g. the
DataFusion `vector_search` scan (#613) — then get **unranked** results whenever
`limit >= number of matched rows`. The same bug existed in full_text's separate
`SearchResult::top_k`. (`hybrid_search` re-ranks its results itself, so it is not
affected.)

**Fix**: sort the no-truncation path by relevance rank in both the vector and full_text
`SearchResult::top_k`, with a consistent NaN-safe comparator and smaller-row-id
tie-break; update the unit tests that asserted the old insertion-order behavior.

### DIFF
--- a/crates/paimon/src/full_text.rs
+++ b/crates/paimon/src/full_text.rs
@@ -126,9 +126,8 @@ impl SearchResult {
 
     /// Return top-k results by score (descending).
     pub fn top_k(&self, k: usize) -> Self {
-        if self.row_ids.len() <= k {
-            return self.clone();
-        }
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #615, opened 2026-07-26
fix(spec): validate sequence.field against table schema

jackylee-ch:
`sequence.field` was only partly validated: a missing column, a repeated field,
`merge-engine=first-row`, a cross-partition update table, or an aggregate function
on the sequence field were all accepted and persisted, and the write path then fell
back to the auto-increment sequence — silently ignoring the user's ordering intent.

**Fix**: validate at create and alter time, mirroring all five checks in Java
`SchemaValidation#validateSequenceField`. The aggregation check lived inside
`AggregationConfig`, reachable only on aggregation/partial-update engines, so it is
hoisted into a free function keyed on the option map alone.

### DIFF
--- a/crates/paimon/src/spec/aggregation.rs
+++ b/crates/paimon/src/spec/aggregation.rs
@@ -185,9 +185,6 @@ impl<'a> AggregationConfig<'a> {
         fields: &[DataField],
         primary_keys: &[String],
     ) -> crate::Result<()> {
-        // Same source as the read path: `sequence.field` parsed by CoreOptions.
-        let core_options = CoreOptions::new(self.options);
-        let sequence_fields = core_options.sequence_fields();
         for (key, value) in self.options {
             let Some((col, kind)) = parse_field_scoped_option_key(key) else {
                 continue;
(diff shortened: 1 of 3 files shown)


JingsongLi:
Claims to be aligned with Java’s `validateSequenceField`, but omits validation for cross-partition updates. `sequence.field` can still be configured even when not all partition keys are included in the primary key, whereas Java explicitly rejects this combination. Both the `create` and `alter` paths should include partition keys and primary keys in the validation.

JingsongLi:
Please resolve conflicts.

jackylee-ch:
@JingsongLi conflict fixed, PTAL

JingsongLi:
- `Schema::validate_sequence_field` omits the `options.fieldAggFunc(field) == null` validation found in Java’s `SchemaValidation.validateSequenceField`.
- Currently, Rust incorrectly accepts:
  `sequence.field=ts`
  `fields.ts.aggregate-function=sum`
- I added a temporary regression test; it was expected to return an error from `build()`, but it actually returned `Ok(Schema)`. This issue has been consistently reproducible.
- This is not only incompatible with Java; during a partial update, the sequence field may be skipped by the aggregation configuration logic, potentially leading to a situation where “the configuration is written to the metadata but is silently ignored in practice.”
- I recommend that `validate_sequence_field` directly reject all `fields.<sequence-field>.aggregate-function` entries, and that create/alter tests be added.

jackylee-ch:
Confirmed — all five Java checks are covered now. The aggregation one was in #638; I've folded it into this PR so it's reviewable in one place, and closed #638.

JingsongLi (approved):
+1

jackylee-ch:
@JingsongLi Merged latest main — #660 hoisted the validators into `validate_final_schema`, so this PR's delta was re-applied there. No logic or test changes from the approved commit; CI is green. PTAL.

PR #618, opened 2026-07-26
feat(io): let FileIO wrap a caller-provided filesystem operator

### LINKED ISSUE #616: FileIO cannot use a caller-provided opendal operator
### Motivation

`Storage::build` resolves the backing opendal operator through a closed match on URI schemes, so an embedding engine cannot supply its own operator — for example one wrapped with a custom service, a metrics/caching layer, or a preconfigured object-store client.

Java Paimon treats configurable file IO as a first-class concern: `FileIO` is an interface, `FileIOLoader` is a ServiceLoader SPI matched by scheme, and `CatalogContext` accepts `preferIO`/`fallbackIO` loaders programmatically. paimon-rust has no equivalent injection point.

### Proposal

The minimal hook: `FileIOBuilder::with_operator(op)` accepting a prebuilt `opendal::Operator` that takes precedence over scheme resolution. A scheme-keyed loader registry mirroring Java's `FileIOLoader` could follow as a larger change, but the builder hook alone unblocks embedders.

### Use case

StreamFusion (github.com/datafusion-contrib/StreamFusion) stores Flink operator state in local Paimon PK tables; its write path needs a thin custom fs service (the stock service's per-write `create_dir_all` dominated a checkpoint-path CPU profile). It currently carries this hook as a fork patch.

jordepic:
### Purpose

Linked issue: close #616

`Storage::build` resolves the backing opendal operator through a closed scheme match, so an embedding engine cannot supply its own operator (custom service, metrics/caching layer, preconfigured client). Java Paimon exposes this axis through the `FileIO` interface, the `FileIOLoader` SPI, and `CatalogContext`'s `preferIO`/`fallbackIO`; this is the minimal Rust analog.

### Brief change log

- `FileIOBuilder::with_operator(op)` stores a prebuilt `opendal::Operator`; when set, it takes precedence over scheme resolution in `build`.
- `Storage::Custom { op }` variant carries it through; `create()` reuses the fs-relative path handling.

### Tests

- `cargo test -p paimon --lib io::` (85 tests) passes; the builder path is exercised by the existing FileIO suites plus downstream use in StreamFusion, which drives all its state-table IO through an operator injected here.

### API and Format

Additive API on `FileIOBuilder`; no storage-format change.

### Documentation

Doc comments on the new builder method.

### DIFF
--- a/crates/paimon/src/io/file_io.rs
+++ b/crates/paimon/src/io/file_io.rs
@@ -372,6 +372,7 @@ pub struct FileIOBuilder {
     scheme_str: Option<String>,
     props: HashMap<String, String>,
     cache: Option<Arc<LocalCache>>,
+    operator: Option<Operator>,
 }
 
 impl FileIOBuilder {
@@ -380,11 +381,22 @@ impl FileIOBuilder {
             scheme_str: Some(scheme_str.to_string()),
             props: HashMap::default(),
             cache: None,
+            operator: None,
         }
     }
 
-    pub(crate) fn into_parts(self) -> (String, HashMap<String, String>) {
-        (self.scheme_str.unwrap_or_default(), self.props)
+    pub(crate) fn into_parts(self) -> (String, HashMap<String, String>, Option<Operator>) {
+        (self.scheme_str.unwrap_or_default(), self.props, self.operator)
+    }
+
+    /// Uses a caller-provided opendal operator instead of building one from the scheme:
+    /// embedders bring their own storage backend (for example a customized local-filesystem
+    /// service) without registering a scheme. Paths are handed to the operator in
+    /// scheme-relative form, exactly as the built-in local-filesystem backend receives them,
+    /// so the operator's root decides what they resolve against.
+    pub fn with_operator(mut self, operator: Operator) -> Self {
+        self.operator = Some(operator);
+        self
     }
 
     pub fn with_prop(mut self, key: impl ToString, value: impl ToString) -> Self {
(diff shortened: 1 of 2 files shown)


jordepic:
cc @JingsongLi — this is the minimal Rust analog of Java Paimon's configurable FileIO (`FileIOLoader`/`preferIO`); would appreciate your review. #619 is a companion from the same downstream use.

JingsongLi:
`Storage::Custom` always uses `fs_relative_path` without exception; for example, `s3://bucket/a` is converted to `3://bucket/a`. As a result, preconfigured operators such as S3 or OSS are effectively unusable, and directory listings will return incorrect paths. It is recommended to include a scheme/path resolver or to explicitly restrict the API and name it the “filesystem operator.”

jordepic:
Thanks @JingsongLi — you're right, the hook resolved every path with the filesystem rules, so a scheme'd path like `s3://bucket/a` was silently mangled to `3://bucket/a`. Took your second suggestion and scoped the API explicitly: renamed to `with_fs_operator`, documented the filesystem contract (absolute paths, `file:` URLs, Windows drive paths, resolved against the operator's root), and `Storage::CustomFs` now rejects scheme'd paths with a clear error instead of misresolving them — object-store operators need the bucket/scheme resolution the built-in backends perform, which this hook deliberately doesn't reimplement. Added resolution and rejection tests, and fixed the rustfmt CI failure.

jordepic:
Sorry for the ping here @JingsongLi !  Would you mind taking a look at this one? Thank you!

Also, I don't yet have an apache email but I'd love to join the Paimon slack! I'm doing a lot of work here for 
https://github.com/datafusion-contrib/StreamFusion/tree/main

JingsongLi (approved):
+1

PR #619, opened 2026-07-26
perf(arrow): evaluate In/NotIn with a hash set for byte, string, and integer columns

### LINKED ISSUE #617: In/NotIn predicates evaluate one comparison kernel per literal — O(rows × literals)
### Problem

`evaluate_set_membership_predicate` in `arrow/residual.rs` (shared by the parquet row filter and the exact residual backstop) evaluates an `In`/`NotIn` literal set by running one whole-column comparison kernel per literal and OR-combining the masks. Cost is O(rows × literals).

This is fine for hand-written filters with a few literals, but an engine that pushes a batch of point-lookup keys down as an `In` predicate (hundreds of literals against a table read) makes every probe quadratic — e.g. 500 keys × 500K rows = 250M comparisons per pushed batch.

### Proposal

Build a hash set of the literal values once and answer membership with a single pass over the column, for the common column shapes (Binary/LargeBinary/BinaryView, Utf8/LargeUtf8/Utf8View, Int8–Int64). Everything else keeps the per-literal loop, preserving its exact semantics including its error behavior for unconvertible literals.

Measured in StreamFusion (github.com/datafusion-contrib/StreamFusion), which pushes each streaming batch's state keys down as `In`: without the fix, per-batch probes were quadratic; with it, the probe is one hash-set pass at parquet decode.

jordepic:
### Purpose

Linked issue: close #617

The shared residual evaluator (used by both the parquet row filter and the exact batch backstop) evaluates `In`/`NotIn` as one comparison kernel per literal, OR-combining the masks — O(rows × literals). An engine pushing a batch of point-lookup keys as an `In` predicate makes every table probe quadratic.

### Brief change log

- `set_membership_hash_mask`: build a hash set of the literal values once, answer membership in a single pass over the column, for Binary/LargeBinary/BinaryView, Utf8/LargeUtf8/Utf8View, and Int8–Int64 columns.
- Any other column/literal shape — and any literal that would not convert for the column — returns `None` and falls back to the existing per-literal loop, preserving its exact semantics including its error behavior for unconvertible literals.
- Null rows produce `false` for both `In` and `NotIn`, matching the loop's `sanitize_filter_mask` convention.

### Tests

- New: `test_in_hash_path_filters_exactly_with_nulls`, `test_not_in_hash_path_excludes_nulls`, `test_in_hash_path_on_binary_column` (In and NotIn masks incl. nulls), `test_in_unconvertible_literal_still_errors` (error behavior preserved through the fallback).
- `cargo test -p paimon --lib arrow::residual` (26), `read_builder` (33), and `arrow::format::parquet` (41) all pass.

### API and Format

None — internal evaluation only; results are unchanged.

### Documentation

Doc comment on the fast path explains the shape and the fallback contract.

### DIFF
--- a/crates/paimon/src/arrow/residual.rs
+++ b/crates/paimon/src/arrow/residual.rs
@@ -540,6 +540,10 @@ fn evaluate_set_membership_predicate(
         });
     }
 
+    if let Some(mask) = set_membership_hash_mask(array, op, literals) {
+        return Ok(mask);
+    }
+
     let mut combined = match op {
         PredicateOperator::In => BooleanArray::from(vec![false; array.len()]),
         PredicateOperator::NotIn => {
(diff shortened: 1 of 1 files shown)


jordepic:
cc @JingsongLi — companion to #618; would appreciate your review.

JingsongLi (approved):
+1

PR #620, opened 2026-07-27
fix(spec): validate bucket-key against table schema

jackylee-ch:
`bucket-key` is never validated against the schema. A typo'd or stale key resolves
to no field index in `TableWrite` (a lenient `filter_map` lookup), so
`bucket_key_indices` ends up empty and the assigner falls through to
`ConstantBucketAssigner(0)`: every row silently lands in bucket 0 instead of being
hashed. A blank value was worse — `bucket_key()` returned `Some(vec![""])`, which
also suppressed the primary-key fallback.

**Fix**: validate `bucket-key` at create and alter time, mirroring Java
`TableSchema#originalBucketKeys` — each key must exist, must not repeat, must not
be a partition field, and on a primary-key table must be part of the primary key.
Blank entries are dropped and an all-blank value resolves to `None`, so the
primary-key fallback applies as in Java.

This rejects table definitions that were previously accepted, but any such table
was already writing everything to bucket 0.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -783,10 +783,20 @@ impl<'a> CoreOptions<'a> {
     }
 
     /// Explicit bucket key columns. If not set, defaults to primary keys for PK tables.
+    ///
+    /// Blank entries are dropped and an all-blank option resolves to `None`,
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #621, opened 2026-07-27
fix(spec): accept Java's long memory-size units and reject overflow

jackylee-ch:
`parse_memory_size` only recognized short unit spellings, so a valid Java option
value such as `target-file-size = '256 mebibytes'` parsed as `None` and the caller
silently fell back to its default. It also multiplied unchecked:
`'9007199254740993 kb'` wraps to a negative size in a release build and panics in
a debug build.

**Fix**: accept every alias in Java `MemorySize.MemoryUnit` (`bytes`, `kibibytes`,
`mebibytes`, `gibibytes`, `tebibytes` alongside the short forms), and use
`checked_mul` so an overflowing value returns `None` — the callers then fall back
to their default, which is what Java's thrown "numeric overflow" leads to as well.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -1218,14 +1218,15 @@ impl<'a> CoreOptions<'a> {
     }
 }
 
-/// Parse a memory size string to bytes using binary (1024-based) semantics.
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #623, opened 2026-07-27
fix(datafusion): apply dynamic options to vector search

### LINKED ISSUE #622: [Bug] vector_search ignores session dynamic options
### Search before asking

- [x] I searched the existing issues and did not find a duplicate.

### Problem

`SQLContext` stores options set through statements such as:

```sql
SET 'paimon.blob-as-descriptor' = 'true';
```

Normal Paimon table scans merge these session-scoped dynamic options into the loaded `Table`. The automatically registered `vector_search` table function instead captures the raw `Catalog`, loads the table directly, and materializes matched rows from that unmodified table.

As a result, `vector_search` ignores session dynamic options. In particular, selecting a BLOB column from `vector_search` resolves and reads the actual BLOB payload even when `paimon.blob-as-descriptor` is enabled, rather than returning the serialized descriptor.

### Expected behavior

Tables loaded by the automatically registered `vector_search` function should receive the same session dynamic options as ordinary table scans. The standalone public registration API should remain backward compatible.

### Proposed fix

Pass the shared dynamic-options map when `SQLContext` registers `vector_search`, and merge a snapshot of those options into the table before creating the vector-search provider. Add a regression test covering `blob-as-descriptor`.

shyjsarah:
### Purpose

Linked issue: close #622

`SQLContext` applies session-scoped Paimon dynamic options to ordinary table scans, but the automatically registered `vector_search` function loaded its table directly from the raw catalog. This caused supported read options such as `blob-as-descriptor` to be ignored when vector-search results were materialized.

### Brief change log

- pass the shared `DynamicOptions` instance when `SQLContext` registers `vector_search`
- apply the current supported dynamic options to the loaded vector-search table
- exclude time-travel selectors because vector search currently resolves ANN candidates from the latest snapshot; forwarding them could search one snapshot and materialize another
- preserve the existing standalone `register_vector_search` and `VectorSearchFunction::new` APIs by using empty options there
- add a regression test covering both `blob-as-descriptor` propagation and time-travel-option exclusion

### Tests

- `cargo check -p paimon-datafusion --all-features`
- `cargo clippy -p paimon-datafusion --all-features --lib --tests --no-deps -- -D warnings`
- `cargo test -p paimon-datafusion test_vector_search_applies_supported_session_dynamic_options --lib`
- `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_java_vindex_table`

The full `paimon-datafusion --lib` run completed 311 tests successfully; 7 existing warehouse-fixture tests failed because the local `PAIMON_TEST_WAREHOUSE` did not contain `partitioned_log_table` / `multi_partitioned_log_table`.

### API and Format

### Documentation

No documentation changes are required; this makes `vector_search` honor supported session read options without implying time-travel support.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -198,7 +198,12 @@ impl SQLContext {
                 Some(session_state),
             )),
         );
-        register_table_functions(&self.ctx, &catalog, default_db.unwrap_or("default"));
+        register_table_functions(
+            &self.ctx,
+            &catalog,
+            default_db.unwrap_or("default"),
+            self.dynamic_options.clone(),
+        );
         self.catalogs.insert(catalog_name.clone(), catalog);
         if is_first {
             self.set_current_catalog(catalog_name).await?;
@@ -3257,9 +3262,15 @@ fn register_table_functions(
     ctx: &SessionContext,
     catalog: &Arc<dyn Catalog>,
     default_database: &str,
+    dynamic_options: DynamicOptions,
 ) {
     crate::blob_view::register_blob_view(ctx, Arc::clone(catalog), default_database);
-    crate::vector_search::register_vector_search(ctx, Arc::clone(catalog), default_database);
+    crate::vector_search::register_vector_search_with_dynamic_options(
+        ctx,
+        Arc::clone(catalog),
+        default_database,
+        dynamic_options,
+    );
     #[cfg(feature = "fulltext")]
     crate::full_text_search::register_full_text_search(ctx, Arc::clone(catalog), default_database);
     crate::hybrid_search::register_hybrid_search(ctx, Arc::clone(catalog), default_database);
--- a/crates/integrations/datafusion/src/vector_search.rs
+++ b/crates/integrations/datafusion/src/vector_search.rs
@@ -45,8 +45,6 @@ use futures::{stream, TryStreamExt};
 use paimon::catalog::Catalog;
 use paimon::spec::{
     BigIntType, CoreOptions, DataField, DataType, ROW_ID_FIELD_ID, ROW_ID_FIELD_NAME,
-    SCAN_SNAPSHOT_ID_OPTION, SCAN_TAG_NAME_OPTION, SCAN_TIMESTAMP_MILLIS_OPTION,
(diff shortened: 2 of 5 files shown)


crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -198,7 +198,12 @@ impl SQLContext {
                 Some(session_state),
             )),
         );
-        register_table_functions(&self.ctx, &catalog, default_db.unwrap_or("default"));
+        register_table_functions(
+            &self.ctx,
+            &catalog,
+            default_db.unwrap_or("default"),
+            self.dynamic_options.clone(),
```
XiaoHongbo-Hope, line 207: This also forwards time-travel options such as scan.version, but vector search still uses the latest snapshot.

shyjsarah: Good catch. The DE vector-search path reads the latest index snapshot, so forwarding a time-travel selector could indeed search the latest snapshot and materialize rows from an older one. I updated the patch to filter `scan.version`, `scan.timestamp-millis`, `scan.snapshot-id`, and `scan.tag-name` while still forwarding supported options such as `blob-as-descriptor`. The regression test now covers both behaviors.

crates/integrations/datafusion/src/vector_search.rs
```diff
@@ -426,6 +461,21 @@ impl ExecutionPlan for VectorSearchExec {
     }
 }
 
+/// Vector search currently resolves candidates from the latest snapshot, so forwarding a
+/// time-travel selector would search one snapshot and materialize rows from another.
+fn vector_search_dynamic_options(dynamic_options: &DynamicOptions) -> HashMap<String, String> {
+    let mut options = dynamic_options.read().unwrap().clone();
+    for key in [
+        SCAN_VERSION_OPTION,
```
JingsongLi: Why need to remote these options? Can we just keep it as is.

shyjsarah: Good point. We can keep these options as-is. I updated vector search to honor the resolved time-travel snapshot instead of always using the latest snapshot, and added regression tests covering `scan.version` together with `blob-as-descriptor`.

JingsongLi (approved):
+1

PR #625, opened 2026-07-28
feat(datafusion): support alter column

QuakeWang:
### Purpose

DataFusion SQL supports adding, dropping, and renaming columns, but rejects `ALTER COLUMN` even though the core schema layer already supports type and nullability changes.

This change exposes the existing schema evolution capabilities through SQL while preserving schema and primary-key index consistency.

### Brief change log

- Support `ALTER COLUMN ... TYPE` and `SET DATA TYPE`.
- Support `SET NOT NULL` and `DROP NOT NULL`.
- Preserve existing nullability for type-only changes.
- Reject unsupported conversion expressions and non-atomic mixed ALTER operations.
- Prevent destructive changes to primary-key vector and full-text index columns.

### Tests

- Added SQL mapping and validation unit tests.
- Added a filesystem catalog integration test.
- Added primary-key index schema guard tests.
- Passed formatting, Clippy, and affected test suites.

### API and Format

Adds DataFusion SQL DDL support. No storage format changes.

### Documentation

Updated the supported SQL statement list.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -28,6 +28,8 @@
 //! - `ALTER TABLE db.t ADD COLUMN col TYPE`
 //! - `ALTER TABLE db.t DROP COLUMN col`
 //! - `ALTER TABLE db.t RENAME COLUMN old TO new`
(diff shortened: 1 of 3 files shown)


JingsongLi (approved):
+1

PR #629, opened 2026-07-29
fix(table): skip unrelated BLOB and vector files in row-range reads

XiaoHongbo-Hope:
### Purpose

Row-range pruning for data-evolution tables currently retains an entire connected file component. A wide anchor file can therefore keep many disjoint rolled BLOB or vector sidecars, causing oversized splits and unnecessary planning and read work.

### Changes

- Keep dedicated files that intersect the selected row ranges.
- Preserve one provider witness only when no selected file can validate that provider.
- Align vector provider identity with reader bunching.
- Net ADD/DELETE entries before pruning in full and delta scans.
- Read only the selected rolled BLOB/vector sidecars.

### Tests

- `cargo test -p paimon table::table_scan::tests`
- `cargo test -p paimon table::data_evolution_reader::tests`
- `cargo clippy --locked -p paimon --lib --tests --no-deps -- -D warnings`

### DIFF
--- a/crates/paimon/src/table/data_evolution_reader.rs
+++ b/crates/paimon/src/table/data_evolution_reader.rs
@@ -5592,7 +5592,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_scan_and_read_retains_complete_rolled_dedicated_group() {
+    async fn test_scan_and_read_prunes_unselected_rolled_dedicated_sources() {
         let tempdir = tempdir().unwrap();
         let table_path = local_file_path(tempdir.path());
         let bucket_dir = tempdir.path().join("bucket-0");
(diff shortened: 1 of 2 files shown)


JingsongLi:
- `table_scan.rs:270` first performs the manifest ADD/DELETE merge, followed by row-range pruning at `table_scan.rs:271`, but `table_scan.rs:286` is where `manifest_entries_after_merge` is recorded. Therefore, this field actually records the count “after merge and row-range pruning.”
- Additionally, `table_scan.rs:275` only overwrites `manifest_entries_after_manifest_filters` with the count after merge and pruning when a row range exists; the semantics of this field are inconsistent depending on whether a row range is present or not.
- For example, if 2 live entries remain after a merge and 1 is subsequently removed by a row range, the current `after_merge` value is incorrectly recorded as 1. It is recommended to save the count immediately after the merge, have the number of row ranges expressed separately by `pruned_by_row_ranges`, and add a test that distinguishes between the two phases.

XiaoHongbo-Hope:
> * `table_scan.rs:270` first performs the manifest ADD/DELETE merge, followed by row-range pruning at `table_scan.rs:271`, but `table_scan.rs:286` is where `manifest_entries_after_merge` is recorded. Therefore, this field actually records the count “after merge and row-range pruning.”
> * Additionally, `table_scan.rs:275` only overwrites `manifest_entries_after_manifest_filters` with the count after merge and pruning when a row range exists; the semantics of this field are inconsistent depending on whether a row range is present or not.
> * For example, if 2 live entries remain after a merge and 1 is subsequently removed by a row range, the current `after_merge` value is incorrectly recorded as 1. It is recommended to save the count immediately after the merge, have the number of row ranges expressed separately by `pruned_by_row_ranges`, and add a test that distinguishes between the two phases.

Updated

JingsongLi (approved):
+1

PR #633, opened 2026-07-30
perf(parquet): read row groups concurrently

XiaoHongbo-Hope:
## What

- Read independent Parquet row groups concurrently for predicate-free scans.
- Preserve row-group and batch output order.
- Stream one batch per row group with backpressure instead of materializing a full row group.
- Share concurrency and projected-byte budgets across all readers and DataFusion scan partitions.
- Keep predicate and row-selection reads on the existing path.

## Resource controls

- `read.parquet.row-group.parallelism` (default: `8`; set to `1` to disable)
- `read.parquet.row-group.max-inflight-bytes` (default: `256 mb`)

The byte budget uses projected uncompressed Parquet bytes and is shared by the whole scan.

## Why

A compacted Parquet file is one scan partition, so remote row-group reads were serialized even though the ranges are independent. Per-file concurrency alone could multiply memory and object-store requests when DataFusion runs several scan partitions.

## Benchmark

- Direct remote read (28 row groups, three-column projection): `4.27s -> 1.07s`
- End-to-end aggregate query, four alternating runs: median `5.156s -> 3.245s` (`1.59x`)
- Disabling row-group concurrency returned the query to about `6s`.

Row count, batch count, total count, and result hashes matched.

## Tests

- Ordering and overlapping-read regression test
- Multi-reader shared-budget test
- Slow-consumer batch-backpressure test
- Permit release and option-validation tests
- Paimon lib: `2054 passed, 1 ignored`
- DataFusion scan tests: `29 passed`
- Clippy, formatting, and diff checks passed

### DIFF
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -35,8 +35,8 @@ use bytes::Bytes;
 use futures::future::BoxFuture;
 use futures::{FutureExt, StreamExt, TryFutureExt, TryStreamExt};
 use parquet::arrow::arrow_reader::{
-    ArrowPredicate, ArrowPredicateFn, ArrowReaderOptions, RowFilter as ParquetRowFilter,
-    RowSelection, RowSelector,
+    ArrowPredicate, ArrowPredicateFn, ArrowReaderMetadata, ArrowReaderOptions,
+    RowFilter as ParquetRowFilter, RowSelection, RowSelector,
 };
 use parquet::arrow::async_reader::{AsyncFileReader, MetadataFetch};
 use parquet::arrow::{AsyncArrowWriter, ParquetRecordBatchStreamBuilder, ProjectionMask};
@@ -53,6 +53,12 @@ use std::ops::Range;
 use std::sync::Arc;
 
 pub(crate) struct ParquetFormatReader;
+
+/// Maximum number of Parquet row groups decoded concurrently for an
+/// unfiltered full-file read.
+const PARQUET_ROW_GROUP_READ_CONCURRENCY: usize = 8;
+/// Bound projected uncompressed data buffered by concurrent row-group reads.
+const PARQUET_ROW_GROUP_IN_FLIGHT_BYTES: u64 = 256 * 1024 * 1024;
 
 struct ParquetRowFilterPredicate {
     inner: Box<dyn RowFilter>,
@@ -301,7 +307,8 @@ impl FormatFileReader for ParquetFormatReader {
         batch_size: Option<usize>,
         row_selection: Option<Vec<RowRange>>,
     ) -> crate::Result<ArrowRecordBatchStream> {
-        let arrow_file_reader = ArrowFileReader::new(file_size, reader);
+        let shared_reader: Arc<dyn FileRead> = reader.into();
+        let arrow_file_reader = ArrowFileReader::new(file_size, Arc::clone(&shared_reader));
 
         let empty_predicates = Vec::new();
         let (preds, file_fields): (&[Predicate], &[DataField]) = match predicates {
@@ -361,7 +368,7 @@ impl FormatFileReader for ParquetFormatReader {
             .collect();
 
         let mask = ProjectionMask::roots(&parquet_schema, root_indices);
-        batch_stream_builder = batch_stream_builder.with_projection(mask);
+        batch_stream_builder = batch_stream_builder.with_projection(mask.clone());
 
         let mut decoder_predicates = build_parquet_row_filter(&parquet_schema, preds, file_fields)?
             .map(ParquetRowFilter::into_predicates)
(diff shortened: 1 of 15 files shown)


crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -460,6 +467,61 @@ impl FormatFileReader for ParquetFormatReader {
...
+                            ArrowFileReader::new(file_size, shared_reader),
+                            reader_metadata,
+                        )
+                        .with_projection(mask)
+                        .with_row_groups(vec![row_group_index]);
+                        if let Some(size) = batch_size {
+                            builder = builder.with_batch_size(size);
+                        }
+                        let stream = builder.build().map_err(Error::from)?;
+                        stream.try_collect::<Vec<_>>().await.map_err(Error::from)
```
QuakeWang: `try_collect::<Vec<_>>()` materializes all batches in each row group before yielding any output. With multiple row groups running concurrently, decoded Arrow batches are retained in addition to the Parquet bytes used to calculate`row_group_parallelism`, so actual in-flight memory can substantially exceed the intended budget and batch-level backpressure is lost.

XiaoHongbo-Hope: Fixed in 4553e82. Each row-group producer now has a one-batch channel and reserves capacity before polling the next batch, restoring batch-level backpressure. The concurrency and projected-byte semaphores are shared by the entire scan, including concurrent DataFusion partitions. Both limits are configurable, and the new tests cover slow consumers, multiple readers sharing one budget, ordering, and permit release.

QuakeWang (approved):
+1

JingsongLi:
Row-group producers fail to stop in a timely manner after downstream cancellation
Location: parquet.rs:517, parquet.rs:667
Issue: The producer first calls `reserve()` to reserve a channel, then waits for `stream.next()`; if the downstream drops the receiver due to `LIMIT`, an error, or query cancellation, the detached task waiting for I/O will not receive the cancellation notification and will continue to hold the `ParquetReadPermit`.
Trigger Conditions: Start `forward_row_group_batches` using a pending stream, then drop the receiver; if the task has not exited within 100 milliseconds, the test consistently fails.
Impact: Unnecessary object storage requests and decoding continue to run; more importantly, they continue to consume scan-shared row-group/byte permits, which may slow down DataFusion partitions that are still active.
The fix is very simple:
```
let next = tokio::select! {
    _ = sender.closed() => return,
    next = stream.next() => next,
};
```
It is recommended that the coordinator listen for `row_group_tx.closed()` while waiting for `read_budget.acquire(...)`; after canceling the `acquire` future, the acquired semaphore permits will be automatically released. After applying the above fix, both my cancellation reproduction test and the original backpressure test passed.

JingsongLi:
The budget passing process can be simplified as follows:
- Add an optional `with_parquet_read_budget` parameter to `PaimonReadBuilder`, which DataFusion injects before `new_read()`.
- Avoid first creating and then discarding a default budget for each partition, only to override it later via the exposed `TableRead::with_parquet_read_budget` method.
- If shared resources such as caches and memory reservations are added later, they can be centralized into a single `ReadContext` to reduce the current field passing across a dozen or so readers.

XiaoHongbo-Hope:
> The budget passing process can be simplified as follows:
> 
> * Add an optional `with_parquet_read_budget` parameter to `PaimonReadBuilder`, which DataFusion injects before `new_read()`.
> * Avoid first creating and then discarding a default budget for each partition, only to override it later via the exposed `TableRead::with_parquet_read_budget` method.
> * If shared resources such as caches and memory reservations are added later, they can be centralized into a single `ReadContext` to reduce the current field passing across a dozen or so readers.

Thanks, updated

XiaoHongbo-Hope:
> Row-group producers fail to stop in a timely manner after downstream cancellation Location: parquet.rs:517, parquet.rs:667 Issue: The producer first calls `reserve()` to reserve a channel, then waits for `stream.next()`; if the downstream drops the receiver due to `LIMIT`, an error, or query cancellation, the detached task waiting for I/O will not receive the cancellation notification and will continue to hold the `ParquetReadPermit`. Trigger Conditions: Start `forward_row_group_batches` using a pending stream, then drop the receiver; if the task has not exited within 100 milliseconds, the test consistently fails. Impact: Unnecessary object storage requests and decoding continue to run; more importantly, they continue to consume scan-shared row-group/byte permits, which may slow down DataFusion partitions that are still active. The fix is very simple:
> 
> ```
> let next = tokio::select! {
>     _ = sender.closed() => return,
>     next = stream.next() => next,
> };
> ```
> 
> It is recommended that the coordinator listen for `row_group_tx.closed()` while waiting for `read_budget.acquire(...)`; after canceling the `acquire` future, the acquired semaphore permits will be automatically released. After applying the above fix, both my cancellation reproduction test and the original backpressure test passed.

Thanks, updated

JingsongLi (approved):
+1

PR #634, opened 2026-07-30
fix(datafusion): normalize Paimon column identifiers consistently

QuakeWang:
### Purpose

Paimon-specific SQL handlers bypass DataFusion's logical planner and previously used raw parser identifiers. This caused DDL and DML column handling to behave inconsistently with DataFusion's identifier-normalization setting.

### Brief change log

  - Apply consistent identifier normalization to Paimon DDL, partition operations, UPDATE, MERGE, and INSERT OVERWRITE.
  - Preserve quoted identifiers and reject unknown or duplicate normalized target columns.
  - Add regression tests for normalization-enabled and disabled behavior.

### Tests

  - DataFusion identifier-normalization integration tests

### Documentation

Updated `docs/src/sql.md`.


JingsongLi (approved):
+1

PR #638, opened 2026-07-31
fix(spec): reject aggregation on a sequence field for every merge engine

jackylee-ch:
Java `SchemaValidation#validateSequenceField` checks `fieldAggFunc(field) == null`
for every field in `sequence.field`, and that validator runs for every merge
engine. The Rust equivalent lived inside
`AggregationConfig::validate_field_scoped_options`, which is only reached when
`merge-engine=aggregation` — so on the default deduplicate engine a schema
combining `sequence.field = 'ts'` with `fields.ts.aggregate-function = 'sum'` was
silently accepted and persisted.

**Fix**: hoist the check into a free `validate_no_aggregation_on_sequence_field`
keyed only on the option map, and call it from `Schema::new` and
`TableSchema::apply_changes` alongside the other sequence-field validations, so it
no longer depends on the merge engine. The existing aggregation-engine test is
widened to cover all five engines, plus create/alter tests on the default engine.

### DIFF
--- a/crates/paimon/src/spec/aggregation.rs
+++ b/crates/paimon/src/spec/aggregation.rs
@@ -181,9 +181,6 @@ impl<'a> AggregationConfig<'a> {
         fields: &[DataField],
         primary_keys: &[String],
     ) -> crate::Result<()> {
-        // Same source as the read path: `sequence.field` parsed by CoreOptions.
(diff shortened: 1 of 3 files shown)


jackylee-ch:
Folded into #615 so the full `validateSequenceField` parity is reviewable in one place.

PR #641, opened 2026-07-31
fix(spec): reject reserved system field names at create and alter time

jackylee-ch:
`TableSchema::validate_no_reserved_fields` already rejects the five Java
`SpecialFields.SYSTEM_FIELD_NAMES` (`_SEQUENCE_NUMBER`, `_VALUE_KIND`, `_LEVEL`,
`rowkind`, `_ROW_ID`) and the `_KEY_` prefix — but it is only reachable from
`validate_resolved_structure`, which runs when loading an externally-supplied
schema. Neither `Schema::new` nor `TableSchema::apply_changes` checked names, so
`CREATE TABLE` with a column named `_ROW_ID` was accepted and persisted through
every catalog path (including DataFusion DDL), and that column was then silently
shadowed by the system value on read.

**Fix**: split the name half into a free
`validate_no_reserved_field_names(&[DataField])` and call it from both schema
paths plus the existing resolved-structure check. The field-id half is unchanged.

`pk_hybrid_table_with_reserved_column` deliberately builds a schema carrying
`_ROW_ID` to exercise the read-side guard, so it now deserializes that schema from
JSON instead of going through `Schema::builder` — which is precisely the case the
read guard exists for: metadata written by another engine.

### DIFF
--- a/crates/paimon/src/spec/schema.rs
+++ b/crates/paimon/src/spec/schema.rs
@@ -203,31 +203,17 @@ impl TableSchema {
     /// colliding with a system field (e.g. `_ROW_ID`) is otherwise excluded
     /// from the physical read and silently filled with the system value.
     fn validate_no_reserved_fields(&self) -> crate::Result<()> {
-        // Java SpecialFields.SYSTEM_FIELD_NAMES.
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #642, opened 2026-07-31
fix(predicate): resolve _ROW_ID by name, not by its placeholder leaf index

plusplusjiajia:
### Purpose

 _ROW_ID is not a table column, so PredicateBuilder cannot resolve it and callers hand-build the leaf with an index of their own — 0, as the existing test_with_filter_extracts_row_id_ranges does. That index is a placeholder, but Rust binds predicate leaves by index at construction and never rebinds by name, so every
  index-based resolution binds the predicate to whatever field sits at that index. On a table whose first field is a partition key, WHERE _ROW_ID >= 10 pruned partitions by comparing them against row ids; on a primary-key table it was pushed down as a key conjunct and dropped key versions before the merge; a residual
  _ROW_ID != 5 filtered the first user column.

  Java rebinds by name (PredicateRemapper, and rowIdSafeResidualFilter for row ids); pypaimon's _change_index looks the field up by name. Rust was the only implementation with no name-level step.

### Brief change log

  Resolve by name. spec::is_row_id_column is the single definition, and every positional resolution consults it first: the three Predicate index-mapping primitives (which is what keeps a row id off partition keys, bucket keys and PK pushdown), residual mask evaluation, per-file remapping, scan widening, the projected
  field-id set, Parquet/ORC decoder pushdown, stats pruning, and the data-evolution BLOB dependency check. DataEvolutionReader attaches _ROW_ID when a residual needs it but the caller did not project it.

  Row ranges. extract_row_id_ranges returns Some(vec![]) for an empty union rather than None — "matches no rows" and "nothing extracted" are different, and reporting the latter let an enclosing AND skip the subtree and overstate its ranges. Ranges are only a superset, so ranges_represent_conjunct lets a conjunct leave the
  residual only when the ranges stand in for it exactly, and row_ranges_from_filter clears a derived range when its filter is replaced.

  Fail closed. Only data-evolution reads can synthesize _ROW_ID; elsewhere the predicate is unenforceable and every alternative is a silent wrong answer, so reject_row_id_filter rejects it on the read path (not the builder — TableRead is public): primary-key merges, format tables, Vortex's scanless fast path, and the Mosaic
  and Blob readers. This is only for _ROW_ID — it is the only reserved name a predicate references, and the only one with no schema position at all.

### DIFF
--- a/crates/paimon/src/arrow/filtering.rs
+++ b/crates/paimon/src/arrow/filtering.rs
@@ -17,7 +17,7 @@
 
 use crate::arrow::schema_evolution::create_index_mapping;
 pub(crate) use crate::predicate_stats::{predicates_may_match_with_schema, StatsAccessor};
-use crate::spec::{DataField, Predicate, PredicateOperator};
+use crate::spec::{is_row_id_column, DataField, Predicate, PredicateOperator};
 
 /// Remap predicates from table-level indices to file-level indices.
 /// Predicates referencing fields not present in the file are resolved based on
@@ -44,6 +44,12 @@ fn remap_predicate(predicate: &Predicate, mapping: &[Option<usize>]) -> Predicat
             op,
             literals,
         } => {
+            // `_ROW_ID` is not a file column and has no per-file position, so
+            // mapping its placeholder index would collapse the leaf to a
+            // constant. Keep it; the residual resolves it by name.
+            if is_row_id_column(column) {
+                return predicate.clone();
+            }
             match mapping.get(*index).copied().flatten() {
                 Some(file_index) => Predicate::Leaf {
                     column: column.clone(),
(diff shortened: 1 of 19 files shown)


JingsongLi:
`_ROW_ID = ...` predicates targeting non-Data Evolution tables are extracted and removed from the residual; they subsequently become silently ineffective, resulting in the return of all data. This behavior has been reproduced in local tests. A check for Data Evolution should be performed prior to extraction, or the predicates should be retained in the residual and rejected across the board.

plusplusjiajia:
> `_ROW_ID = ...` predicates targeting non-Data Evolution tables are extracted and removed from the residual; they subsequently become silently ineffective, resulting in the return of all data. This behavior has been reproduced in local tests. A check for Data Evolution should be performed prior to extraction, or the predicates should be retained in the residual and rejected across the board.

@JingsongLi  Thanks for catching this — confirmed and fixed in the second commit. _ROW_ID >= 102 returned all four rows of a file at first_row_id = 100. Took your second option. One thing worth your check: I gated on row-tracking.enabled || data-evolution.enabled rather than data evolution alone, since what matters is whether data files record a first_row_id, which row tracking assigns.

JingsongLi (approved):
+1

PR #644, opened 2026-07-31
fix(table): keep every split of a format table under a positive LIMIT

### LINKED ISSUE #643: Format table LIMIT can return fewer rows than the table has
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/paimon-rust/issues) and found nothing similar.

### Paimon Rust version

main (34449a7)

### Compute Engine

DataFusion

### Minimal reproduce step

A format table with two parquet files, the first one empty:

```rust
write_parquet(&table_dir.join("part-0.parquet"), &[]);        // 0 rows
write_parquet(&table_dir.join("part-1.parquet"), &[1, 2, 3]); // 3 rows
```

```sql
SELECT id FROM paimon.test_db.events LIMIT 1;
```

### What doesn't meet your expectations?

The query returns 0 rows and reports success. The table has 3 rows, so `LIMIT 1` must return 1.

`FormatTableScan::apply_limit_pushdown` keeps only the first `limit` splits:

https://github.com/apache/paimon-rust/blob/34449a7/crates/paimon/src/table/format_table_scan.rs#L275-L284

A format table has no manifest, so scan planning does not know how many rows a data file holds. Keeping `limit` files reads a number of files as if it were a number of rows. An empty data file is enough to break it, and an empty data file is ordinary output: an engine that writes one file per task leaves one behind whenever a task matched nothing.

Java does not do this. `FormatTableScan.FormatTableScanPlan.splits` drops splits only for a non-positive limit, and says why:

```java
// Keep all splits for a positive limit because FormatDataSplit has no row count.
if (limit != null && limit <= 0) {
    return new ArrayList<>();
}
```

paimon-rust's own Paimon-table path is also stricter than this: `PaimonTableScan::apply_limit_pushdown` skips a split whose merged row count is unknown, instead of counting it as if it held enough rows.

### Anything else?

The same shape can under-read without an empty file whenever a filter is pushed down exactly, because the first `limit` files may hold no matching row.

- [X] I'm willing to submit a PR!

sundapeng:
### Purpose

Linked issue: close #643

A format table has no manifest, so scan planning does not know how many rows a data file holds. `FormatTableScan::apply_limit_pushdown` kept only the first `limit` splits anyway, which reads the number of files as if it were a number of rows.

An empty data file is enough to break it, and an empty data file is ordinary output: an engine that writes one file per task leaves one behind whenever a task matched nothing. A table of three rows whose first file is empty answers `SELECT id FROM t LIMIT 1` with zero rows, and reports success.

### Brief change log

- Keep every split for a positive limit and let execution stop once it has enough rows. A limit of zero still needs no split.

This is what Java does. `FormatTableScan.FormatTableScanPlan.splits` returns an empty list only for a non-positive limit, with the same reason in a comment. paimon-rust's own Paimon-table path is already stricter than the format path was: `PaimonTableScan::apply_limit_pushdown` skips a split whose merged row count is unknown rather than counting it as if it held enough rows.

### Tests

`crates/integrations/datafusion/tests/format_table_statistics.rs`: `test_format_table_limit_is_not_answered_from_a_file_count` writes an empty file followed by a three-row file and asserts `LIMIT 1` returns a row. It fails on main with `left: 0, right: 1`.

```
cargo fmt --all -- --check
cargo clippy --locked -p paimon -p paimon-datafusion --all-targets -- -D warnings
cargo test --locked -p paimon --lib
cargo test --locked -p paimon-datafusion --lib --test format_table_statistics --test count_pushdown --test scan_pruning_trace --test read_tables
```

### API and Format

No API or storage format change.

### Documentation

None needed.

### DIFF
--- a/crates/integrations/datafusion/tests/format_table_statistics.rs
+++ b/crates/integrations/datafusion/tests/format_table_statistics.rs
@@ -197,3 +197,17 @@ async fn test_format_table_with_empty_file_counts_zero() {
         0
     );
 }
+
+/// A positive `LIMIT` must not be answered from a file count. A format table has no
+/// row counts, so keeping only the first `limit` files is a guess, and an empty data
+/// file — ordinary output from an engine that writes one file per task — makes the
+/// guess wrong.
+#[tokio::test]
+async fn test_format_table_limit_is_not_answered_from_a_file_count() {
+    let (_tmp, context, table_dir) = setup_format_table().await;
+    write_parquet(&table_dir.join("part-0.parquet"), &[]);
+    write_parquet(&table_dir.join("part-1.parquet"), &[1, 2, 3]);
+
+    let scanned = scanned_rows(&context, "SELECT id FROM paimon.test_db.events LIMIT 1").await;
+    assert_eq!(scanned, 1, "LIMIT 1 over 3 rows must return a row");
+}
--- a/crates/paimon/src/table/format_table_scan.rs
+++ b/crates/paimon/src/table/format_table_scan.rs
@@ -272,13 +272,17 @@ impl<'a> FormatTableScan<'a> {
         }
     }
 
+    /// A limit of zero needs no split at all. A positive one keeps every split: a format
+    /// table has no row counts, so the number of files says nothing about the number of
+    /// rows, and dropping files would answer the query from a guess.
+    ///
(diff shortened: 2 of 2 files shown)


QuakeWang (approved):
+1

PR #645, opened 2026-08-01
perf(vindex): read search indexes through positional ranges

JunRuiLee:
### Purpose

This is PR 1/3 of a stacked vector-search performance series.

The regular/global-index vindex path currently buffers the complete index file
before search. For large indexes, a single query usually needs only metadata and
a subset of posting lists, so the full read adds unnecessary I/O and peak memory.

This change adds a safe positional-read foundation and switches the **single-query
regular/global-index vindex path** to it.

### Scope boundary

- Multi-query global-index searches intentionally remain on the buffered path in
  this PR; #646 replaces that fallback with native batch range reads.
- PK-vector ANN segment loading also remains buffered in this PR and is handled by
  #647.
- Lumina keeps its existing buffered read path.

### Brief change log

- Add a `SeekRead` adapter backed by Paimon's asynchronous `FileRead`.
- Poll asynchronous storage reads on the caller's Tokio runtime instead of a
  second runtime.
- Return `UnexpectedError` instead of panicking when range-backed vindex search is
  driven outside a Tokio runtime.
- Run synchronous vindex and Lumina search work on a dedicated executor, avoiding
  `block_in_place` and host blocking-pool deadlocks.
- Honor `global-index.thread-num` as the per-query file fan-out while separating
  that logical limit from the process executor's physical OS-thread cap.
- Bound physical workers to `max(32, 4 * available_parallelism)` and reclaim
  growth workers after one idle minute, while retaining the CPU-sized baseline.
- Preserve strict sequential execution for a query when
  `global-index.thread-num = 1`.
- Validate requested ranges and report short reads explicitly.
- Coalesce nearby reads with bounded read amplification and share the range-I/O
  semaphore across cloned readers.
- Forward `try_clone_reader` and read capabilities through the type-erased vindex
  input adapter; the range reader deliberately reports storage-agnostic defaults
  because it accepts arbitrary range counts and the optimal window is backend-specific.
- Keep queued-task cancellation cheap and avoid blocking a current-thread Tokio
  runtime when an already-started search is cancelled.

### Tests

- `cargo test -p paimon vindex --lib` — 118 passed
- `cargo test -p paimon test_global_index_thread_num --lib` — 2 passed
- `cargo test -p paimon --test pk_vector_baseline_test` — 8 passed
- `cargo test -p paimon --test pk_vector_batch_test` — 10 passed
- `cargo test -p paimon --test pk_vector_java_fixture_test` — 1 passed
- `python3 scripts/dependencies.py verify`
- `python3 scripts/release_licenses.py --check` — 21 generated legal files validated
- `git diff --check`

Coverage includes caller-runtime polling, a non-Tokio error path, current-thread runtimes, clone support,
short reads, bounded coalescing, physical worker clamping, idle worker reclamation,
and local filesystem reads when the host blocking pool has only one thread.

### API and Format

### Documentation

The existing `global-index.thread-num` API documentation now includes regular
vector global-index search. No user-facing storage or query API changes are required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4535,6 +4535,7 @@ dependencies = [
  "bytes",
  "chrono",
  "crc32fast",
+ "crossbeam-channel",
  "futures",
  "hex",
  "hmac 0.12.1",
--- a/benchmarks/tpcds/DEPENDENCIES.rust.tsv
+++ b/benchmarks/tpcds/DEPENDENCIES.rust.tsv
@@ -81,6 +81,7 @@ core-foundation-sys@0.8.7		X									X
 cpufeatures@0.2.17		X									X					
 cpufeatures@0.3.0		X									X					
 crc32fast@1.5.0		X									X					
+crossbeam-channel@0.5.16		X									X					
 crossbeam-deque@0.8.7		X									X					
 crossbeam-epoch@0.9.20		X									X					
 crossbeam-utils@0.8.22		X									X					
--- a/bindings/c/DEPENDENCIES.rust.tsv
+++ b/bindings/c/DEPENDENCIES.rust.tsv
@@ -60,6 +60,7 @@ core-foundation-sys@0.8.7		X									X
 cpufeatures@0.2.17		X									X				
 cpufeatures@0.3.0		X									X				
 crc32fast@1.5.0		X									X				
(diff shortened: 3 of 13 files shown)


JingsongLi (approved):
+1

PR #646, opened 2026-08-01
perf(vindex): execute compatible queries through native batch search

JunRuiLee:
### Purpose

This is PR 2/3 of a stacked vector-search performance series and builds on #645.

PR #645 introduced positional range reads while deliberately keeping multi-query
global-index vindex searches on the whole-file buffered path. This PR replaces
that fallback with the native batch-search APIs from `paimon-vindex-core`, while
retaining positional range reads.

### Brief change log

- Prepare and validate each query before backend execution.
- Group queries with compatible `top_k`, `nprobe`, and row-id filters.
- Execute each compatible group in a single native vindex batch-search call.
- Rely on vindex-core's bounded posting-list read batches and oversized-list
  streaming instead of imposing a Rust-side query-count cap.
- Retain scalar execution for groups containing a single query.
- Restore results to their original query order.
- Validate backend result dimensions before slicing per-query results.
- Reuse probed posting lists across compatible queries instead of repeating
  scalar range reads.

PK-vector ANN segments remain on their existing buffered path until #647. The
reader validation and optimization hooks introduced here intentionally prepare
the lifecycle consumed by #647.

### Tests

- `cargo test -p paimon vindex --lib` — 123 passed
- `cargo test -p paimon vector_search_builder --lib` — 77 passed
- `cargo test -p paimon --lib` — 2096 passed, 1 ignored
- `cargo fmt --check`
- `git diff --check`

Coverage verifies mixed-group ordering, scalar equivalence, filtered native
batch execution, range reuse, avoidance of full-file reads, reuse across the
former 16-query boundary, and graceful failure outside a Tokio runtime.

### API and Format

No public API or storage-format changes. Existing batch result ordering is
preserved.

### Documentation

No user-facing documentation changes are required.

### DIFF
--- a/crates/paimon/src/table/vector_search_builder.rs
+++ b/crates/paimon/src/table/vector_search_builder.rs
@@ -65,6 +65,7 @@ use arrow_select::interleave::interleave_record_batch;
 use futures::{stream, TryStreamExt};
 use paimon_vindex_core::distance::MetricType;
 use paimon_vindex_core::index::VectorIndexReader as VIndexReader;
+use paimon_vindex_core::io::SeekRead;
 use roaring::RoaringTreemap;
 use std::cmp::Ordering;
 use std::collections::{BinaryHeap, HashMap, HashSet};
@@ -98,11 +99,30 @@ impl VectorIndexBackend {
     }
 }
 
-fn current_tokio_runtime_handle() -> crate::Result<tokio::runtime::Handle> {
-    tokio::runtime::Handle::try_current().map_err(|error| crate::Error::UnexpectedError {
-        message: "Vector index range reader requires a Tokio runtime".to_string(),
-        source: Some(Box::new(error)),
+async fn execute_vindex_searches<S: SeekRead + 'static>(
+    io_meta: GlobalIndexIOMeta,
+    options: HashMap<String, String>,
+    vector_searches: Vec<VectorSearch>,
+    source: S,
+    file_name: String,
+    shard_concurrency: usize,
+) -> crate::Result<Vec<Option<HashMap<u64, f32>>>> {
+    let panic_context = if vector_searches.len() > 1 {
+        "vindex global-index batch search task failed"
+    } else {
+        "vindex global-index search task failed"
+    };
+    execute_global_index(panic_context, move || {
+        let mut reader = VindexVectorGlobalIndexReader::new(io_meta, options)
+            .with_batch_shard_concurrency(shard_concurrency);
+        reader
+            .visit_batch_vector_search(&vector_searches, |_| Ok(source))
+            .map_err(|e| crate::Error::DataInvalid {
+                message: format!("Failed to read vindex index file '{}': {}", file_name, e),
+                source: Some(Box::new(e)),
+            })
     })
+    .await
 }
 
 pub struct VectorSearchBuilder<'a> {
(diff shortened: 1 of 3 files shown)


crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1482,65 +1482,40 @@ async fn evaluate_batch_vector_search(
...
+                                    message: format!(
+                                        "Failed to open vindex file '{}' for range reads: {}",
+                                        file_name, e
+                                    ),
+                                    source: None,
+                                }
+                            })?;
+                            let source = VindexFileReader::new(
+                                Arc::new(file_reader),
+                                current_tokio_runtime_handle()?,
```
JingsongLi: [P2] Preserve the non-Tokio batch fallback

This now unconditionally requires a Tokio handle, but before this change the `vector_searches.len() > 1` branch read the file into a `Cursor` and therefore worked when polled by async-std, smol, or a plain `futures` executor. A public batch with two or more queries will now fail before searching with `Vector index range reader requires a Tokio runtime`.

Please retain the whole-file `Cursor` fallback when `Handle::try_current()` fails, or make the range bridge executor-agnostic, and add a non-Tokio batch regression test. The existing test only covers the pre-existing single-query limitation.

crates/paimon/src/vindex/reader.rs
```diff
@@ -165,61 +222,175 @@ fn search_vindex(
...
+        let params = VectorSearchParams::new(prepared.top_k, prepared.nprobe);
+        let (labels, distances) = match &prepared.filter_bytes {
+            Some(filter) => reader
+                .search_batch_with_roaring_filter(&queries, indices.len(), params, filter)
+                .map_err(|e| crate::Error::DataInvalid {
+                    message: format!("paimon-vindex-core filtered batch search failed: {}", e),
+                    source: Some(Box::new(e)),
+                })?,
+            None => reader
+                .search_batch(&queries, indices.len(), params)
```
JingsongLi: [P2] Bound the native batch by its full working set

Every compatible group is flattened and passed to one native batch call. In `paimon-vindex-core` 0.3.0, this allocates a `query_count * nlist` `f32` centroid-product matrix, plus query copies, per-query heaps, and `query_count * top_k` result arrays; L2 IVF-PQ also retains per-query `m * ksub` tables. Posting-list reads are byte-bounded, but this query-side scratch is not.

For example, 10,000 queries with `nlist=4096` consume about 164 MB per active index-file job for the centroid matrix alone. With the default 32 concurrent shard jobs, those matrices alone can exceed 5 GB. The previous scalar loop bounded this scratch to one query at a time. Please chunk compatible groups using a working-set budget derived from at least `nlist`, dimension, `top_k`, and shard concurrency, then restore results to their original positions.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1482,65 +1482,40 @@ async fn evaluate_batch_vector_search(
...
+                            let file_reader = input.reader().await.map_err(|e| {
+                                crate::Error::DataInvalid {
+                                    message: format!(
+                                        "Failed to open vindex file '{}' for range reads: {}",
+                                        file_name, e
+                                    ),
+                                    source: None,
+                                }
+                            })?;
+                            let source = VindexFileReader::new(
```
JingsongLi: [P2] Apply `global-index.thread-num` to aggregate range I/O

Each admitted shard now creates an independent `VindexFileReader`, and each reader owns a private 32-permit range-read semaphore. The outer scheduler already admits up to `global-index.thread-num` shard jobs (default 32), so native multi-range reads can drive up to `32 * 32 = 1024` concurrent `FileRead::read` calls. This defeats the option's documented per-operation global-index I/O fan-out limit; in particular, setting it to `1` no longer reproduces strictly sequential I/O. The removed batch path issued one whole-file read per admitted shard.

Please share one operation-wide semaphore across all readers, or otherwise divide the range-read budget across admitted shard jobs.

JingsongLi (commented):
I found three issues in the new native-batch and range-read path.

JingsongLi (approved):
+1

PR #647, opened 2026-08-01
perf(vector-search): execute PK-vector leaves concurrently

JunRuiLee:
### Purpose

This is PR 3/3 of the stacked vector-search performance series. The preceding
PRs #645 and #646 have been merged into `main`, and this branch has been rebased
onto the current `main` so the PR diff is isolated to concurrent PK-vector leaf
execution.

PK-vector bucket search previously did not execute all independent search work
concurrently. ANN segments and exact-fallback files were scheduled in separate
phases, and ANN CPU work could run inline or serially despite a configured
`global-index.thread-num` greater than one.

This change represents ANN segments and uncovered exact files as independent
leaves, executes them through one bounded concurrency window, and moves blocking
index searches to the shared process-level executor.

### Brief change log

- Combine ANN-segment and exact-file work into one concurrent bucket leaf plan.
- Load each ANN segment lazily inside its leaf instead of preloading every index.
- Keep vindex segments range-backed on the caller runtime and Lumina segments
  buffered.
- Share the configured range-read budget across lazy vindex segment readers.
- Run blocking ANN and regular global vector-index searches on the dedicated
  process-level executor.
- Apply one per-query concurrency budget across buckets, ANN segments, and exact
  fallback files.
- Apply a shared process-level budget across concurrent queries while keeping
  its logical capacity separate from the executor's physical OS-thread cap.
- Drain submitted leaves and report errors deterministically by submission order.
- Preserve deterministic Top-K merging regardless of leaf completion order.
- Validate `global-index.thread-num` against Tokio's semaphore limit.
- Keep permits owned by started ANN and regular global vector-index work after
  caller cancellation without blocking Tokio runtime threads.

### Tests

- `cargo check -p paimon`
- `cargo test -p paimon --lib` — 2142 passed, 1 ignored
- `cargo test -p paimon --test pk_vector_baseline_test` — 8 passed
- `cargo test -p paimon --test pk_vector_batch_test` — 10 passed
- `cargo test -p paimon --test pk_vector_java_fixture_test` — 1 passed
- `git diff --check`

### API and Format

No public API or storage-format changes. The existing
`global-index.thread-num` option and default value are retained.

### Documentation

No user-facing documentation changes are required. The concurrency and execution
changes are internal to vector-index reads.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -123,6 +123,15 @@ const DYNAMIC_BUCKET_TARGET_ROW_NUM_OPTION: &str = "dynamic-bucket.target-row-nu
 const DEFAULT_DYNAMIC_BUCKET_TARGET_ROW_NUM: i64 = 200_000;
 const DEFAULT_GLOBAL_INDEX_ROW_COUNT_PER_SHARD: i64 = 100_000;
 const DEFAULT_GLOBAL_INDEX_THREAD_NUM: i64 = 32;
+const MAX_GLOBAL_INDEX_THREAD_NUM: i64 = {
+    let tokio_max = (usize::MAX >> 3) as u64;
+    let i32_max = i32::MAX as u64;
+    if tokio_max < i32_max {
+        tokio_max as i64
+    } else {
+        i32_max as i64
+    }
+};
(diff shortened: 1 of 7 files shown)


JingsongLi (approved):
+1

PR #648, opened 2026-08-01
feat: rust examples for creating and querying Paimon tables

ganesh-skumar:
### Purpose

Add Rust examples demonstrating how to create a Paimon table with sample data and query it using DataFusion.

### Brief change log

- Add `create_table.rs` example to create a Paimon table and insert sample data.
- Add `datafusion_query.rs` example demonstrating the DataFusion DataFrame API.
- Add `paimon-datafusion` as a workspace dependency for the examples.


ganesh-skumar:
Hi, While checking the CI failure, the rust examples require adding `paimon-datafusion` and `datafusion` dependencies  to paimon crate because they're located under crates/paimon/examples. As a result, the generated DEPENDENCIES.rust.tsv files also need to be updated.

I was wondering whether it would be more appropriate to place these examples in a separate example crate to avoid introducing these additional dependencies into the core paimon crate.

Or is updating the generated DEPENDENCIES.rust.tsv files the expected approach in this case?

Thanks!

crates/paimon/examples/create_table.rs
```diff
@@ -0,0 +1,104 @@
...
+// set the catalog path and run example using:
+// cargo run --package paimon --example create_table
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    // Open local catalog
+    let catalog = create_catelog().await?;
+
+    // Create new database
+    catalog
+        .create_database("my_db", false, HashMap::new())
```
ArnavBalyan: Could we make this setup example safely rerunnable? With ignore_if_exists = false, a second invocation would fail. Perhaps the example can drop/recreate its sample objects everytime the example is run for a clean start.

Cargo.toml
```diff
@@ -41,6 +49,7 @@ arrow-string = "58.0"
 datafusion = "54.0.0"
 datafusion-ffi = "54.0.0"
 paimon = { version = "0.4.0", path = "crates/paimon" }
+paimon-datafusion = { path = "crates/integrations/datafusion" }
```
ArnavBalyan: Would it be better to place datafusion_query.rs under the paimon-datafusion crate?

crates/paimon/examples/create_table.rs
```diff
@@ -0,0 +1,104 @@
...
+    let msg = writer.prepare_commit().await?;
+
+    txn.commit(msg).await?;
+
+    Ok(())
+}
+
+pub async fn create_catelog() -> Result<Arc<dyn Catalog>, Box<dyn Error>> {
+    let mut options = Options::new();
+    options.set(CatalogOptions::WAREHOUSE, "/path-to/testdata");
```
ArnavBalyan: Could the path be provided as a cli argument, sharable by both examples. Since the user's would need to edit this code to wire up their real warehouse file otherwise.

crates/paimon/examples/create_table.rs
ganesh-skumar: got it, updated to ignore db if exists and creates fresh table.

Cargo.toml
ganesh-skumar: good idea, created integrations/datafusion/examples/datafusion_query.rs

crates/paimon/examples/create_table.rs
ganesh-skumar: updated to pass path via cli

crates/paimon/examples/create_table.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Check if table exists in catalog
+    let table_exists = match catalog.get_table(&identifier).await {
+        Ok(_) => true,
+        Err(paimon::Error::TableNotExist { .. }) => false,
+        Err(error) => return Err(error.into()),
+    };
+
+    // if exists, drop and create fresh table and data
+    if table_exists {
+        catalog.drop_table(&identifier, false).await?;
```
leaves12138, line 81: Could we avoid dropping an existing table by default? The warehouse is supplied as an arbitrary CLI path, so a user who points this example at an existing warehouse containing `my_db.users` would permanently delete that table and all of its data. Please fail when the table already exists, require an explicit destructive flag such as `--overwrite`, or use a clearly isolated example namespace/warehouse.

crates/paimon/examples/create_table.rs
```diff
@@ -0,0 +1,122 @@
...
+    writer.write_arrow_batch(&batch).await?;
+
+    let msg = writer.prepare_commit().await?;
+
+    txn.commit(msg).await?;
+
+    Ok(())
+}
+
+pub async fn create_catelog(warehouse: String) -> Result<Arc<dyn Catalog>, Box<dyn Error>> {
```
leaves12138: Nit: please rename `create_catelog` to `create_catalog`. The same typo also appears in `datafusion_query.rs`.

leaves12138 (commented):
I ran both examples end to end, including rerunning the setup and querying the resulting table. The functional flow works, but there is one destructive default behavior that should be addressed.

crates/paimon/examples/create_table.rs
ganesh-skumar: Fixed

crates/paimon/examples/create_table.rs
ganesh-skumar: Thanks, added table recreation based on flag.

crates/paimon/examples/create_table.rs
```diff
@@ -0,0 +1,132 @@
...
+    let mut args = std::env::args().skip(1);
+
+    let warehouse = args.next().ok_or_else(|| {
+        std::io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            "usage: cargo run --package paimon --example create_table -- <warehouse-path> ",
+        )
+    })?;
+
+    let overwrite = args.any(|arg| arg == "--overwrite");
```
QuakeWang, line 44: Could we make `--overwrite` order-independent, or document the required order? The first argument is always treated as the warehouse, so `--overwrite <warehouse-path>` fails; only `<warehouse-path> --overwrite` works. At minimum, please update the usage string to `<warehouse-path> [--overwrite]`.

ganesh-skumar: Thanks! Updated the usage doc with the correct argument order. I also tried making it order-independent, but it made the code a bit messy for a beginner-friendly create_table example.

QuakeWang (approved):
+1

ArnavBalyan (approved):
Thanks for the review @QuakeWang and thanks for the PR @ganesh-skumar!

PR #658, opened 2026-08-03
feat(go): add table write bindings

XiaoHongbo-Hope:
### Purpose

Expose standard table writes to Go.

### Changes

- Add write, prepare, merge, commit, retry, overwrite, truncate, and abort APIs.
- Carry append/overwrite mode from the builder into the committer.
- Preserve Arrow caller ownership by copying input into C-owned memory.
- Add Go integration tests and usage documentation.

### Scope

Commit messages can be merged only within one Go process. Cross-process
serialization is out of scope. With multiple primary-key fixed-bucket
writers, callers must assign each `(partition, bucket)` to one writer.

This PR adds Go APIs and does not change the storage format.


JingsongLi (approved):
+1

PR #660, opened 2026-08-03
fix(spec): validate final schema consistently

QuakeWang:
### Purpose

Schema creation and schema evolution used separate validation paths. As a result, alter operations could produce final schemas that bypassed deletion-vector and primary-key index constraints.

### Brief change log

- Centralize final-schema validation for create and alter paths.
- Validate deletion-vector, PK vector, and PK full-text index configurations.
- Reject malformed index column lists and unsafe BLOB or bucket-key column changes.
- Preserve defensive runtime coverage for legacy or externally written schemas.

### Tests

### API and Format

### Documentation

Not required.


JingsongLi (approved):
+1

PR #661, opened 2026-08-03
docs: align the global index reference with the supported types and options

jackylee-ch:
`drop_global_index` documented only BTree and bitmap, but the drop path accepts
every type the create procedures build — `SUPPORTED_GLOBAL_INDEX_TYPES_FOR_DROP` is
`"btree, bitmap, lumina, lumina-vector-ann, ivf-flat, ivf-pq"`, enforced through
`normalize_global_index_type_for_drop`, and
`test_drop_global_index_accepts_lumina_type` / `..._vindex_type` already cover the
wider set.

The Global Index Options table also omitted two options the code reads:
`global-index.thread-num` (default 32, used by the concurrent vector-search path)
and `global-index.column-update-action` (default `THROW_ERROR`, used when a commit
updates an indexed column).

### DIFF
--- a/docs/src/sql.md
+++ b/docs/src/sql.md
@@ -1049,7 +1049,9 @@ CALL sys.drop_global_index(
 );
 ```
 
-BTree and bitmap global indexes can be dropped through this procedure currently.
+`index_type` accepts every type the create procedures build: `btree`, `bitmap`,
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #662, opened 2026-08-03
fix(spec): say precision, not length, in two type-parse errors

jackylee-ch:
`LocalZonedTimestampType::from_str` and `TimeType::from_str` both parse a
`precision_str` into a `precision` field, but report `"Invalid <T> length. Unable to
parse length as a u32."` The sibling `TimestampType::from_str` already has the right
wording, so this copies it. `BINARY`, `CHAR`, `VARBINARY` and `VARCHAR` keep
"length" — those really are lengths.

Also a README typo (`bug reportor` -> `bug report`), `rust` -> `Rust` in the tagline
to match `docs/mkdocs.yml`, and a trailing space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 [Latest Version]: https://img.shields.io/crates/v/paimon.svg
 [crates.io]: https://crates.io/crates/paimon
 
-The rust implementation of Apache Paimon. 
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #663, opened 2026-08-03
fix(spec): normalize file-format options to lowercase

jackylee-ch:
Java routes every file-format option through `CoreOptions.normalizeFileFormat`,
which lowercases the value. Rust returned it verbatim, and `DataFileWriter`
interpolates it straight into the file name — so `'file.format' = 'PARQUET'` wrote
`data-<uuid>-0.PARQUET` where Java writes `.parquet`. Reads still worked (the format
dispatch lowercases the path), so the effect was inconsistent file naming rather
than a hard failure.

**Fix**: trim and lowercase in `file_format`, `changelog_file_format` and
`vector_file_format`, and treat a blank value as unset so it falls back to the
default instead of yielding an empty extension. Two consumers that re-did
`.trim().to_ascii_lowercase()` themselves now rely on the accessor.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -974,11 +974,15 @@ impl<'a> CoreOptions<'a> {
     ///
     /// Java leaves this unset by default. When present, vector columns are
     /// written to files named `*.vector.<format>`.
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #664, opened 2026-08-03
docs: refresh the getting-started dependency and feature notes

jackylee-ch:
Three entries in `docs/src/getting-started.md` predate the 0.3.0 release:

- the dependency snippets pin `paimon = "0.2.0"`, while crates.io publishes 0.3.0
  and `docs/src/sql.md` already says 0.3.0;
- the default-feature sentence lists only `storage-fs` and `storage-memory`, but
  `crates/paimon/Cargo.toml` has
  `default = ["storage-memory", "storage-fs", "storage-oss"]`;
- `storage-cos`, `storage-azdls`, `storage-obs` and `storage-gcs` are footnoted as
  unreleased with a git-dependency workaround, but they shipped in v0.3.0
  (`git tag --contains` on the commit that added them returns `v0.3.0`), so the
  footnote and its snippet are dropped.

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -25,15 +25,15 @@ Add `paimon` to your `Cargo.toml`:
 
 ```toml
 [dependencies]
-paimon = "0.2.0"
+paimon = "0.3.0"
 tokio = { version = "1", features = ["full"] }
 ```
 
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #666, opened 2026-08-03
fix(datafusion): reduce scan planning peak memory

### LINKED ISSUE #665: DataFusion scan planning can exhaust memory by cloning split metadata
## Bug

For scans with many manifest entries and data files, the DataFusion planning path can retain multiple full copies of immutable split metadata:

- Table scan planning builds a Vec<DataSplit> containing DataFileMeta values.
- PaimonScanBuilder copies the plan with splits().to_vec(), which deep-clones the partition, bucket path, data files, deletion files, and row ranges.
- Asynchronous readers clone splits again before execution.
- Manifest reads collect every per-manifest result Vec before flattening them, and manifest ADD/DELETE merging allocates another output Vec.

The resulting peak memory is a multiple of the planned metadata size and can exhaust a query worker during planning, before it returns any rows.

## Expected behavior

Planning should transfer ownership where possible and share immutable split metadata across necessary clones. Manifest results should be merged incrementally and reuse their allocation without changing scan semantics.

## Compatibility requirements

- Preserve the public DataSplit getters.
- Preserve JSON and Java binary split serialization.
- Preserve manifest merge and scan behavior.

XiaoHongbo-Hope:
### Purpose

Closes #665.

Reduce peak memory during DataFusion scan planning. The previous path deep-cloned immutable `DataSplit` metadata and retained extra per-manifest result containers, which could multiply memory usage for large tables.

### Changes

- Consume `Plan` through `into_splits()` instead of cloning all planned splits.
- Store immutable `DataSplit` metadata in `Arc`-backed storage so necessary clones are shallow.
- Fold buffered manifest results into the aggregate incrementally and reuse the manifest-entry allocation during ADD/DELETE netting.
- Add tests for shared split metadata, ownership transfer, and serialization compatibility.

### Compatibility

No storage-format or serialized-split changes. Existing getters keep the same signatures, and JSON plus Java binary golden tests pass.

Planning memory still grows with the number of live files. This change removes multiplicative copies and intermediate containers; buffered manifest reads remain bounded at 64 concurrent operations.

### Validation

- Full GitHub CI passes on Linux, macOS, and Windows.
- Local source, table-scan, and DataFusion tests pass; formatting and strict Clippy checks pass.

### AI assistance

Prepared with Codex assistance.

### DIFF
--- a/crates/integrations/datafusion/src/full_text_search.rs
+++ b/crates/integrations/datafusion/src/full_text_search.rs
@@ -185,7 +185,7 @@ impl TableProvider for FullTextSearchTableProvider {
         PaimonScanBuilder {
             table,
             schema: &self.schema(),
-            plan: &plan,
+            plan,
             scan_trace: None,
             projection,
             pushed_predicate: None,
--- a/crates/integrations/datafusion/src/hybrid_search.rs
+++ b/crates/integrations/datafusion/src/hybrid_search.rs
@@ -287,7 +287,7 @@ impl TableProvider for HybridSearchTableProvider {
         let input = PaimonScanBuilder {
             table,
             schema: &input_schema,
-            plan: &plan,
+            plan,
             scan_trace: None,
             projection: Some(&input_projection),
             pushed_predicate: None,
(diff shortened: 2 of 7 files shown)


JingsongLi (approved):
+1

PR #667, opened 2026-08-04
test(vindex): make executor cancellation test deterministic

QuakeWang:
### Purpose

`cancelling_waiter_does_not_stop_submitted_task` waited for the releaser thread instead of the executor task itself. Under scheduling contention, it could read the completion flag before the task resumed, causing intermittent CI failures.

### Brief change log

- Replace the timing-based completion assertion with an explicit oneshot signal.
- Cancel the waiter before releasing the submitted task, then wait for actual task completion with a bounded timeout.

### Tests

### API and Format

### Documentation

Not required.


JingsongLi (approved):
+1

PR #669, opened 2026-08-04
docs: document the $files and $branches system tables

jackylee-ch:
`$files` and `$branches` are registered system tables — `SYSTEM_TABLE_NAMES` lists
both, and the 0.3.0 release notes already advertise `$files` — but the SQL
reference documented only 9 of the 11, so users had no way to learn the column
names without reading `system_tables/files.rs`. `$files` is the most useful of the
set for debugging compaction and skew, which makes the omission the more visible
one.

**Fix**: add a `### $files` and a `### $branches` section to `docs/src/sql.md`,
following the existing `$tags`/`$table_indexes` shape (purpose sentence, sample
query, column table). Column names and order are taken verbatim from
`files_schema()` (20 columns) and `branches_schema()` (2 columns); Arrow types are
rendered in the SQL spelling the surrounding sections already use. The
`$branches` section also notes the one behavior that sets it apart from its
neighbours: it ignores a `$branch_<name>` prefix and always reports the base
table's branches.

Docs only — no code or test changes.

### DIFF
--- a/docs/src/sql.md
+++ b/docs/src/sql.md
@@ -1732,6 +1732,24 @@ Columns:
 | `create_time` | TIMESTAMP | Tag creation time |
 | `time_retained` | STRING | Retention duration |
 
+### $branches
+
+View all branches of a table:
+
+```sql
+SELECT * FROM paimon.default.my_table$branches;
+```
+
+Columns:
+
+| Column | Type | Description |
+|---|---|---|
+| `branch_name` | STRING | Branch name |
+| `create_time` | TIMESTAMP | Branch creation time |
+
+Unlike the other system tables, `$branches` ignores a `$branch_<name>` prefix and
+always reports the branches of the base table.
+
 ### $manifests
 
 View manifest files of the latest snapshot:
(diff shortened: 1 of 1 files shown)


docs/src/sql.md
```diff
@@ -1754,6 +1772,39 @@ Columns:
...
+
+```sql
+SELECT * FROM paimon.default.my_table$files;
+```
+
+Columns:
+
+| Column | Type | Description |
+|---|---|---|
+| `partition` | STRING | Partition spec for the file, or `NULL` for unpartitioned tables |
```
QuakeWang: The Rust implementation returns `"{}"` for unpartitioned tables, not `NULL`, so this description does not match the actual `$files` output.

docs/src/sql.md
```diff
@@ -1754,6 +1772,39 @@ Columns:
...
+| `bucket` | INT | Bucket id the file belongs to |
+| `file_path` | STRING | Data file name |
```
QuakeWang: `file_path` contains the full data file path, using `external_path` when present, rather than only the data file name.

docs/src/sql.md
jackylee-ch: Right, `format_partition` returns `"{}"` when there are no partition fields. Fixed.

docs/src/sql.md
jackylee-ch: Correct — it is the full path, falling back to `split.data_file_path` when `external_path` is absent. Fixed.

jackylee-ch:
@QuakeWang PTAL

JingsongLi (approved):
+1

PR #670, opened 2026-08-04
fix(spec): reject changelog-producer on tables without primary keys

jackylee-ch:
A non-`none` `changelog-producer` on a table without primary keys was accepted at
both create and alter time. An append table has no merge step, so no changelog can
ever be produced: the option is persisted into the schema and then silently ignored
by the write path, which derives `input_changelog` from the producer alone and never
reaches a compaction that could emit changelog files. Users get a table that claims
to produce a changelog and quietly does not. Java rejects this in
`SchemaValidation#validateTableSchema` with "Can not set changelog-producer on table
without primary keys, please define primary keys."

**Fix**: add `validate_changelog_producer_requires_primary_keys` to
`Schema::validate_final_schema`, so it runs on both paths, as #660 arranged. Only a
non-`NONE` producer is rejected — an append table may still spell the default out
explicitly, matching Java's `!= ChangelogProducer.NONE` guard.

No behavior change for any table that has primary keys, or for `changelog-producer=none`.


QuakeWang (approved):
LGTM

PR #671, opened 2026-08-04
fix(spec): parse commit duration options and align their defaults with Java

jackylee-ch:
`commit.timeout`, `commit.min-retry-wait` and `commit.max-retry-wait` are
`durationType()` in Java, but Rust parsed bare milliseconds only: `commit.timeout =
'2 min'` failed to parse and silently fell back to the default, so a plausible
configuration was accepted and ignored. Two of the defaults also diverged —
min-retry-wait was 1000ms against Java's 10ms, and the timeout was 120000ms where
Java leaves it unset and lets `commit.max-retries` bound the loop alone. Those
compound: Rust slept ~75s across its retry budget where Java sleeps ~10s, and could
then abandon the commit on wall clock with "there may exist commit conflicts between
multiple jobs" while Java was still retrying.

**Fix**: add `parse_duration_millis`, mirroring `TimeUtils.parseDuration` and shaped
like the existing `parse_memory_size` — every Java unit label, case-insensitive,
bare number as millis, `None` on overflow. Align the two defaults, using `u64::MAX`
as the unbounded sentinel; the value is only ever compared against elapsed time, so
no arithmetic can overflow.

Bare-millisecond values keep working, since Java reads an empty unit label as millis too.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -95,8 +95,12 @@ const IGNORE_DELETE_FALLBACK_KEYS: &[&str] = &[
 const DIFF_PARALLELISM_OPTION: &str = "diff.parallelism";
 const DEFAULT_DIFF_PARALLELISM: usize = 4;
 const DEFAULT_COMMIT_MAX_RETRIES: u32 = 10;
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #674, opened 2026-08-04
feat(python): expose data evolution row ranges

XiaoHongbo-Hope:
## Summary

- expose `ReadBuilder.with_row_ranges()` for Python scan planning
- keep empty ranges as an explicit empty result instead of a full scan
- carry planned ranges in splits for local or distributed readers
- reject inverted ranges and declare the API in the Python type stub

This lets Python callers pass precomputed Data Evolution row ID ranges to the existing core planner and use its manifest and file pruning.

## Tests

- `python -m pytest tests/test_read.py -q` (59 passed)
- `cargo test --locked -p paimon empty_row_ranges_is_preserved` (2 passed)
- `cargo clippy --locked -p paimon --no-deps -- -D warnings`
- `cargo clippy --locked -p pypaimon_rust --no-deps -- -D warnings`
- `git diff --check`


QuakeWang (approved):
+1

JingsongLi:
+1

PR #677, opened 2026-08-04
feat(table): support scan.watermark batch time travel

### LINKED ISSUE #676: Support watermark-based batch time travel (scan.watermark)
### Search before asking

### Motivation

Java supports batch time travel by watermark via `scan.watermark` and `VERSION AS OF 'watermark-<value>'` (`StaticFromWatermarkStartingScanner` + `SnapshotManager.laterOrEqualWatermark`): it resolves the earliest snapshot whose watermark is greater than or equal to the requested value and scans it in full (`ScanMode.ALL`).

Rust already parses the `watermark` field on `Snapshot` (spec layer is complete), but `scan.watermark` is on the `CoreOptions::validate_scan_options` blocklist, so any table options carrying it are rejected with `Error::Unsupported` before any IO. This is also a Java-interop trap: Flink/Spark-written tables whose options include `scan.watermark` cannot be read from Rust at all.

### Solution

Implement watermark-based batch time travel, mirroring Java semantics:

1. `SnapshotManager::later_or_equal_watermark(watermark)`: return the earliest snapshot with `watermark >= requested`, or `None`. Snapshots without a watermark are skipped — both `None` and `Some(i64::MIN)`, since Java writers (Flink) use `Long.MIN_VALUE` as the no-watermark sentinel. Binary search over the actual snapshot id list, tolerating id gaps from deleted snapshots (same pattern as the existing `later_or_equal_time_millis`).
2. `CoreOptions`: parse `scan.watermark` as a first-class `TimeTravelSelector` — mutually exclusive with `scan.timestamp-millis` / `scan.version` / `scan.snapshot-id` / `scan.tag-name`, strict i64 parsing. Remove it from the unsupported blocklist, and accept it under `scan.mode=from-snapshot` (Java's `CoreOptions.startupMode()` maps `scan.watermark` to `FROM_SNAPSHOT`).
3. `travel_to_snapshot`: resolve both direct `scan.watermark` and Java-compatible `scan.version=watermark-<value>` (tag first, then watermark prefix, then snapshot id); when no snapshot matches, fail at scan planning with Java's message (`There is currently no snapshot later than or equal to watermark[...]`), preserving Java `tryTravelToSnapshot`'s silent-fallback behavior at `copy_with_time_travel` time.
4. `Table::copy_with_options`: changing `scan.watermark` invalidates the cached resolved snapshot, like the other selectors.
5. Docs: `docs/src/sql.md` Time Travel section documents both Java-compatible `VERSION AS OF 'watermark-<value>'` and session-scoped `SET 'paimon.scan.watermark'`.

One deliberate difference from the current Java implementation is documented in code: Java's binary search can retain the raw midpoint snapshot after walking backward over missing watermark metadata. The Rust implementation only returns a snapshot whose own effective watermark satisfies `watermark >= requested`, preserving the selector contract when watermark metadata is sparse.

### Anything else?

No response

### Willingness to contribute

u70b3:
### Purpose

Linked issue: close #676

Support watermark-based batch time travel, mirroring Java's `StaticFromWatermarkStartingScanner` and `TimeTravelUtil.adaptScanVersion`: direct `scan.watermark` and `VERSION AS OF 'watermark-<value>'` resolve the earliest snapshot whose watermark is greater than or equal to the requested value and scan it in full. Today `scan.watermark` is on the `validate_scan_options` blocklist, so Java-written tables carrying it cannot be read from Rust at all.

### Brief change log

- `SnapshotManager::later_or_equal_watermark`: binary search over the actual snapshot id list (gap-tolerant, same pattern as `later_or_equal_time_millis`). Snapshots without a watermark are skipped — both `None` and `Some(i64::MIN)`, since Flink writers use `Long.MIN_VALUE` as the no-watermark sentinel.
- `CoreOptions`: `scan.watermark` becomes a first-class `TimeTravelSelector` (mutual exclusion with the other selectors, strict i64 parsing); removed from the unsupported scan-option blocklist; accepted under `scan.mode=from-snapshot` (Java's `startupMode()` maps it to `FROM_SNAPSHOT`).
- `scan.version`: mirror Java's tag-first resolution order — existing tag → `watermark-<value>` → snapshot id — enabling `VERSION AS OF 'watermark-<value>'` in DataFusion SQL.
- `travel_to_snapshot`: direct and version-prefixed watermark selectors share one resolver; no match fails at scan planning with Java's message, while `copy_with_time_travel` keeps Java's silent-fallback behavior.
- `Table::copy_with_options`: changing `scan.watermark` invalidates the cached resolved snapshot.
- C bindings: the `unsupported_scan_option_is_rejected` test now uses `incremental-between` as its example (`scan.watermark` is supported now).
- `docs/src/sql.md`: new "By Watermark" subsection documenting both `VERSION AS OF 'watermark-<value>'` and the dynamic option.

Assumptions / deviations to be aware of (per the AI-assisted PR policy):

- **Deliberate difference from the current Java implementation**: Java's binary search can retain the raw midpoint snapshot after walking backward over missing watermark metadata. This implementation only returns a snapshot whose own effective watermark satisfies `watermark >= requested`, preserving the selector contract when watermark metadata is sparse. Documented in the method's doc comment.
- Watermarks are assumed non-decreasing in snapshot order (guaranteed by Flink/Java writers); the binary search relies on this.
- Rust's own commit path never writes watermarks, so the new tests commit `Snapshot`s with watermarks directly through `SnapshotManager`; no write-path changes are included.

### Tests

- `cargo clippy --locked --all-targets -p paimon -p paimon-datafusion -- -D warnings`
- `cargo test --locked -p paimon --lib` — 2101 passed, 0 failed, 1 ignored (11 new watermark tests: sentinel/missing/duplicate watermarks, exact/between/out-of-range matches, snapshot-id gaps, selector mutual exclusion, Java `watermark-<value>` resolution with tag precedence, cache invalidation, and scan-time errors)
- `cargo test --locked -p paimon-datafusion --test time_travel_schema_tests` — 6 passed, including an end-to-end `VERSION AS OF 'watermark-<value>'` test matching Java's 1/9/10/no-match cases
- `cargo test --locked -p paimon-c unsupported_scan_option` — 1 passed

### API and Format

No storage format changes. The public Rust API gains `SnapshotManager::later_or_equal_watermark`, and `SCAN_WATERMARK_OPTION` becomes public alongside the other selector constants.

### Documentation

`docs/src/sql.md` Time Travel section gains a "By Watermark" subsection documenting both `VERSION AS OF 'watermark-<value>'` and the session-scoped `SET 'paimon.scan.watermark'` dynamic option.

### DIFF
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -227,8 +227,9 @@ pub unsafe extern "C" fn paimon_table_free(table: *mut paimon_table) {
 }
 
 /// Time-travel selector option names, in the core's resolution priority order.
-const TIME_TRAVEL_SELECTORS: [&str; 4] = [
+const TIME_TRAVEL_SELECTORS: [&str; 5] = [
     "scan.timestamp-millis",
+    "scan.watermark",
     "scan.version",
     "scan.snapshot-id",
     "scan.tag-name",
@@ -300,8 +301,9 @@ pub unsafe extern "C" fn paimon_table_new_read_builder(
 
 /// Create a ReadBuilder from a Table with scan options (e.g. time-travel
 /// selectors `scan.snapshot-id` / `scan.tag-name` / `scan.timestamp-millis` /
-/// `scan.version`). At most one time-travel selector may be set. A selector that
-/// does not resolve to a snapshot is an error (never a silent read-of-latest).
+/// `scan.watermark` / `scan.version`). At most one time-travel selector may be
+/// set. A selector that does not resolve to a snapshot is an error (never a
+/// silent read-of-latest).
 ///
 /// # Safety
 /// `table` must be a valid pointer. `options` must be a valid pointer to
@@ -2292,6 +2294,48 @@ mod tests {
             assert!(
                 message.contains("scan.snapshot-id") && message.contains("scan.tag-name"),
                 "message should name both selectors, got: {message}"
+            );
+            paimon_table_free(table);
+        }
+    }
+
+    #[test]
+    fn watermark_conflicting_with_other_selector_is_rejected() {
+        unsafe {
+            let table = boxed_test_table();
+            let k1 = CString::new("scan.watermark").unwrap();
+            let v1 = CString::new("1").unwrap();
+            let k2 = CString::new("scan.snapshot-id").unwrap();
+            let v2 = CString::new("1").unwrap();
+            let opts = [opt(&k1, &v1), opt(&k2, &v2)];
+            let (code, message) = assert_rb_err_code_message(
+                paimon_table_new_read_builder_with_options(table, opts.as_ptr(), 2),
+            );
+            assert_eq!(code, PaimonErrorCode::InvalidInput as i32);
+            assert!(
+                message.contains("scan.watermark") && message.contains("scan.snapshot-id"),
+                "message should name both selectors, got: {message}"
+            );
+            paimon_table_free(table);
+        }
+    }
+
+    #[test]
+    fn unresolved_watermark_does_not_silently_read_latest() {
+        unsafe {
+            // The test table commits no watermarks, so any watermark selector
+            // is unresolvable; the binding must error instead of falling back.
+            let table = boxed_test_table();
+            let k = CString::new("scan.watermark").unwrap();
+            let v = CString::new("1").unwrap();
+            let opts = [opt(&k, &v)];
+            let (code, message) = assert_rb_err_code_message(
+                paimon_table_new_read_builder_with_options(table, opts.as_ptr(), 1),
+            );
+            assert_eq!(code, PaimonErrorCode::InvalidInput as i32);
+            assert!(
+                message.contains("did not resolve"),
+                "message should report the selector did not resolve, got: {message}"
             );
             paimon_table_free(table);
         }
--- a/bindings/c/src/table.rs
+++ b/bindings/c/src/table.rs
@@ -2301,8 +2301,8 @@ mod tests {
     fn unsupported_scan_option_is_rejected() {
         unsafe {
             let table = boxed_test_table();
-            let k = CString::new("scan.watermark").unwrap();
-            let v = CString::new("0").unwrap();
+            let k = CString::new("incremental-between").unwrap();
+            let v = CString::new("1,2").unwrap();
             let opts = [opt(&k, &v)];
             // Core's validate_scan_options rejects this before resolution; the
             // binding surfaces core's Unsupported code.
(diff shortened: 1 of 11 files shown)


JingsongLi:
The watermark lookup logic differs from the Java implementation. When a snapshot's watermark is `Long.MIN_VALUE`, the Java implementation returns `null` if the latest snapshot hits this sentinel value, whereas the Rust implementation skips it and returns an earlier snapshot. The error—returning snapshot 1—has been reproduced by querying for `100` with the range `[100, MIN_VALUE]`.

u70b3:
Thanks for checking this case — I can confirm the reproduction: for `[100, Long.MIN_VALUE]` querying `100`, this implementation returns snapshot 1 while released Java returns `null`. This is the deliberate difference documented in the PR description, so let me explain the reasoning and how I'd propose to sequence it.

The Rust behavior implements the semantics of apache/paimon#9037 rather than released Java: `null` and `Long.MIN_VALUE` are treated uniformly as "missing", and there is no early return based on the latest snapshot's value. A latest-value check is unsound for mixed histories — `[100, Long.MIN_VALUE]` is exactly that shape: the latest snapshot carries no watermark, yet an earlier real watermark satisfies the query, so the selector contract requires snapshot 1 to be found. This is the same conclusion as your review on #9037: "[100, null] must still find the earlier valid watermark." The sentinel case follows identically once both are treated as missing. (For `[100, null]` there is in fact no released-Java behavior to mirror at all — the fast path unboxes `snapshot(latest).watermark()` and NPEs.)

So the two PRs are a pair: #9037 fixes the Java searches, and this PR implements the same semantics in Rust from the start, so no released Rust version ever disagrees with Java on this lookup. The other documented deviation (returning the walked-back watermark-bearing snapshot instead of the raw midpoint, which may carry no watermark at all) is likewise aligned with #9037.

My proposal: hold this PR until #9037 is merged — at that point the Rust behavior matches the Java implementation by construction, including this exact case. If you'd prefer not to gate Rust on the Java fix, I can instead reinstate the latest-sentinel early return here to mirror released Java, and drop it in a follow-up once #9037 lands. Which sequencing would you prefer?

u70b3:
apache/paimon#9037 has been merged (commit f7d9593) — thanks for shepherding it through, @JingsongLi!

With that, the case you raised is now aligned across both implementations: merged Java treats `null` and `Long.MIN_VALUE` uniformly as missing via `isMissingWatermark`, with no early return on the latest snapshot's value. So `[100, Long.MIN_VALUE]` querying `100` returns snapshot 1 in Java as well — identical to this PR's `later_or_equal_watermark`, which skips both `None` and `Some(i64::MIN)` the same way. The Rust implementation now matches Java by construction, as planned.

CI is all green. Could you take another look when you have a moment? Would love to get this in.

JingsongLi (approved):
+1

PR #682, opened 2026-08-05
fix(datafusion): stream BTree global-index filters during execution

XiaoHongbo-Hope:
Recreates the execution-time streaming optimization from #678 directly on the latest main. This PR does not contain or depend on #672 and does not push LIMIT into global-index planning.

What changed:
- Stream exact single-leaf BTree Eq, StartsWith, and Like row IDs in fixed-size execution batches.
- Keep the DataFusion residual filter, so normal Limit can stop after enough valid rows while TopK still consumes all shards.
- Assign overlapping shard coverage to disjoint row-id intervals without a query-wide deduplication bitmap.
- Reuse prepared snapshot and manifest-list metadata across row-id batches, with a bounded manifest-content cache.
- Reject oversized encoded or decompressed BTree blocks before allocation.

Out of scope:
- Bitmap global indexes.
- Compound global-index predicates.
- Planner-time LIMIT pushdown or candidate early stop.

Validation:
- 89 BTree-related paimon tests passed.
- 45 global-index scanner tests passed.
- Prepared snapshot manifest reuse regression passed.
- DataFusion equality LIMIT, prefix LIKE LIMIT, and filtered ORDER BY LIMIT regression passed.
- cargo clippy for paimon and paimon-datafusion all targets passed with warnings denied.
- cargo fmt and git diff check passed.

Related to #665 and supersedes the closed #678 implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4607,6 +4607,7 @@ dependencies = [
  "arrow-array",
  "arrow-schema",
  "arrow-select",
+ "async-stream",
  "async-trait",
  "bytes",
  "chrono",
--- a/crates/integrations/datafusion/Cargo.toml
+++ b/crates/integrations/datafusion/Cargo.toml
@@ -34,6 +34,7 @@ vortex = ["paimon/vortex"]
 
 [dependencies]
 arrow-select = { workspace = true }
+async-stream = "0.3.6"
 async-trait = "0.1"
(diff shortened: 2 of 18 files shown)


XiaoHongbo-Hope:
Closing this duplicate. The clean implementation will continue in #678 as requested.

PR #683, opened 2026-08-06
fix(python): declare abort and default_database in the type stub

jackylee-ch:
`TableCommit.abort` (`write.rs:239`) and `register_catalog`'s `default_database`
(`context.rs:420`) are implemented and tested but absent from the type stub, so type
checkers reject both. Adds them, plus a docstring for `with_limit` noting it is a
planning hint rather than an exact cap.


JingsongLi (approved):
+1

PR #684, opened 2026-08-06
fix(spec): reject table-read.sequence-number.enabled without primary keys

jackylee-ch:
`table-read.sequence-number.enabled` was accepted on tables without primary keys. The
sequence number lives in the merge key, so an append table has no such column: the read
path injects a field id no data file carries and the column comes back entirely NULL
instead of raising.

**Fix**: validate at create and alter time, mirroring Java
`SchemaValidation#validateChangelogReadSequenceNumber`. Only `true` is rejected, so an
append table may still set it to `false` explicitly.

The option key moves to `core_options.rs` because `spec` does not depend on `table`; the
two existing literals are left alone.


JingsongLi (approved):
+1

PR #688, opened 2026-08-07
test(spec): cover untested partial-update and blob-view error paths

jackylee-ch:
Four `PartialUpdateConfig` branches and eight `BlobViewStruct::deserialize` branches have
no test, while the siblings right next to them are covered. `validated_sequence_groups`
tests the primary-key conflict and the duplicate-protected-field case but not "field does
not exist in the table schema"; `validated_aggregate_functions` tests the per-field
unknown function but not the `fields.default-aggregate-function` one, and never reaches
either empty-field-name guard (`fields..list-agg-delimiter`). `BlobViewStruct` has three
tests for eight error branches.

**Tests only, no production change.** A `payload()` helper assembles the BlobViewStruct
wire format from parts so each header field can be corrupted independently — a negative
identifier length cannot be produced through `serialize()`.

Each new test was confirmed to fail with its guard disabled and to pass once restored.

`blob_descriptor.rs` still has four uncovered branches; happy to fold them in if you would
rather have both files done at once.

### DIFF
--- a/crates/paimon/src/spec/blob_view_struct.rs
+++ b/crates/paimon/src/spec/blob_view_struct.rs
@@ -214,4 +214,127 @@ mod tests {
             matches!(err, Error::DataInvalid { message, .. } if message.contains("trailing bytes"))
         );
     }
+
+    /// Assemble a payload from parts so that individual header fields can be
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #690, opened 2026-08-07
fix(arrow): list every writable format in the write-format error

jackylee-ch:
`create_format_writer` says `expected .parquet, .row` when it rejects a path, but the same
function also dispatches `.blob` and, under the `vortex` feature, `.vortex`. A caller
passing an unsupported extension is told two of the four accepted ones.

**Fix**: derive the message from a `supported_write_formats()` list, mirroring the read
side, which already builds its error from `supported_read_formats()`. The list carries the
same `#[cfg(feature = "vortex")]` gate as the dispatch arm, so the message stays accurate
in both builds.

The new test asserts every entry of that list appears in the message, so a future writer
arm cannot drift from it again. It fails on the old literal.

### DIFF
--- a/crates/paimon/src/arrow/format/mod.rs
+++ b/crates/paimon/src/arrow/format/mod.rs
@@ -230,6 +230,16 @@ fn supported_read_formats() -> Vec<&'static str> {
     ]
 }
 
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
+1

PR #691, opened 2026-08-07
fix(auth): Fail closed on query-auth tables outside the read boundary

plusplusjiajia:
### Purpose

  `CoreOptions::ensure_read_authorized()` refuses a `query-auth.enabled` table because this client cannot enforce the server's row filter or column masking. It sits on the read boundary, so paths that read
  without crossing it are never gated: rewrites, index builds and drops, bucket assignment, commits, and partition listing. `Table::partition_stats()` is the clearest case — public, reaching Python, returning the
  partition values and column stats the DataFusion `$partitions` table already refuses to serve.

  ### Brief change log

  The gate is added at each entry point, at the narrowest one that still has the table. Two placements are worth a look:

  - `TableWrite::new` rather than `DynamicBucketAssigner`, which gets only a `FileIO` and a location. It refuses every write, not just dynamic-bucket ones — the commit is blocked anyway, so failing at open avoids
  writing files that can never be committed.
  - `list_partitions_from_file_system` rather than its three callers, since it is publicly re-exported and `$partitions` retries it directly when the catalog call fails.

  Left ungated on purpose: `TableCommit::abort`, which deletes only the files the caller just wrote, and path-based readers like `Manifest::read`, which have no table options to consult.

### DIFF
--- a/crates/paimon/src/catalog/partition_listing.rs
+++ b/crates/paimon/src/catalog/partition_listing.rs
@@ -32,6 +32,8 @@ use crate::Result;
 /// Scan a table's manifest entries and aggregate them into [`Partition`] rows,
 /// matching the shape catalogs would otherwise return from a metastore.
 pub async fn list_partitions_from_file_system(table: &Table) -> Result<Vec<Partition>> {
+    // Manifests carry partition values and per-column stats.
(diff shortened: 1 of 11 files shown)


JingsongLi (approved):
+1

PR #692, opened 2026-08-07
fix(scan): preserve partition group order

XiaoHongbo-Hope:
### Purpose

`PaimonTableScan` grouped manifest entries by partition and bucket in a `HashMap`. Iterating that map produced a process-dependent split order, so LIMIT pushdown could retain different split groups for the same manifest entries.

Use nested `IndexMap`s to preserve first-seen partition order and first-seen bucket order within each partition, mirroring Java's nested `LinkedHashMap`s. This makes split planning deterministic without promising query result order when no ordering is requested.

### Tests

- Added interleaved partition/bucket grouping order coverage
- `table_scan::tests`: 82 passed


QuakeWang (approved):
LGTM

PR #693, opened 2026-08-08
fix(scan): respect family-specific global-index search modes

XiaoHongbo-Hope:
### Purpose

Rust scans only honored the legacy `global-index.search-mode` option. Paimon's family-specific options therefore had no effect: for example, `scalar-index.search-mode=full` silently used FAST mode and could omit rows outside current scalar-index coverage.

### Changes

- Resolve scalar, vector, and full-text search modes with the family option taking precedence and `global-index.search-mode` remaining as the compatibility fallback.
- Use the corresponding family mode in scalar scans, vector search, and full-text/hybrid search.
- Add a scalar FULL regression that builds a BTree index, appends matching unindexed rows, and verifies both planning ranges and returned rows.
- Exercise family-specific vector and full-text options in existing fallback tests.

### Tests

- `cargo test -p paimon spec::core_options::tests -- --nocapture` (66 passed)
- `cargo test -p paimon table::btree_global_index_build_builder::tests -- --nocapture` (29 passed)
- Targeted vector, full-text, and hybrid search-mode tests
- `cargo clippy -p paimon --all-targets --features fulltext -- -D warnings`
- `git diff --check`


QuakeWang (approved):
LGTM

JingsongLi:
+1

PR #694, opened 2026-08-08
fix(arrow): enforce exact Mosaic predicate filtering

QuakeWang:
### Purpose

Mosaic only used row-group statistics for predicate pushdown, so core reads could return non-matching rows. Residual evaluation could also panic or fail open when predicates referenced physically missing columns.

### Brief change log

  - Apply shared row-level residual filtering after Mosaic row-group pruning.
  - Include non-projected predicate columns in the scan.
  - Remap predicates against physical columns so missing columns follow NULL semantics.
  - Add regression coverage for missing columns, projection, deletion vectors, and direct core reads.

### Tests

  - `cargo test --locked -p paimon-datafusion --test mosaic_tables`


JingsongLi (approved):
+1

PR #695, opened 2026-08-08
fix(storage): retry temporary OSS failures

XiaoHongbo-Hope:
### Purpose

OSS can return temporary failures such as HTTP 503. The OSS service classifies these errors as temporary, but the operator had no retry layer, so metadata and data operations failed immediately.

### Changes

- Attach OpenDAL RetryLayer to OSS operators with exponential backoff and jitter.
- Support `fs.oss.retry.count` (default `10`) and `fs.oss.retry.interval.millisecond` (default `500`).
- Add a regression test that returns OSS `QpsLimitExceeded` once and verifies the read succeeds on retry.
- Document the retry options.

### Tests

- `cargo test --offline -p paimon io::storage_oss::tests`
- `cargo check --offline -p paimon --no-default-features --features storage-oss`
- `cargo clippy --offline -p paimon --no-default-features --features storage-oss -- -D warnings`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -726,6 +726,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "backon"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


QuakeWang (approved):
+1

PR #696, opened 2026-08-08
feat(storage): support Jindo for OSS scan planning

XiaoHongbo-Hope:
### Purpose

Add an opt-in, read-only JindoSDK backend for OSS scans. OpenDAL remains the default. PyPaimon can select it for native planning through apache/paimon#9124.

### Changes

- add the optional `storage-jindo` feature, selected with `fs.oss.impl=jindo`
- dynamically load an installed JindoSDK without bundling its binary
- preserve HTTP/HTTPS endpoint protocol when adapting Jindo configuration
- support stat, paginated list, and ranged/full reads
- fetch list pages on demand and convert Jindo list exceptions into Rust errors
- initialize lazily and run blocking SDK calls on Tokio blocking workers
- reject unsupported write operations explicitly

### Benchmark

Release-mode, partition-filtered planning on two fixed OSS-backed snapshots. Both backends ran on the same host with identical predicates, in reversed order and two independent processes per backend.

| Workload | Planned scale | OpenDAL warm median | Jindo warm median |
| --- | ---: | ---: | ---: |
| A | ~281k files | 1.55s | 1.25s |
| B | ~37k files / ~27k splits | 5.92s | 2.19s |

Both backends produced matching snapshot, split, file-count, and planned-byte results. This measures scan planning, not data throughput. Request count and peak OSS QPS were not measured.

### Validation

- 6 Jindo storage tests and Python SDK discovery test
- default OSS and `storage-jindo` builds
- Clippy, formatting, dependency report, and diff checks
- read-only OSS REST-catalog test covering snapshot listing, planning, and row reading
- JindoSDK 6.10.4 local OSS-protocol probes: HTTPS emits TLS, list exceptions return errors, three-page listing is complete, and consuming one entry does not request later pages

### Notes

No JindoSDK binary is packaged. Selecting Jindo without the feature or a loadable SDK returns a configuration error. A FileIO configured with Jindo rejects writes, deletes, renames, and copies.

### DIFF
--- a/crates/paimon/src/io/storage_jindo.rs
+++ b/crates/paimon/src/io/storage_jindo.rs
@@ -25,6 +25,7 @@ use libloading::Library;
 use opendal::raw::*;
 use opendal::{Buffer, Builder, BytesRange, Capability, EntryMode, ErrorKind, Metadata};
 use opendal::{Error as OpendalError, OperationContext, Operator, Result as OpendalResult};
+use tokio::sync::OnceCell;
 
 use crate::error::Error;
 use crate::Result;
@@ -55,9 +56,16 @@ type HandleErrorMessage = unsafe extern "C" fn(JdoPtr) -> *const c_char;
 type Init = unsafe extern "C" fn(JdoPtr, *const c_char);
 type GetFileStatus = unsafe extern "C" fn(JdoPtr, *const c_char, JdoPtr) -> JdoPtr;
 type FreeFileStatus = unsafe extern "C" fn(JdoPtr);
+type FileStatusPath = unsafe extern "C" fn(JdoPtr) -> *const c_char;
 type FileStatusType = unsafe extern "C" fn(JdoPtr) -> i8;
 type FileStatusSize = unsafe extern "C" fn(JdoPtr) -> i64;
 type FileStatusMtime = unsafe extern "C" fn(JdoPtr) -> i64;
+type ListDir = unsafe extern "C" fn(JdoPtr, *const c_char, bool, JdoPtr) -> JdoPtr;
+type FreeListDirResult = unsafe extern "C" fn(JdoPtr);
+type ListDirResultSize = unsafe extern "C" fn(JdoPtr) -> i64;
+type IsListDirResultTruncated = unsafe extern "C" fn(JdoPtr) -> bool;
+type ListDirResultNextMarker = unsafe extern "C" fn(JdoPtr) -> *const c_char;
+type ListDirFileStatus = unsafe extern "C" fn(JdoPtr, usize) -> JdoPtr;
 type GetObject = unsafe extern "C" fn(JdoPtr, *const c_char, *mut c_char, i64, i64, JdoPtr) -> i64;
 
 #[derive(Clone)]
@@ -137,10 +145,9 @@ impl Builder for JindoBuilder {
         let config = self.config.ok_or_else(|| {
             OpendalError::new(ErrorKind::ConfigInvalid, "missing Jindo configuration")
         })?;
-        let client = JindoClient::new(&config, &self.bucket)?;
         Ok(JindoService {
             info: ServiceInfo::new("jindo", "/", &self.bucket),
-            client: Arc::new(client),
+            client: Arc::new(LazyJindoClient::new(config, self.bucket)),
         })
     }
 }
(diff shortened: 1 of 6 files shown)


JingsongLi:
Have you compared Jindo and OpenDAL? What are the differences? (benchmark)

XiaoHongbo-Hope:
> Have you compared Jindo and OpenDAL? What are the differences? (benchmark)

yes, I did a test, Jindo performance is a little better, but there is a lot of qps issue during my testing, so may need another testing. And, I think this PR is best to have, not very urgent now. Because qps issue is addressed in other way.

XiaoHongbo-Hope:
> Have you compared Jindo and OpenDAL? What are the differences? (benchmark)

I updated the performance result in PR descrption.

QuakeWang (approved):
LGTM

JingsongLi:
Requirement fit: **SUPPORTED**. Implementation: **FINDINGS**, reviewed at `45e887beb14cd1f67b1ae3ca5bfb10b26fdb21bc`.

The opt-in planning backend has a concrete consumer in the merged apache/paimon#9124, and the supplied measurements demonstrate an end-to-end planning benefit for the two tested snapshots. Keeping the claim scoped to planning is appropriate: the reported data reads are slightly slower and request/QPS effects remain unmeasured.

**[P2] Preserve explicit SDK environment selection before Python auto-discovery.** In [`bindings/python/src/context.rs:122-124`](https://github.com/apache/paimon-rust/blob/45e887beb14cd1f67b1ae3ca5bfb10b26fdb21bc/bindings/python/src/context.rs#L122-L124), a user who sets `JINDOSDK_LIBRARY_PATH` (or `JINDOSDK_HOME`) and has `pyjindo` installed gets `fs.jindo.library.path` injected from the Python package. `library_candidates` then treats that injected value as an exclusive explicit path and never consults the user's environment. Thus the documented SDK selection works for Rust callers but selects a different SDK through Python; if the package's library is not loadable, even a valid environment-selected SDK cannot be used. Please skip package discovery when either explicit SDK environment setting is present, or make package discovery a fallback after those settings, and cover this precedence in a binding test.

I traced configuration, lazy initialization, paginated listing, ranged reads, read-only operations, and the FileIO callers. `cargo +1.94.0 test -p paimon --features storage-jindo --lib storage_jindo` passes all 6 tests locally. I did not rerun the native SDK/OSS probes or the reported benchmarks; these unit tests do not validate the dynamic ABI or live storage behavior.

crates/paimon/src/io/storage_jindo.rs
```diff
@@ -0,0 +1,1092 @@
...
+    }
+
+    #[test]
+    fn test_build_does_not_load_sdk() {
+        let config = JindoStorageConfig {
+            library_path: Some(PathBuf::from("/missing/libjindosdk_c.so")),
+            user: DEFAULT_JINDO_USER.to_string(),
+            properties: HashMap::new(),
+        };
+        let operator = jindo_config_build(&config, "bucket").unwrap();
```
QuakeWang, line 1073: Could we add a test that actually loads the SDK and covers stat, paginated listing, and ranged reads? The current tests stop before exercising those paths, so regressions there would not be caught.

JingsongLi (commented):
Requirement fit: **SUPPORTED**. The opt-in backend has a native planning consumer and the reported same-snapshot measurements show a planning benefit. The claim should remain scoped to planning; the reported data reads are slower and OSS request/QPS effects were not measured.

Implementation: **CLEAN** at `1975653c9129f18582738d907c53e7197df1f2a9` for the code paths I could verify. The latest commit fixes my earlier SDK path precedence finding: explicit catalog and `JINDOSDK_*` settings now take precedence over Python package discovery. I traced selection, lazy initialization, stat/list/range handling, read-only behavior, and cleanup again. I found no new actionable code defect.

Verification: `cargo test -p paimon --features storage-jindo --lib storage_jindo` passed (6 tests); `cargo test -p pypaimon_rust --lib test_pyjindo_discovery_precedence` passed; the current head's CI succeeded. **Production validation remains pending:** there is no Jindo SDK installed in this review environment, and the committed tests still do not load one. The SDK-backed stat, paginated listing, ranged read, and failure cases raised in QuakeWang's current review should be exercised before treating this backend as production-ready. The PR description's local probes are useful evidence, but they are not a repeatable regression gate.

PR #698, opened 2026-08-09
fix(scan): prune primary-key files with key stats

QuakeWang:
### Purpose

Rust key-value files store primary-key statistics in `key_stats`, while scan pruning only evaluated `value_stats`. As a result, Rust-written primary-key files with empty value stats could not be pruned by primary-key predicates.

### Brief change log

  - Evaluate file-safe primary-key predicates against `key_stats`.
  - Resolve key-stat layouts using trimmed primary-key order across schema evolution.
  - Preserve fail-open behavior for missing or invalid statistics.
  - Add regression tests for Rust-written files, composite keys, schema evolution, and corrupted stats.

### Tests

### DIFF
--- a/crates/paimon/src/table/pk_vector_scan.rs
+++ b/crates/paimon/src/table/pk_vector_scan.rs
@@ -752,17 +752,11 @@ mod tests {
             )
         }
 
-        /// Build a real single-file primary-key table via the public write path, in a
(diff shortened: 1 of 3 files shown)


XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #699, opened 2026-08-10
[core] Reduce global index planning IO

XiaoHongbo-Hope:
## Purpose

Align Rust global-index planning with Java and PyPaimon to avoid unnecessary object-store IO and unrelated index candidates.

## Changes

- prune sorted global-index manifest entries with the scan partition filter before evaluating index shards, matching Java and PyPaimon
- use the file size already stored in `IndexFileMeta`, falling back to metadata lookup for legacy or invalid sizes, matching `GlobalIndexIOMeta` in Java and PyPaimon
- keep deletion-vector selection and reader cache behavior unchanged

## Performance

On a partitioned data-evolution table with an equality lookup and a partition predicate:

- native FAST planning: 8.18s -> 2.93s
- retained index entries: 762 -> 176
- indexed candidate ranges: 4 -> 1
- peak RSS in a plan-plus-explain benchmark: about 2.45 GiB -> 586 MiB

The result split and row range were unchanged.

## Tests

- `cargo test -p paimon global_index_scanner --lib` (46 passed)
- `cargo test -p paimon table_scan::tests --lib` (84 passed)
- `git diff --check`

### DIFF
--- a/crates/paimon/src/table/global_index_scanner.rs
+++ b/crates/paimon/src/table/global_index_scanner.rs
@@ -122,7 +122,7 @@ struct GlobalIndexScanResult {
 /// The scanner filters index manifest entries for global index files,
 /// uses BTreeIndexMeta for file-level pruning, then reads matching
 /// BTree or bitmap files to evaluate predicates and collect row IDs.
-/// Opened BTreeIndexReaders are cached for reuse across evaluations.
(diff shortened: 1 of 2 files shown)


QuakeWang (approved):
LGTM

JingsongLi (approved):
+1

PR #700, opened 2026-08-10
fix(table): reject row ranges for format tables at read construction

jackylee-ch:
`FormatTableScan::plan` rejects row ranges, but `FormatReadBuilder::new_read` never reads
`self.row_ranges`, so `with_row_ranges(..).new_read()` on a format table silently ignores the
request instead of reporting it. Reachable from core Rust callers, which reach
`FormatReadBuilder` through `ReadBuilder::new_read`.

**Fix**: fail closed in `new_read` too, matching `ReadBuilder::new_read`, which guards the
same way for the same reason.

### DIFF
--- a/crates/paimon/src/table/format_read_builder.rs
+++ b/crates/paimon/src/table/format_read_builder.rs
@@ -127,6 +127,13 @@ impl<'a> FormatReadBuilder<'a> {
     }
 
     pub(crate) fn new_read(&self) -> Result<TableRead<'a>> {
+        // Fail closed here as well as in `FormatTableScan::plan`, so a caller that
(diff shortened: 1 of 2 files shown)


JingsongLi:
The Python bindings did not pass `row_ranges` from `PyReadBuilder` to `PyTableRead`.

jackylee-ch:
Right, and `PyTableRead` should not carry them: core `new_read` does not consume `row_ranges`
either — they reach readers through `DataSplit`, set at planning. The guard is for core Rust
callers, not the binding: a format table can neither honour nor report the request.

JingsongLi (approved):

PR #701, opened 2026-08-10
docs: correct five system-table column descriptions

jackylee-ch:
Five system-table column descriptions do not match what the code returns:

- `$tags.create_time` / `.time_retained` are documented as carrying data, but `tags.rs:117`
  fills both with `new_null_array` — the Rust snapshot has no tag creation time or retention.
- `$partitions.created_by` / `updated_by` are documented as snapshot ids; they are the audit
  user (`spec/partition.rs:40`).
- `$table_indexes.partition` is documented as `NULL` for unpartitioned tables; it is `"{}"`.
- `$physical_files_size` counts everything under a bucket directory as data, but
  `referenced_files.rs:677` excludes names starting with `index-`.

The `$table_indexes` one is the same mistake QuakeWang caught for `$files` in #669.

### DIFF
--- a/docs/src/sql.md
+++ b/docs/src/sql.md
@@ -1729,8 +1729,8 @@ Columns:
 | `schema_id` | BIGINT | Schema ID |
 | `commit_time` | TIMESTAMP | Commit time |
 | `record_count` | BIGINT | Record count |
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #702, opened 2026-08-10
fix(table): list the supported formats when a format table is rejected

jackylee-ch:
A format table with an unreadable `file.format` reports only the rejected value, so the user
cannot tell which formats are available — and the sets differ from Java in both directions
(Java has CSV/TEXT/JSON, Rust has `.row`), so the Java docs do not answer it either. Java's
`FormatTable.parseFormat` appends the supported list.

**Fix**: build the message from a `supported_format_table_formats()` helper, as #690 just did
for the write side. Also documents the `global-index.thread-num` upper bound added by #647,
which the doc comment and the option table still omit.

Adding CSV/TEXT/JSON support is out of scope. Adds the first tests to `format_table_scan.rs`.


JingsongLi (approved):
+1

PR #703, opened 2026-08-10
perf(scan): prune index manifest entries

QuakeWang:
### Purpose

Scan planning retained index-manifest entries that could not affect the query, including DELETE entries, deletion vectors from unselected partitions, and sorted global indexes unrelated to predicate fields. These entries were ignored later but still increased planning CPU and memory usage.

### Brief change log

  - Retain only live entries required by active index consumers.
  - Apply partition pruning to sorted global indexes and deletion vectors.
  - Select sorted indexes by predicate field IDs, including extra fields.
  - Preserve fail-loud validation for retained sorted-index candidates and entries whose field relevance cannot be determined.

### Tests

  - `cargo test -p paimon-datafusion --test procedures test_create_global_index`

### DIFF
--- a/crates/paimon/src/table/table_scan.rs
+++ b/crates/paimon/src/table/table_scan.rs
@@ -31,7 +31,7 @@ use super::stats_filter::{
     data_file_matches_predicates_with_key_stats, group_by_overlapping_row_id, FileStatsRows,
     ResolvedStatsSchema,
 };
-use super::Table;
+use super::{find_field_id_by_name, Table};
 use crate::io::FileIO;
 use crate::spec::{
     avro::SharedSchemaCache, bucket_dir_name, BinaryRow, BucketFunctionType, CoreOptions,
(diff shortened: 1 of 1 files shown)


crates/paimon/src/table/table_scan.rs
```diff
@@ -394,33 +394,56 @@ fn split_row_ranges_for_files(
...
     }
-    partition_filter
-        .map(|filter| filter.matches_entry(&entry.partition))
-        .transpose()
-        .map(|matched| matched.unwrap_or(true))
+    let Some(global_index) = entry.index_file.global_index_meta.as_ref() else {
+        // Keep malformed entries so the scanner can preserve fail-loud validation.
+        return Ok(true);
+    };
+    Ok(predicate_field_ids.contains(&global_index.index_field_id)
```
XiaoHongbo-Hope, line 442: Malformed index_meta for an unrelated sorted index is pruned before validation, so it no longer fails loudly as described in the PR.

QuakeWang: @XiaoHongbo-Hope Good catch. This is intentional: unrelated indexes are pruned before `index_meta` validation, matching Java. I've updated the PR description to clarify the fail-loud scope.

XiaoHongbo-Hope:
+1

XiaoHongbo-Hope (approved):

PR #704, opened 2026-08-11
perf(parquet): prune IN predicates with page stats

QuakeWang:
### Purpose

Parquet page pruning reused the shared statistics evaluator, but `ParquetPageStats` did not enable min/max pruning for `IN`. As a result, page indexes were loaded for `IN` predicates while non-matching pages were still retained.

### Brief change log

  - Enable min/max pruning for `IN` predicates using Parquet page statistics.
  - Add a multi-page regression test verifying that only the matching page is retained.

### Tests

  - `cargo check`
  - `cargo build`
  - `cargo test -p paimon-rest-server --all-targets`


JingsongLi (approved):
+1

PR #705, opened 2026-08-12
perf(arrow): prune IN predicates with row-group stats

QuakeWang:
### Purpose

`IN` min/max pruning is opt-in in the shared statistics evaluator. Parquet page statistics enabled it in #704, but Parquet and Mosaic row-group statistics still used the default fail-open behavior. As a result, non-matching row groups could not be skipped, including in Parquet files without page indexes.

### Brief change log

  - Enable min/max pruning for `IN` predicates using Parquet and Mosaic row-group statistics.
  - Preserve fail-open behavior for missing, inverted, or incompatible statistics.
  - Add regression tests for row-group selection and avoided Mosaic column reads.

### Tests


JingsongLi:
+1

JingsongLi (approved):

PR #707, opened 2026-08-13
feat(vindex): build indexes from streamed data splits

jerry-024:
### Purpose

The previous vindex builder collected every Arrow batch and then flattened all vectors into another contiguous buffer. For a 10M x 768 float32 shard, those two full-size copies alone could exceed 57 GiB before index training started.

This change builds each vindex shard from a single streamed `DataSplit`, keeping builder-owned memory bounded while preserving one index file per shard.

### Brief change log

- Stream Arrow record batches once, validating vector shape, nulls, sliced-array offsets, and contiguous row IDs batch by batch.
- Spill validated raw vectors to one anonymous local temporary file while incrementally feeding the trainer.
- Add Java-compatible `<index-type>.train.sample-ratio` and field-level overrides; the ratio affects training only and every row is still indexed.
- After training, read the spill file through a bounded aligned buffer and add vectors in chunks.
- Stream the completed index directly to FileIO without a full serialized `Vec<u8>` or a second local temporary file.
- Best-effort delete visible output objects on failures, including outputs from earlier shards through `TableCommit::abort`.

### Tests

- [x] `RUSTC_BOOTSTRAP=1 RUSTFLAGS='-Zcrate-attr=feature(stdarch_neon_f16)' cargo test -p paimon --lib vindex`
- [x] `RUSTC_BOOTSTRAP=1 RUSTFLAGS='-Zcrate-attr=feature(stdarch_neon_f16)' cargo check -p paimon --lib --features fulltext`
- [x] Covers option precedence and invalid training ratios.
- [x] Covers sliced List and FixedSizeList Arrow arrays.
- [x] Covers multiple source batches, incremental builds, one file per shard, and cleanup when a later shard fails.

### API and Format

Adds the optional vindex build setting `<index-type>.train.sample-ratio`, defaulting to `1.0` with valid range `(0, 1]`. There is no index storage-format change; each logical shard still produces one existing-format index file.

### Documentation

The SQL documentation now describes `train.sample-ratio`, its valid range, and that all rows remain indexed.

### Notes

IVF-Flat core still retains raw vector state and creates per-list serialization copies. Chunked IVF-Flat serialization is intentionally deferred to a separate `paimon-vindex-core` change.

The current FileIO async-writer wrapper cannot explicitly abort an unfinished multipart upload. Failed builds delete visible objects best-effort; object-store lifecycle rules should reclaim invisible abandoned multipart uploads.

### DIFF
--- a/crates/paimon/src/table/vindex_index_build_builder.rs
+++ b/crates/paimon/src/table/vindex_index_build_builder.rs
@@ -21,18 +21,22 @@ use crate::spec::{
 };
 use crate::table::source::exclude_row_ranges;
 use crate::table::{
-    CommitMessage, DataSplitBuilder, RowRange, SnapshotManager, Table, TableCommit,
+    CommitMessage, DataSplit, DataSplitBuilder, RowRange, SnapshotManager, Table, TableCommit,
 };
 use crate::vindex::{is_vindex_index_type, VindexVectorIndexOptions};
 use crate::{Error, Result};
 use arrow_array::{Array, FixedSizeListArray, Float32Array, Int64Array, ListArray, RecordBatch};
-use bytes::Bytes;
+use arrow_buffer::MutableBuffer;
 use futures::TryStreamExt;
 use paimon_vindex_core::index::{VectorIndexConfig, VectorIndexTrainer, VectorIndexWriter};
 use paimon_vindex_core::io::PosWriter;
 use std::collections::HashMap;
+use std::io::{Read, Seek, SeekFrom};
+use tokio::io::AsyncWriteExt;
+use tokio_util::io::SyncIoBridge;
 
 const INDEX_DIR: &str = "index";
+const VECTOR_BUFFER_BYTES: usize = 8 * 1024 * 1024;
 
 pub struct VindexIndexBuildBuilder<'a> {
     table: &'a Table,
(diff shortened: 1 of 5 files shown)


crates/paimon/src/table/vindex_index_build_builder.rs
```diff
@@ -245,13 +431,38 @@ impl<'a> VindexIndexBuildBuilder<'a> {
...
+                .table
+                .file_io()
+                .new_output(&index_path)?
+                .async_writer()
+                .await?;
+            let mut output = SyncIoBridge::new(async_writer);
+            tokio::task::spawn_blocking(move || -> std::io::Result<()> {
+                let mut writer = writer;
+                writer.write(&mut PosWriter::new(&mut output))?;
+                output.shutdown()
```
shyjsarah, line 445: **Suggestion:** Since a running `spawn_blocking` task cannot be aborted, cancelling the outer build future may still allow this worker to finish `output.shutdown()` after the surrounding cleanup path has been dropped. The resulting index file would not be referenced by any commit message or manifest.

If orphan index files are already covered by a separate cleanup mechanism, could we document or test that assumption here? Otherwise, it may be worth keeping a provisional output guard or another cleanup mechanism until the file is transferred into a `CommitMessage`.

QuakeWang (approved):
LGTM

crates/paimon/src/table/vindex_index_build_builder.rs
jerry-024: Good catch. I think this should be handled by orphan-file cleanup rather than adding cancellation-specific ownership logic to the index builder. A guard here would still not cover equivalent leaks caused by process/runtime shutdown, while orphan cleanup can consistently remove any unreferenced index files after the retention window. I will address the gap in the orphan cleanup path instead.

PR #708, opened 2026-08-14
fix(vindex): avoid full index reads during refine

jerry-024:
### Purpose

During vector refine and raw reranking, resolve_raw_vector_metric read the entire Vindex file only to obtain the metric. For a 1.95 GB index, this still reconstructed the full file from local-cache blocks and copied it again, making refine slow even on a cache hit.

### Brief change log

- Resolve the Vindex metric directly from global-index manifest metadata when the metric is present.
- Keep compatibility with Java and legacy indexes whose metadata is empty by range-reading at most the 256-byte Vindex header.
- Pass the returned Bytes directly to the Vindex reader, removing the additional to_vec copy.
- Preserve the underlying error source for header reader failures.

### Tests

- Added coverage proving manifest metadata resolves the metric without an index file read.
- Added coverage for the Java and legacy empty-metadata header fallback.
- cargo +1.97.0 test -p paimon test_resolve_raw_vector_metric -- --nocapture
- cargo +1.97.0 clippy -p paimon --lib -- -D warnings
- cargo +1.97.0 fmt --all -- --check

### API and Format

No API or storage-format changes. Existing Java and legacy Vindex files remain compatible through the bounded header fallback.

### Documentation

### DIFF
--- a/crates/paimon/src/table/vector_search_builder.rs
+++ b/crates/paimon/src/table/vector_search_builder.rs
@@ -17,7 +17,7 @@
 
 use crate::arrow::format::FilePredicates;
 use crate::arrow::residual::{evaluate_predicates_mask, widen_scan_fields};
-use crate::io::FileIO;
+use crate::io::{FileIO, FileRead};
 use crate::lumina::reader::LuminaVectorGlobalIndexReader;
 use crate::lumina::{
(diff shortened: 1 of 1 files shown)


QuakeWang (approved):
LGTM

PR #711, opened 2026-08-14
feat: add vector search timing and I/O diagnostics

jerry-024:
## Summary

Add opt-in timing and I/O diagnostics for locating latency, concurrency waits, and read amplification in vector search. Logs are emitted through the `log` crate only when `PAIMON_LOG_VECTOR_SEARCH_TIMING=1` and the `paimon::vector_search` target is enabled at debug level.

## Changes

- Report API and evaluation timings for snapshot and manifest loading, global-index search, refinement, raw fallback, result merging, and finalization.
- Report vindex reader lifecycle timings, native search wall time, batch chunking, scalar fallbacks, concurrency, and memory budget for both batch and single-query entry points.
- Track logical ranges, requested bytes, physical reads, returned bytes, read-ahead hits, caller I/O wait, and internal range-read permit wait across cloned vindex readers.
- Report full-file read time and byte counts for the Lumina backend and the no-runtime vindex fallback.
- Stabilize the existing disk-cache CRC restart test by constructing its corrupted fixture directly (`e6f3b35`; unrelated test-only cleanup).

## Testing

- [x] `cargo +1.97.0 fmt --all -- --check`
- [x] `cargo +1.97.0 clippy -p paimon --lib -- -D warnings`
- [x] `cargo +1.97.0 test -p paimon --lib` (2302 passed, 1 ignored)
- [x] Range-reader timing/concurrency tests and scalar vindex search regression test
- [x] `git diff --check`

## Notes

- No public API or storage-format changes.
- Timing fields that aggregate concurrent work use `*_sum_ms` names and may exceed wall-clock totals.

### DIFF
--- a/crates/paimon/src/vindex/reader.rs
+++ b/crates/paimon/src/vindex/reader.rs
@@ -48,6 +48,8 @@ struct VindexBatchStats {
     batch_index_parallelism: usize,
 }
 
+type VindexBatchSearchResult = (Vec<Option<HashMap<u64, f32>>>, Option<VindexBatchStats>);
+
 trait ErasedSeekRead: Send {
     fn pread_erased(&mut self, ranges: &mut [ReadRequest<'_>]) -> io::Result<()>;
 
@@ -433,7 +435,7 @@ fn search_batch_vindex(
     vector_searches: &[VectorSearch],
     shard_concurrency: usize,
     timing_enabled: bool,
-) -> crate::Result<(Vec<Option<HashMap<u64, f32>>>, Option<VindexBatchStats>)> {
+) -> crate::Result<VindexBatchSearchResult> {
     let mut results: Vec<Option<HashMap<u64, f32>>> =
         (0..vector_searches.len()).map(|_| None).collect();
     let mut groups: Vec<(PreparedSearch, Vec<usize>)> = Vec::new();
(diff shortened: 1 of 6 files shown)


crates/paimon/src/vindex/reader.rs
```diff
@@ -225,13 +296,16 @@ impl VindexVectorGlobalIndexReader {
...
+        let (results, batch_stats) = search_batch_vindex(
             reader,
             metadata,
             &self.options,
             vector_searches,
             self.batch_index_parallelism,
-        )
+            self.timing_enabled,
+        )?;
+        self.batch_stats = batch_stats;
```
shyjsarah, line 310: **Minor:** `search_batch()` stores the collected stats on the reader, but only `visit_batch_vector_search()` consumes them and emits `paimon_vindex_reader`. The PK-vector scorer calls `load_validated()` + `search_batch()` directly (`vector_search_builder.rs:923-935`), so with timing enabled that route collects batch/range-I/O diagnostics but returns without publishing them.

Could we make telemetry reporting independent of the convenience wrapper (for example, return/commit the stats from the search API or pass an explicit diagnostics reporter), and add coverage for the PK/hybrid route?

shyjsarah:
+1

shyjsarah (approved):
+1

LsomeYeah (approved):
+1

PR #714, opened 2026-08-14
fix(dlf): refresh expiring credentials

XiaoHongbo-Hope:
## What changed

- Parse the standard ECS `Expiration` field into milliseconds so DLF catalog credentials are refreshed before expiry.
- Keep a refresh-capable REST token FileIO attached to loaded tables instead of freezing the first data token.
- Coalesce concurrent refreshes, rebuild the storage operator with rotated credentials, and preserve the catalog local cache.
- Document ECS RAM-role authentication and REST data-token configuration.

Static AK/SK credentials remain unchanged and are not rotated.

## Validation

- `cargo test -p paimon --lib` (2254 passed, 1 ignored)
- `cargo test -p paimon token`
- `cargo test -p paimon io::file_io::` (31 passed)
- All non-Python workspace targets compile, including DataFusion, C bindings, REST server, and integration tests.


JingsongLi (approved):
+1

PR #715, opened 2026-08-14
fix(datafusion): normalize index_type in global index procedures

jackylee-ch:
`sys.create_global_index` matched `index_type` case-insensitively for btree and
bitmap but exactly for the vindex types, so `index_type => 'BTREE'` worked while
`index_type => 'IVF-FLAT'` was rejected as unsupported. `sys.drop_global_index`
already canonicalizes every type. None of the three procedures trimmed the
argument, so `' btree '` was reported as unsupported too.

The three procedures now canonicalize `index_type` at their boundary through a
shared helper (trim, then lowercase), mirroring
`indexType.toLowerCase(Locale.ROOT).trim()` in Java's Flink and Spark
`Create/DropGlobalIndexProcedure`. The core builders keep exact matching: they
are the analog of Java's `GlobalIndexer` and persist `index_type` into index
metadata, which the read path matches exactly. An unsupported type is echoed
back as written so a typo stays recognizable.

`create_lumina_index` shares the helper but has no end-to-end test here because
it needs `LUMINA_LIB_PATH`, which CI does not provide.


JingsongLi (approved):
+1

PR #716, opened 2026-08-15
fix(python): Release GIL during catalog I/O

XiaoHongbo-Hope:
### Purpose

`PaimonCatalog` performed blocking catalog and REST I/O while holding the Python GIL. This stalls other Python threads and makes an in-process Python REST server time out after 30 seconds.

### Changes

- Release the GIL while creating a catalog.
- Release the GIL while listing databases/tables and loading a table.
- Add an in-process REST server regression test covering all four calls.

### Verification

- `test_catalog_gil.py`: 1 passed in 0.56s.
- PyPaimon mock REST native planning: 1 split in 0.041s; the previous binding timed out after 30s.
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/bindings/python/src/context.rs
+++ b/bindings/python/src/context.rs
@@ -119,8 +119,8 @@ pub struct PaimonCatalog {
 impl PaimonCatalog {
     /// Create a Paimon catalog that can be registered into a DataFusion session.
     #[new]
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #717, opened 2026-08-16
fix(scan): preserve Data Evolution file order in row-id groups

XiaoHongbo-Hope:
### Purpose

Native-plan CI exposed inconsistent Data Evolution `split.files` ordering across implementations.

Java preserves input manifest order within each overlapping row-id group ([apache/paimon#6496](https://github.com/apache/paimon/pull/6496)). This PR aligns Rust with that behavior by retaining each file's original position while grouping overlapping ranges.

Groups remain ordered by row-id range; files within a group preserve manifest order. Files without `first_row_id` remain standalone groups.

### Tests

- overlapping, chained, and non-overlapping ranges;
- input order within a group;
- files without `first_row_id`.

### DIFF
--- a/crates/paimon/src/table/stats_filter.rs
+++ b/crates/paimon/src/table/stats_filter.rs
@@ -571,29 +571,39 @@ fn data_evolution_predicate_must_match(
 
 /// Groups data files by overlapping `row_id_range` for data evolution.
 ///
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #718, opened 2026-08-16
fix(data-evolution): require row tracking and row IDs

XiaoHongbo-Hope:
### Purpose

Reject invalid Data Evolution tables and files before they can produce unreadable data.

Java requires data-evolution.enabled to be used with row-tracking.enabled. Rust previously allowed CREATE or schema changes without row tracking, so writes succeeded without first_row_id and later scans failed. This PR adds the same schema validation for creation and SchemaChange.

Scan planning also validates first_row_id before row-range grouping. This keeps legacy invalid tables fail-closed and aligns planning with Java DataFileMeta.nonNullRowIdRange and PyPaimon non_null_row_id_range.

### Tests

- cargo test -p paimon -q
- cargo test -p paimon-datafusion --test variant_pushdown data_evolution_row_id_survives_variant_extraction_pushdown -q
- cargo clippy -p paimon --tests -- -D warnings
- cargo fmt --all -- --check

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -6421,7 +6421,7 @@ mod tests {
         let sql_context = make_sql_context(catalog.clone()).await;
 
         sql_context
-            .sql("CREATE TABLE mydb.t1 (id INT, payload BLOB NOT NULL) WITH ('data-evolution.enabled' = 'true')")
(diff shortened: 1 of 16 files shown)


JingsongLi (changes requested):
Row tracking validation is incomplete. It fails to enforce three Java-side constraints: Primary Key is prohibited, `bucket` must be `-1`, and `clustering.incremental` is disallowed for Data Evolution. However, empirical testing confirms that a schema can be successfully created with the configuration `PK + bucket=1 + row tracking + data evolution`.

JingsongLi (approved):
+1

PR #719, opened 2026-08-16
fix(file_index): align file index format with Java V1

QuakeWang:
### Purpose

Linked issue: #34

The existing Rust FileIndex implementation used a Rust-only wire format that differed from Java Paimon V1 in byte order, version, string encoding, and integer widths. This prevented Java and Rust from exchanging FileIndex files reliably. The reader also assumed a fixed-size header and did not fully validate encoded ranges.

### Brief change log

  - Align FileIndex reading and writing with Java V1.
  - Preserve absent and empty index payloads separately.
  - Validate truncated headers, invalid offsets, lengths, and overflows.
  - Support headers larger than 1 MiB.
  - Add Java golden fixtures and boundary tests.

### Tests

  - Rust 1.91 workspace MSRV check
  - Java/Rust V1 interoperability verification

### API and Format

FileIndex payload APIs now use `Option<Bytes>` to distinguish absent indexes from empty payloads. The wire format is changed to Java-compatible V1. The legacy Rust-only format is intentionally unsupported.

### Documentation

Updated inline format documentation.

### DIFF
--- a/crates/paimon/src/file_index/file_index_format.rs
+++ b/crates/paimon/src/file_index/file_index_format.rs
@@ -17,33 +17,159 @@
 
 use std::collections::HashMap;
 
-use bytes::{Buf, BufMut, Bytes, BytesMut};
+use bytes::{BufMut, Bytes, BytesMut};
 
 use crate::{
     io::{FileIO, FileRead, FileStatus, InputFile, OutputFile},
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #723, opened 2026-08-17
perf(vindex): split build timing logs by phase

jerry-024:
### Purpose

Vector index build diagnostics currently do not distinguish storage I/O, Parquet decoding, index training, and index population costs. This change adds opt-in phase-level timing logs so build bottlenecks can be identified without adding log noise when diagnostics are not enabled.

Enable the logs with:

`PAIMON_LOG_VECTOR_INDEX_BUILD_TIMING=1`

### Brief change log

- Split source batch wait, OSS/file reads, and Parquet decoding into independent timings.
- Split raw temporary-file writes/rereads, index training, and index add into independent timings.
- Report serialization/upload, snapshot commit, total, unattributed time, and build counters in one structured log event.
- Keep timing collection and logging disabled by default.

### Tests

- [x] `cargo test -p paimon`
- [x] Run vector index build tests with timing disabled and enabled; verify logs are emitted only when enabled and phase fields are populated.
- [x] `git diff --check`

### API and Format

### Documentation

No documentation changes. The environment variable is intended for opt-in build diagnostics.

### DIFF
--- a/crates/paimon/src/table/data_file_reader.rs
+++ b/crates/paimon/src/table/data_file_reader.rs
@@ -20,7 +20,7 @@ use crate::arrow::format::create_format_reader_with_budget;
 use crate::arrow::schema_evolution::{create_index_mapping, NULL_FIELD_INDEX};
 use crate::arrow::ParquetReadBudget;
 use crate::deletion_vector::{DeletionVector, DeletionVectorFactory};
-use crate::io::FileIO;
+use crate::io::{FileIO, FileRead};
 use crate::spec::{
     is_variant_extraction_row_type, DataField, DataFileMeta, DataType, Predicate, ROW_ID_FIELD_NAME,
 };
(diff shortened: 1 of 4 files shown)


crates/paimon/src/table/data_file_reader.rs
```diff
@@ -33,7 +33,51 @@ use arrow_cast::cast;
...
+    inner: Box<dyn FileRead>,
+    timing: Arc<DataFileReadTiming>,
+}
+
+#[async_trait::async_trait]
+impl FileRead for TimedFileRead {
+    async fn read(&self, range: Range<u64>) -> crate::Result<bytes::Bytes> {
+        let start = Instant::now();
+        let result = self.inner.read(range).await;
+        self.timing.add_file_read(start.elapsed());
```
shyjsarah, line 77: `file_read_nanos` sums the elapsed time of every `FileRead::read`, but Parquet range fetching can run multiple reads concurrently (up to `RANGE_FETCH_CONCURRENCY`). Overlapping waits are therefore double-counted, so the emitted `oss_read_ms` is cumulative request wait rather than storage-I/O wall time and may exceed the enclosing phase/total. Could we either measure the wall-clock I/O span, or rename this to something like `oss_read_wait_sum_ms` and report the actual concurrency? The current hard-coded `data_file_read_concurrency=1` is misleading for the same reason.

crates/paimon/src/table/vindex_index_build_builder.rs
```diff
@@ -191,13 +267,23 @@ impl<'a> VindexIndexBuildBuilder<'a> {
...
         }
 
+        let commit_start = vector_index_build_timing_enabled().then(Instant::now);
         commit
             .commit_if_latest_snapshot(messages, snapshot.id())
             .await?;
+        if let Some(commit_start) = commit_start {
+            let commit = commit_start.elapsed();
+            for timing in timings {
+                timing.log(&self.index_type, commit);
```
shyjsarah, line 284: This applies the single operation-level commit duration to every shard-level timing event. For a multi-shard build, summing events counts commit N times; an early shard's `total_ms` also excludes time spent building later shards, so it is neither shard wall time nor operation wall time. Suggest keeping shard events limited to shard phases and emitting one operation-level event for all-shard build time, the single commit, and end-to-end total (ideally linked by a build ID).

crates/paimon/src/table/vindex_index_build_builder.rs
```diff
@@ -172,8 +247,9 @@ impl<'a> VindexIndexBuildBuilder<'a> {
         );
         let shard_count = shards.len();
         let mut messages = Vec::with_capacity(shard_count);
+        let mut timings = Vec::with_capacity(shard_count);
```
shyjsarah, line 250: This reserves one `VectorIndexBuildTiming` slot per shard even when `PAIMON_LOG_VECTOR_INDEX_BUILD_TIMING` is disabled, leaving the default-off path with O(shard_count) diagnostic memory. Could we read the enable flag once and allocate/push only when enabled (or aggregate bounded operation-level stats instead)?

crates/paimon/src/table/vindex_index_build_builder.rs
```diff
@@ -28,15 +29,89 @@ use crate::{Error, Result};
...
+        let accounted = self
+            .source_batch_wait
+            .saturating_add(self.raw_temp_write)
+            .saturating_add(self.train_finish)
+            .saturating_add(self.raw_temp_reread)
+            .saturating_add(self.index_add)
+            .saturating_add(self.serialize_upload)
+            .saturating_add(commit);
+        let unattributed = total.saturating_sub(accounted);
+        eprintln!(
```
shyjsarah, line 85: Could this use the logging facade instead of `eprintln!`? Direct stderr bypasses the embedding application's filtering, targets, sinks, structured handlers, and test logger, and performs synchronous output in this async path. Nearby vector diagnostics already use `log::debug!`; a stable target such as `paimon::vector_index_build` would keep this routable while the env flag can still avoid timing overhead when disabled.

shyjsarah (commented):
Reviewed the phase-level timing changes. The build path itself looks sound and CI is green, but I found four issues in the diagnostic semantics/implementation that are worth addressing so the emitted data remains actionable. Details are inline.

JingsongLi (approved):
+1

PR #725, opened 2026-08-18
fix(lumina): drop unfilled search slots and rank NaN scores last

jackylee-ch:
The label buffer is allocated with `vec![0u64; k]`, while `collect_results` treats
only `u64::MAX` — the C ABI's `-1` — as "no result". Row id `0` is legal, so a
slot the searcher leaves untouched is indistinguishable from a real hit, and it
carries distance `0.0`, which `convert_distance_to_score` maps to `1.0` — for L2
(`1/(1+d)`, `d >= 0`) and cosine the highest score the metric can produce. A
search returning fewer neighbours than requested therefore surfaces row 0 as its
best match. The FFI returns a status code and never a count, so Rust cannot
detect a short return; allocating the sentinel is correct either way.

Reverting that allocation makes the new test report
`{22: 0.125, 11: 0.25, 0: 1.0}` — row 0 above both real hits.

**Both review comments were right; this supersedes the original fix.**

JingsongLi: `f32::total_cmp` ranks a positive NaN above every finite score, so
`top_k = 1` over `[(8, 0.5), (7, NaN)]` returned the NaN row. `compare_scores`
ranks both NaN signs below every real score — the score-domain mirror of
`java_float_compare`. The test asserts the finite candidate wins in either
arrival order.

QuakeWang: the eviction branch is unreachable from production, and more so than
stated — both call sites size the buffers to the same `k` passed as `top_k`, so
nothing is evicted and the ordering cannot affect the returned map. It cannot be
made reachable: truncation happens inside the native searcher, and over-fetching
would be a design change, not a fix. The comparator is kept as a property of the
function, whose top-k contract `test_collect_results` already pins, not as a
behaviour fix. The production-visible change is the allocation.

`vindex/reader.rs`'s `collect_results` has the same shape and the same
unreachability; left alone.


crates/paimon/src/lumina/reader.rs
```diff
@@ -75,13 +91,15 @@ fn collect_results(
...
+        let entry = ScoredRow { row_id, score };
         if min_heap.len() < top_k {
-            min_heap.push(ScoredRow { row_id, score });
-        } else if let Some(peek) = min_heap.peek() {
-            if score > peek.score {
-                min_heap.pop();
-                min_heap.push(ScoredRow { row_id, score });
-            }
+            min_heap.push(entry);
+        } else if min_heap
```
QuakeWang, line 127: Both production callers pass exactly `effective_k` labels and the same value as `top_k`, so `labels.len() <= top_k` and this replacement branch cannot execute. The new tests trigger it only by passing more candidates than production can return. If ties are discarded at the native top-k cutoff, those rows never reach this comparator. Please enforce the tie-break before that cutoff (or return more than `k` candidates) and test through the real call contract.

crates/paimon/src/lumina/reader.rs
```diff
@@ -64,8 +64,24 @@ fn collect_results(
...
+                .score
+                .total_cmp(&self.score)
+                .then_with(|| self.row_id.cmp(&other.row_id))
+        }
+    }
+
+    impl ScoredRow {
+        fn is_stronger_than(&self, other: &Self) -> bool {
+            self.score
+                .total_cmp(&other.score)
```
JingsongLi: [P2] Do not promote NaN scores above finite candidates

`f32::total_cmp` ranks a positive NaN above every finite score, so this changes more than the tie-break. With `top_k = 1` and native results `[(row 8, 0.5), (row 7, NaN)]`, the current code ignores the later NaN because `NaN > 0.5` is false, while this comparator evicts row 8 and returns the NaN-scored row. A non-finite stored vector can produce such a distance; the PK-vector metric code uses `java_float_compare` specifically to keep all NaNs worst. Please skip/reject non-finite scores or use NaN-worst comparison semantics here, and change the regression test so a finite candidate wins regardless of arrival order.

JingsongLi (approved):
+1

PR #726, opened 2026-08-18
feat(vindex): add DiskANN and IVF-SQ/RQ support

jerry-024:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/crates/paimon/src/vindex/mod.rs
+++ b/crates/paimon/src/vindex/mod.rs
@@ -39,7 +39,6 @@ const DEFAULT_METRIC: &str = "inner_product";
 const DEFAULT_NLIST: &str = "256";
 const DEFAULT_PQ_M: &str = "16";
 const DEFAULT_PQ_USE_OPQ: &str = "false";
-const DEFAULT_RQ_BITS: &str = "4";
 const DEFAULT_TRAIN_SAMPLE_RATIO: f64 = 1.0;
 const VECTOR_SEARCH_TIMING_ENV: &str = "PAIMON_LOG_VECTOR_SEARCH_TIMING";
 const DISKANN_OPTION_KEYS: &[(&str, &str)] = &[
@@ -193,18 +192,18 @@ impl VindexVectorIndexOptions {
             );
         }
         if index_type == IVF_RQ_IDENTIFIER {
-            native_options.insert(
-                "rq.bits".to_string(),
-                option_value(
+            for key in ["rq.bits", "max-bytes-per-vector"] {
+                if let Some(value) = optional_value(
                     table_options,
                     user_options,
                     field.name(),
                     index_type,
-                    "rq.bits",
-                    "rq.bits",
-                    DEFAULT_RQ_BITS,
-                ),
-            );
+                    key,
+                    key,
+                ) {
+                    native_options.insert(key.to_string(), value);
+                }
+            }
         }
         if index_type == DISKANN_IDENTIFIER {
             for &(native_key, paimon_suffix) in DISKANN_OPTION_KEYS {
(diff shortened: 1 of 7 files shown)


crates/paimon/src/vindex/reader.rs
```diff
@@ -441,7 +447,39 @@ fn prepare_search(
     if top_k == 0 {
         return Ok(None);
     }
-    let nprobe = int_parameter(options, NPROBE_PARAMETER, DEFAULT_NPROBE)?;
+    let mut params = match metadata.index_type {
+        IndexType::DiskAnn => {
+            if options.contains_key(NPROBE_PARAMETER) {
```
JingsongLi: [P2] Ignore tuning options for the other backend

Different global-index types are explicitly allowed to coexist on the same field, and the search path fans the same merged option map out to every matching entry. If IVF and DiskANN indexes coexist, a legitimate `ivf.nprobe` reaches the DiskANN reader here and aborts the entire query; the inverse happens when an IVF reader sees the DiskANN knob. Please consume only the option applicable to `metadata.index_type` and ignore the recognized option for the other backend, or partition the options before fan-out. A mixed IVF + DiskANN regression test would cover this.

crates/paimon/src/vindex/reader.rs
```diff
@@ -30,6 +31,8 @@ use std::time::{Duration, Instant};
 
 const DEFAULT_NPROBE: usize = 16;
 const NPROBE_PARAMETER: &str = "ivf.nprobe";
+const L_SEARCH_PARAMETER: &str = "diskann.l-search";
```
JingsongLi: [P2] Use the canonical DiskANN search option name

Apache Paimon Java and the public option contract use `diskann.l_search`, with an underscore. With that canonical key, this lookup misses and silently falls back to automatic search width, so the requested recall/latency tuning is ignored. Please use `diskann.l_search` as the canonical name; accepting `diskann.l-search` as a compatibility alias is fine.

crates/paimon/src/vindex/mod.rs
```diff
@@ -159,6 +192,34 @@ impl VindexVectorIndexOptions {
                 ),
             );
         }
+        if index_type == IVF_RQ_IDENTIFIER {
+            native_options.insert(
+                "rq.bits".to_string(),
```
JingsongLi: [P2] Let the IVF-RQ capacity budget determine `rq.bits`

This unconditional default prevents `paimon-vindex-core` from inferring the residual width from `max-bytes-per-vector`. The new validation and forwarding logic accepts that capacity option only for DiskANN, so `ivf-rq.max-bytes-per-vector` is rejected in procedure options and silently ignored as a table property, even though Java 0.3 forwards it and core 0.3 uses it to enforce the storage bound. Please forward the applicable capacity options for IVF-RQ and insert `rq.bits` only when it is explicitly configured.

crates/paimon/src/vindex/mod.rs
```diff
@@ -249,17 +310,33 @@ fn is_supported_user_option_key(key: &str, index_type: &str, field_name: &str) -
...
     }
 }
 
 fn is_allowed_paimon_suffix(suffix: &str, index_type: &str) -> bool {
     match suffix {
-        "dimension" | "nlist" | "distance.metric" | "train.sample-ratio" => true,
-        "pq.m" | "pq.use-opq" => index_type == IVF_PQ_IDENTIFIER,
-        _ => false,
+        "dimension" | "distance.metric" => true,
+        "nlist" | "train.sample-ratio" => index_type != DISKANN_IDENTIFIER,
```
JingsongLi: [P2] Accept DiskANN training sampling in procedure options

The Rust documentation says `train.sample-ratio` applies to all vindex types, and the resolver plus build path already consume it without excluding DiskANN. Grouping it with `nlist` here makes `options => diskann.train.sample-ratio=0.5` fail as unknown, while the equivalent table property is honored. Please allow this suffix for DiskANN and add a procedure-option regression test.

JingsongLi (approved):
+1

PR #727, opened 2026-08-18
fix(btree): read LZ4-compressed global index blocks written by Java

jackylee-ch:
A BTree global index written by Java with LZ4 block compression cannot be read
at all: every query fails with `Compression type Lz4 not supported`, even though
the file holds the same data as the zstd and uncompressed variants. The repo
already carries the Java-written fixture for it, used only to assert that the
read fails.

Paimon's LZ4 block is not the LZ4 frame format and not a bare block either. It
is the block writer's var-int uncompressed length, then `Lz4BlockCompressor`'s
8-byte header of two little-endian `i32`s (compressed and uncompressed length),
then the raw LZ4 block. Reading the fixture's bytes confirms the layout: the
var-int says 261 and the header repeats 261 alongside a compressed length of
203.

**Fix**: decode that frame in `decompress_block`. The uncompressed length is
stored twice, so the two copies are cross-checked and a mismatch is rejected as
corrupt, as are negative lengths and a payload shorter than the header claims.
`lz4_flex` is already a dependency.

Scope is the read path only. The writer still falls back to no compression for
LZ4, unchanged — the comment is reworded to say so. A round-trip test would have
passed either way, so the two new tests read the Java fixture directly and assert
all three codecs decode to the same rows.

### DIFF
--- a/crates/paimon/src/btree/sst_file.rs
+++ b/crates/paimon/src/btree/sst_file.rs
@@ -155,7 +155,7 @@ impl SstFileWriter {
                 }
             }
             _ => {
-                // LZ4/LZO not implemented yet, fall back to no compression
+                // LZ4/LZO writing is not implemented yet, fall back to no compression
                 (Cow::Borrowed(block), BlockCompressionType::None)
             }
         }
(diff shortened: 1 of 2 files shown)


JingsongLi:
This has been overtaken by current `main`: #731 now contains the same Java-compatible LZ4 block decoder (including the outer varint/header cross-check) and already turns `test_java_compat_int_lz4` into a passing compatibility test. I also verified locally that merging this head into current `main` conflicts in both `crates/paimon/src/btree/sst_file.rs` and `crates/paimon/src/btree/tests.rs`.

Please rebase and drop the duplicated decoder/test changes (or close this PR if no independent change remains). The additional cross-codec equivalence assertion may still be worth preserving as a small follow-up, but the core fix is already present on `main`.

PR #728, opened 2026-08-18
fix(datafusion): surface tag create-time and retention in $tags

jackylee-ch:
A tag written by Flink or Spark carries `tagCreateTime` and `tagTimeRetained`
next to the snapshot fields, but `$tags` reported both columns as `NULL` for
every row. `TagManager` deserialized the tag file straight into `Snapshot`, which
has no such fields, and nothing rejects unknown keys — so the values on disk were
dropped silently and the table hard-coded two null arrays.

Jackson's shape is easy to get wrong. `JsonSerdeUtil` registers `JavaTimeModule`
without disabling `WRITE_DATES_AS_TIMESTAMPS`, so `tagCreateTime` is the array
`[year, month, day, hour, minute, second, nanoOfSecond]` with trailing zero
components omitted — five items is legal — and `tagTimeRetained` is a decimal
seconds number, not an ISO string. `pypaimon`'s `time_utils` encodes exactly this
and states it matches `LocalDateTimeSerializer` byte for byte.

**Fix**: `TagManager::get_with_metadata` / `list_all_with_metadata` parse the two
fields alongside the snapshot, padding a short array back to seven components.
Parsing is deliberately lenient: an unexpected shape leaves that one column
`NULL` rather than failing the query, so a hand-edited tag file cannot break a
listing. `$tags` renders the retention like Java's `Duration.toString()`.

Read side only. Making `sys.create_tag` write these fields belongs with the
procedure, which #489 is currently changing.

### DIFF
--- a/crates/integrations/datafusion/src/system_tables/tags.rs
+++ b/crates/integrations/datafusion/src/system_tables/tags.rs
@@ -20,9 +20,7 @@
 use std::sync::{Arc, OnceLock};
 
 use async_trait::async_trait;
-use datafusion::arrow::array::{
-    new_null_array, Int64Array, RecordBatch, StringArray, TimestampMillisecondArray,
-};
+use datafusion::arrow::array::{Int64Array, RecordBatch, StringArray, TimestampMillisecondArray};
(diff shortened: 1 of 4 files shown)


JingsongLi (approved):
+1

PR #730, opened 2026-08-19
fix: serialize unbounded varchar as string

JingsongLi:
## Summary

- serialize unbounded `VarCharType` with Java's `STRING` alias
- preserve `STRING NOT NULL` and bounded `VARCHAR(n)` behavior
- add regression coverage for all three serialization forms

## Root cause

`VarCharType` uses display-based Serde serialization, but its `Display` implementation always emitted `VARCHAR(length)`. As a result, the maximum length used to represent `STRING` was persisted as `VARCHAR(2147483647)` instead of Java's canonical `STRING` form.

## Validation

- `cargo test -p paimon --lib` (`2317 passed`, `2 ignored`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2838,9 +2838,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.15"
+version = "0.4.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


leaves12138 (approved):
Verified that unbounded VarChar serialization now matches Java Paimon's canonical STRING / STRING NOT NULL forms while bounded VARCHAR(n) remains unchanged. The parser already accepts both aliases, and the regression coverage covers nullable, non-null, and bounded cases. LGTM.

PR #731, opened 2026-08-19
[core] Support multivalue global index

JingsongLi:
## Context

PR #732 has merged. This branch is rebased directly onto main and contains only the multivalue global-index changes.

The implementation follows the latest Java behavior from apache/paimon #9305: index files store only SortedIndexFileMeta. Schema evolution is handled uniformly by IndexManifestEntry rather than by index-specific element-type metadata.

## What changed

- add the multivalue global-index type for array columns
- build Java-compatible bitmap postings for every non-null array element
- evaluate ARRAY_CONTAINS, ARRAYS_OVERLAP, and ARRAY_CONTAINS_ALL through the global-index scanner
- expose create and drop procedure support and document SQL usage

## Behavior

Null arrays, empty arrays, and null elements do not create postings. Duplicate elements in one row are deduplicated by the bitmap. Empty ARRAY_CONTAINS_ALL falls back to residual evaluation because the index cannot distinguish null arrays from empty arrays.

## Validation

- cargo test -p paimon --lib multivalue --no-fail-fast
- cargo test -p paimon-datafusion --test procedures test_create_global_index_rejects_unsupported_index_types -- --exact
- cargo clippy -p paimon --all-targets -- -D warnings
- cargo fmt --all -- --check

All checks pass locally.


leaves12138 (approved):
Looks good to me. I verified the multivalue index build and query semantics, Java-compatible bitmap encoding, source-row coverage, and the focused regression tests.

crates/paimon/src/table/global_index_scanner.rs
```diff
@@ -536,6 +565,27 @@ impl GlobalIndexScanner {
...
+                let key_type = if is_multivalue_predicate(*op) {
+                    let DataType::Array(array) = data_type else {
+                        return Err(Error::DataInvalid {
+                            message: format!(
+                                "Array global-index predicate {op} requires an ARRAY field type"
+                            ),
+                            source: None,
+                        });
+                    };
+                    array.element_type()
```
QuakeWang, line 580: This uses the current array element type for all index files, including files built before a schema change. For example, after changing `ARRAY<INT>` to `ARRAY<BIGINT>`, an old 4-byte key is passed to the BIGINT comparator and causes a panic when it reads 8 bytes. Could we keep the compatibility check here, or filter and rebuild incompatible index entries? A schema-evolution test would also help cover this case.

JingsongLi: This is not a specific issue but a general one that affects all indexes; we need to consider it separately later.

QuakeWang (approved):
LGTM

PR #732, opened 2026-08-19
feat: add Java-compatible array predicate pushdown

JingsongLi:
## What changed

- add `ARRAY_CONTAINS`, `ARRAYS_OVERLAP`, and `ARRAY_CONTAINS_ALL` to the predicate model and REST deserialization
- evaluate array predicates exactly over Arrow list arrays, including null/empty arrays, duplicate literals, decimals, NaNs, and signed zero
- push DataFusion `array_has`, `array_has_any` / `arrays_overlap`, and `array_has_all`, retaining DataFusion residuals where semantics differ
- prune all-null array files through predicate stats
- extract `TIME` values from Arrow and decode Java-compatible LZ4/LZO bitmap blocks

## Why

These are reusable predicate and read-compatibility capabilities independent of the MultiValue global index implementation in #731. Moving them into a prerequisite PR keeps #731 focused on the writer, metadata, scanner, and procedure wiring.

The exact array residual follows Java Paimon semantics: all NaN payloads compare equal, signed zero remains distinct, decimal values compare across scales, NULL arrays do not match, and an empty `ARRAY_CONTAINS_ALL` literal matches only non-null arrays.

## Validation

- `cargo test -p paimon --all-targets --features fulltext,vortex --no-fail-fast`
- `cargo test -p paimon-datafusion --lib filter_pushdown::tests --no-fail-fast`
- `cargo clippy --workspace --all-targets --features fulltext,vortex -- -D warnings`

All checks pass locally.

### DIFF
--- a/crates/integrations/datafusion/src/filter_pushdown.rs
+++ b/crates/integrations/datafusion/src/filter_pushdown.rs
@@ -16,6 +16,7 @@
 // under the License.
 
 use datafusion::arrow::array::Array;
+use datafusion::arrow::datatypes::DataType as ArrowDataType;
 use datafusion::common::{Column, ScalarValue};
 use datafusion::logical_expr::expr::{InList, ScalarFunction};
 use datafusion::logical_expr::{
@@ -169,6 +170,13 @@ impl<'a> FilterTranslator<'a> {
             // must keep its residual filter for NULL / three-valued semantics.
             Expr::Not(inner) => {
                 let inner = self.translate(inner.as_ref())?;
+                // A positive inexact predicate is only guaranteed to be a
+                // conservative superset. Negating it would turn that into a
+                // subset and could remove rows before DataFusion's residual
+                // runs (notably for floating-array NaN payload semantics).
+                if inner.requires_residual {
+                    return None;
+                }
                 Some(TranslatedPredicate {
                     predicate: Predicate::negate(inner.predicate),
                     requires_residual: true,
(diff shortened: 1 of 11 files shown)


crates/integrations/datafusion/src/filter_pushdown.rs
```diff
@@ -334,6 +347,55 @@ impl<'a> FilterTranslator<'a> {
...
+                    .ok()?
+            }
+            _ => return None,
+        };
+        Some(TranslatedPredicate {
+            predicate,
+            // Paimon's core residual follows Java Float.compare / Double.compare
+            // and canonicalizes all NaNs. DataFusion's Arrow equality keeps NaN
+            // payloads distinct, so retain its residual for floating arrays.
+            requires_residual: matches!(
```
leaves12138, line 411: Keeping a DataFusion residual is not sufficient once this inexact predicate is negated. For `NOT array_has(float_array, NaN(payload A))`, a row containing `NaN(payload B)` matches the Paimon inner predicate because all NaNs are canonicalized, but does not match the DataFusion inner predicate because the payloads differ. After `NOT`, the Paimon predicate rejects a row that DataFusion should keep, and the residual cannot recover a row already removed by pushdown. Please avoid translating `NOT` when the translated child already has `requires_residual` (or otherwise disable floating-array pushdown under negation), and add a regression test for distinct NaN payloads.

leaves12138 (changes requested):
Requesting changes for a correctness issue in negated floating-point array predicate pushdown.

leaves12138 (approved):
Re-reviewed commit f902cf2. Negated inexact predicates now fall open before pushdown, preventing the NaN-payload semantic mismatch from becoming false negatives. I also checked the analyzer-inserted lossless array casts and the overflow-safe decimal comparison; the DataFusion filter-pushdown tests and targeted extreme-scale decimal tests pass locally. LGTM.

PR #733, opened 2026-08-20
feat: route tables by declared TableType to pluggable engines

plusplusjiajia:
### Purpose
REST catalogs can declare table types a Paimon reader cannot serve via the `type` table option (e.g. `type=iceberg-table`). paimon-datafusion tries to read every table as Paimon, so these tables fail with unrelated errors. This PR routes them to pluggable engines instead:

  - `paimon::spec::TableType` mirrors Java's `org.apache.paimon.TableType`, and `CoreOptions::table_type()` returns it, so DataFusion decides from the enum rather than from type strings.
  - **`Catalog::load_table`.** A defaulted method returning `LoadedTable::{Paimon, External}`, so existing implementations keep working. Classification is unconditional — the outcome depends only on the table's own metadata, never on what the caller registered. Both built-in catalogs override it to classify before construction, matching Java's shared dispatch in `CatalogUtils.loadTable`; the default classifies from the constructed table, so a catalog that only implements `get_table` still fails closed.
  - `object-table`, `lance-table` and `iceberg-table` all require an engine: Java builds a dedicated table for each, this client has none, so reading one as Paimon would misread it and writing could put Paimon
  snapshots over foreign data. Both catalogs fail closed on them, covering every raw `get_table` path. Raw metadata stays describable, so `paimon-rest-server` can still answer GetTable and let its clients route.
  - An unknown declared type is rejected before it is persisted, on create and alter alike, and the declared type is immutable afterwards — mirroring `SchemaValidation` and `SchemaManager`.
  - ExternalTableMetadata can only be built through LoadedTable::external, which rejects a type the Paimon reader serves and runs the stored-metadata checks.
  - Historical reads fail loudly rather than silently: every version clause, session scan selectors, and selectors stored in the table's own options are rejected instead of being dropped and answered from current data. `register_catalog_table_engine(&ctx, ..)` installs `PaimonRelationPlanner`, so that holds without `SQLContext` too.
  - Routing lives inside `PaimonCatalogProvider` so downcast paths keep working; `table_exist` mirrors resolution, and engine-resolved providers are wrapped read-only.

  `EngineTableRequest` is `#[non_exhaustive]`, so a follow-up can carry a snapshot selector — with an opt-in on the resolver, so an engine that ignores it never answers from current data — without breaking existing resolvers. A follow-up also adds an Iceberg-backed resolver crate.

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -34,6 +34,7 @@ use datafusion::sql::sqlparser::ast::{Ident, ObjectName, Query, Statement, Visit
 use datafusion::sql::sqlparser::dialect::GenericDialect;
 use datafusion::sql::sqlparser::parser::Parser;
 use paimon::catalog::{Catalog, Identifier, View};
+use paimon::spec::TableType as PaimonTableType;
 
 use crate::error::to_datafusion_error;
 use crate::runtime::{await_with_runtime, block_on_with_runtime};
@@ -45,12 +46,12 @@ pub(crate) type SessionStateProvider = Arc<dyn Fn() -> Option<SessionState> + Se
 
 /// Engine registry shared between the catalog provider and its schema
 /// providers, so registrations stay visible to schemas obtained earlier.
-type TableEngines = Arc<RwLock<HashMap<String, Arc<dyn TableEngineResolver>>>>;
+type TableEngines = Arc<RwLock<HashMap<PaimonTableType, Arc<dyn TableEngineResolver>>>>;
 
-/// Resolves tables owned by a non-Paimon engine, selected by the declared
-/// `type` table option (see [`PaimonCatalogProvider::register_table_engine`]).
-/// `Ok(None)` means not found; errors are propagated verbatim so an engine
-/// failure never masquerades as a missing table.
+/// Resolves tables owned by another engine (see
+/// [`PaimonCatalogProvider::register_table_engine`]). `Ok(None)` means not
+/// found; errors propagate, so an engine failure never looks like a missing
+/// table.
 #[async_trait]
 pub trait TableEngineResolver: Debug + Send + Sync {
     /// Resolve `database.table` to the engine's table provider.
@@ -62,12 +63,11 @@ pub trait TableEngineResolver: Debug + Send + Sync {
 }
 
 /// Read-only wrapper around an engine-resolved provider: reads delegate,
-/// DML is rejected. Keeps routed tables on the same fail-closed stance as
-/// the raw `get_table` paths even when the engine's provider is writable.
+/// DML is rejected even when the engine's own provider is writable.
 #[derive(Debug)]
 struct ReadOnlyTableProvider {
     inner: Arc<dyn TableProvider>,
-    declared: String,
+    declared: PaimonTableType,
     table_name: String,
 }
 
@@ -154,8 +154,8 @@ pub struct PaimonCatalogProvider {
     blob_reader_registry: BlobReaderRegistry,
     session_state: Option<SessionStateProvider>,
     schema_force_view_types: bool,
-    /// Engines for non-Paimon table types, keyed by the declared `type` table
-    /// option. Same poison-recovery stance as `temp_tables`.
+    /// Engines for table types served elsewhere, keyed by declared
+    /// [`PaimonTableType`]. Same poison-recovery stance as `temp_tables`.
     table_engines: TableEngines,
 }
 
@@ -196,25 +196,22 @@ impl PaimonCatalogProvider {
         self
     }
 
-    /// Register an engine for a non-Paimon table type (e.g. `iceberg-table`).
-    /// References whose metadata declares that type resolve through
-    /// `resolver`; everything else takes the Paimon path unchanged. Kept
-    /// inside the provider so the registered catalog type never changes and
-    /// downcast-based paths (temp tables, time travel) keep working.
+    /// Register an engine for a table type the Paimon reader cannot serve
+    /// (e.g. [`PaimonTableType::IcebergTable`]); everything else takes the
+    /// Paimon path unchanged. Kept inside the provider so the registered
+    /// catalog type never changes and downcast-based paths (temp tables,
+    /// time travel) keep working.
     pub fn register_table_engine(
         &self,
-        table_type: impl Into<String>,
+        table_type: PaimonTableType,
         resolver: Arc<dyn TableEngineResolver>,
     ) -> DFResult<()> {
-        // Metadata publishes lowercase TableType values.
-        let table_type = table_type.into().to_ascii_lowercase();
-        // A Paimon-managed type would be split between engines: reads via
-        // the resolver, raw get_table paths via Paimon.
-        if !paimon::spec::is_non_paimon_table_type(&table_type) {
+        // Routing a Paimon-served type would split it between engines:
+        // reads via the resolver, raw get_table paths via Paimon.
+        if !table_type.requires_table_engine() {
             return Err(plan_datafusion_err!(
-                "table type '{table_type}' is Paimon-managed and cannot be routed to a \
-                 table engine; routable types: {:?}",
-                paimon::spec::NON_PAIMON_TABLE_TYPES
+                "table type '{table_type}' is served by the Paimon reader and cannot be \
+                 routed to a table engine"
             ));
         }
         self.table_engines
@@ -476,7 +473,7 @@ pub struct PaimonSchemaProvider {
     blob_reader_registry: BlobReaderRegistry,
     session_state: Option<SessionStateProvider>,
     schema_force_view_types: bool,
-    /// Engines for non-Paimon table types; empty without routing.
+    /// Engines for table types served elsewhere; empty without routing.
     table_engines: TableEngines,
 }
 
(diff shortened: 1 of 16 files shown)


crates/paimon/src/spec/table_type.rs
```diff
@@ -0,0 +1,134 @@
...
+            TableType::LanceTable => "lance-table",
+            TableType::IcebergTable => "iceberg-table",
+        }
+    }
+
+    /// Whether the Paimon reader cannot serve this type, so it must be
+    /// routed to a table engine (see
+    /// [`Catalog::load_table_routing`](crate::catalog::Catalog::load_table_routing)).
+    pub fn requires_table_engine(&self) -> bool {
+        matches!(self, TableType::IcebergTable)
```
JingsongLi: [P1] Route every table kind the Rust reader cannot serve

This classifies only Iceberg as engine-required, so register_table_engine rejects both ObjectTable and LanceTable as if the ordinary Paimon reader supported them. It does not: the Rust read stack has no Object/Lance provider and its normal scan uses Paimon snapshots (including an empty plan when none exists), while Java CatalogUtils constructs dedicated Object, Lance, and Iceberg table implementations. A real object-table or lance-table can therefore fail or silently appear empty, and callers cannot register the resolver introduced here. Please mark every currently non-native type (at least Object/Lance/Iceberg) as engine-required/fail-closed until native providers exist, and replace the Lance rejection test with routing coverage.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -465,9 +603,39 @@ impl SchemaProvider for PaimonSchemaProvider {
...
+                        ));
+                    }
+                    let resolver = table_engines
+                        .get(&declared)
+                        .expect("declared type came from this engine map");
+                    let resolved = resolver
+                        .resolve_table(identifier.database(), identifier.object())
+                        .await?;
+                    // Read-only wrap: DML must not reach the engine provider.
+                    Ok(resolved.map(|inner| {
```
JingsongLi, line 700: [P1] Reject or implement time travel for routed providers

Returning this wrapper breaks the existing raw SQLContext::ctx().sql time-travel path: PaimonRelationPlanner cannot downcast it to PaimonTableProvider, returns Original, and DataFusion 54's default relation planner discards TableFactor.version. I reproduced this on this head: after selecting the Databricks dialect, SELECT * FROM ...it VERSION AS OF 999999 succeeded and returned the current two engine rows instead of rejecting the nonexistent version. The high-level SQLContext::sql path fails closed, but the public raw context is intentionally supported and tested for time travel. Please make the relation planner recognize routed providers and explicitly reject VERSION/TIMESTAMP, or extend the resolver contract to carry those selectors.

JingsongLi (commented):
Detailed review found three correctness regressions and one creation-validation gap. I verified the current head locally; details are inline.

crates/paimon/src/catalog/filesystem.rs
```diff
@@ -297,30 +354,26 @@ impl Catalog for FileSystemCatalog {
...
-                full_name: identifier.full_name(),
-            });
+    async fn load_table_routing(
+        &self,
+        identifier: &Identifier,
+        engine_types: &std::collections::HashSet<TableType>,
+    ) -> Result<crate::catalog::RoutedTableLoad> {
+        let (table_path, schema) = self.fetch_table_schema(identifier).await?;
+        let options = CoreOptions::new(schema.options());
+        let declared = options.table_type()?;
```
JingsongLi, line 362: [P1] Keep type immutable before trusting it for routing

This makes the latest schema's type an authoritative engine switch, but TableSchema::apply_changes still accepts arbitrary SetOption/RemoveOption, and SQL exposes that through ALTER TABLE ... SET TBLPROPERTIES. A populated Paimon table can be changed to type=iceberg-table; with a resolver, later reads are redirected to a same-named engine table, and without one the existing snapshots become inaccessible. Java SchemaManager explicitly rejects semantic changes/removal of CoreOptions.TYPE, even before snapshots exist. Please enforce the same invariant before saving a schema (allowing only case-insensitive/default-equivalent no-ops) and cover filesystem plus REST ALTER paths.

crates/paimon/src/spec/core_options.rs
```diff
@@ -624,11 +626,17 @@ impl<'a> CoreOptions<'a> {
             .unwrap_or(false)
     }
 
+    /// The declared [`TableType`], defaulting to [`TableType::Table`].
+    /// Fails on a value this client does not know.
+    pub fn table_type(&self) -> crate::Result<TableType> {
+        match self.options.get(TABLE_TYPE_OPTION) {
+            Some(value) => value.parse(),
```
JingsongLi, line 633: [P2] Validate the declared type before persisting schema-0

This validation currently first runs while loading/routing. FileSystemCatalog::create_table still creates the directory and writes schema-0 without calling table_type(), and the bundled REST server delegates to it. Thus CREATE TABLE ... WITH ('type'='iceberg-tabel') returns success but every later get_table/routing call fails with unknown table type, leaving an unusable catalog entry. Please invoke this parsing from shared create/schema validation before any metadata write and add filesystem/REST tests that assert rejection leaves no table artifacts.

crates/paimon/src/spec/table_type.rs
plusplusjiajia: @JingsongLi Good catch. Done: object-table and lance-table now require an engine too, and both catalogs fail closed.

crates/paimon/src/catalog/filesystem.rs
plusplusjiajia: @JingsongLi  Agreed, rejecting unknown values wasn't enough. Done: alter_table rejects any change or removal of type, allowing only case-insensitive no-ops, matching SchemaManager.

crates/paimon/src/spec/core_options.rs
plusplusjiajia: @JingsongLi Done: the parse runs in Schema::validate_final_schema, shared by create and alter.

crates/integrations/datafusion/src/catalog.rs
plusplusjiajia: @JingsongLi The reproduction steps helped. Sweeping that surface turned up three more ways in: session selectors, selectors stored in the table's own options, and unsupported scan options. All rejected now, and registration installs the relation planner so the rejection holds without SQLContext.

JingsongLi:
[P2] Keep DataFusion engine registration out of the core Catalog contract

Could we avoid `Catalog::load_table_routing(..., engine_types)`? `engine_types` is DataFusion/session state, not catalog metadata. Passing it into `paimon::Catalog` makes loading the same table depend on which integration/resolvers are currently registered, and it forces third-party catalogs to implement a DataFusion-specific routing protocol (while the default implementation ignores the set).

I do not think simply calling `get_table()` and then inspecting `TableType` in DataFusion is sufficient: by then FS/REST has entered the concrete Paimon `Table` construction path; REST may acquire a data token and construct FileIO, and the external-type fail-closed guard must run before that construction. Calling `get_table_type()` and then `get_table()` would also fetch metadata twice for native tables.

Could core instead expose an unconditional classification, for example:

```rust
async fn load_table(&self, id: &Identifier) -> Result<LoadedTable>;

enum LoadedTable {
    Paimon(Box<Table>),
    External(ExternalTableMetadata),
}
```

The catalog would load metadata once and decide only whether the core reader can construct the table. Object/Lance/Iceberg would always return `External`. DataFusion would then look up its own resolver and report a clear error if none is registered. The existing `get_table()` path can remain fail-closed for external types. In other words, core owns classification and stored-metadata validation; DataFusion owns resolver selection, session semantics, and the read-only wrapper.

This would also remove a hidden invariant in the current public API: `RoutedTableLoad::engine` does not check `declared.requires_table_engine()`. A non-DataFusion caller can include a native type in `engine_types` and make the FS/REST implementations skip native table construction. Correctness currently relies on the private check in the DataFusion registration path.

This boundary also matches the Java direction more closely: core dispatches from table metadata without accepting caller-registered engine types.

plusplusjiajia:
> [P2] Keep DataFusion engine registration out of the core Catalog contract
> 
> Could we avoid `Catalog::load_table_routing(..., engine_types)`? `engine_types` is DataFusion/session state, not catalog metadata. Passing it into `paimon::Catalog` makes loading the same table depend on which integration/resolvers are currently registered, and it forces third-party catalogs to implement a DataFusion-specific routing protocol (while the default implementation ignores the set).
> 
> I do not think simply calling `get_table()` and then inspecting `TableType` in DataFusion is sufficient: by then FS/REST has entered the concrete Paimon `Table` construction path; REST may acquire a data token and construct FileIO, and the external-type fail-closed guard must run before that construction. Calling `get_table_type()` and then `get_table()` would also fetch metadata twice for native tables.
> 
> Could core instead expose an unconditional classification, for example:
> 
> ```rust
> async fn load_table(&self, id: &Identifier) -> Result<LoadedTable>;
> 
> enum LoadedTable {
>     Paimon(Box<Table>),
>     External(ExternalTableMetadata),
> }
> ```
> 
> The catalog would load metadata once and decide only whether the core reader can construct the table. Object/Lance/Iceberg would always return `External`. DataFusion would then look up its own resolver and report a clear error if none is registered. The existing `get_table()` path can remain fail-closed for external types. In other words, core owns classification and stored-metadata validation; DataFusion owns resolver selection, session semantics, and the read-only wrapper.
> 
> This would also remove a hidden invariant in the current public API: `RoutedTableLoad::engine` does not check `declared.requires_table_engine()`. A non-DataFusion caller can include a native type in `engine_types` and make the FS/REST implementations skip native table construction. Correctness currently relies on the private check in the DataFusion registration path.
> 
> This boundary also matches the Java direction more closely: core dispatches from table metadata without accepting caller-registered engine types.

@JingsongLi  Good call on the boundary — engine_types is gone from core. Catalog::load_table returns LoadedTable::{Paimon, External} and classifies unconditionally, so the outcome depends only on the table's own metadata. Core owns classification and the stored-metadata checks; DataFusion owns resolver selection, session semantics, and the read-only wrapper. 
The hidden invariant is closed too: ExternalTableMetadata can only be built through LoadedTable::external, which rejects any type requires_table_engine does not accept.

crates/paimon/src/catalog/mod.rs
```diff
@@ -323,6 +371,20 @@ pub trait Catalog: Send + Sync {
...
 
+    /// Load a table, or classify it as [`LoadedTable::External`] when this
+    /// reader cannot construct it. One metadata round-trip either way, and the
+    /// outcome depends only on the table's own metadata. The default
+    /// implementation always constructs, for catalogs without a table-type
+    /// concept.
+    ///
+    /// # Errors
+    /// Same as [`Catalog::get_table`].
+    async fn load_table(&self, identifier: &Identifier) -> Result<LoadedTable> {
```
JingsongLi, line 382: [P1] Classify the table in the default `load_table` implementation

This default is the compatibility path for existing third-party `Catalog` implementations, but it unconditionally wraps the `Table` returned by `get_table` as `LoadedTable::Paimon`. If such a catalog returns a table whose stored `type` is `object-table`, `lance-table`, or `iceberg-table`, `PaimonSchemaProvider::table` takes the Paimon arm and never consults the registered resolver. `PaimonTableProvider` has no later type guard, so reads can misread the foreign table (or silently appear empty), and INSERT can write Paimon metadata into the foreign location.

This also contradicts the PR description that the default "classifies from the constructed table" so a catalog implementing only `get_table` still fails closed. Please first load the table, parse `CoreOptions::new(table.schema().options()).table_type()?`, and call the checked `LoadedTable::external` constructor when `requires_table_engine()` is true; otherwise return the Paimon variant. The built-in overrides can remain to preserve pre-construction classification and avoid token/FileIO work. A regression test should use a catalog that deliberately does not override `load_table`.

JingsongLi (approved):
+1

PR #735, opened 2026-08-21
fix(datafusion): make table definitions optional for queries

jerry-024:
### Purpose

`vector_search` creates a `PaimonTableProvider` while planning a query. The provider eagerly built a `CREATE TABLE` definition, so tables containing `VECTOR` columns failed before search execution because `VECTOR` cannot currently be round-tripped by `SHOW CREATE TABLE`.

Table definitions are optional query metadata and should not prevent a table from being queried.

### Brief change log

- Treat failure to build the optional table definition as non-blocking when creating a `PaimonTableProvider`.
- Keep explicit `SHOW CREATE TABLE` validation unchanged.
- Add a regression test that plans `vector_search` against a `VECTOR<FLOAT>` column.

### Tests

- Added `test_vector_search_plans_vector_column_without_show_create_support`

### API and Format

### Documentation

### DIFF
--- a/crates/integrations/datafusion/src/table/mod.rs
+++ b/crates/integrations/datafusion/src/table/mod.rs
@@ -113,8 +113,8 @@ impl PaimonTableProvider {
     ///
     /// Loads the table schema and converts it to Arrow for DataFusion.
     pub fn try_new(table: Table) -> DFResult<Self> {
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #736, opened 2026-08-21
perf(vindex): parallelize Parquet reads for index builds

jerry-024:
### Purpose

Speed up vector-index builds by allowing full-file scans to use parallel Parquet row-group reads.

### Changes

- Treat full-coverage `row_ranges` without a deletion vector as an unfiltered scan, allowing the existing parallel row-group path to run. Partial ranges, deletion vectors, and empty ranges keep their previous semantics.
- Keep oversized row groups charged against the full configured Parquet read budget, preserving `max_inflight_bytes` as a strict projected-byte admission limit, and warn once when a row group exceeds it.
- Add opt-in aggregate diagnostics for vector-index build phases and Parquet row-group reads.
- Add a production-path IVF-PQ build benchmark example.

### Reproduce the IVF-PQ build benchmark

The benchmark is in [`ivfpq_build_benchmark.rs`](https://github.com/apache/paimon-rust/blob/46d22b7/crates/paimon/examples/ivfpq_build_benchmark.rs). It builds an IVF-PQ index with dimension 768, cosine distance, `nlist=4096`, and `pq.m=192` on an existing Paimon table:

```bash
PAIMON_CATALOG_OPTIONS='<catalog-options-json>' \
PAIMON_LOG_VECTOR_INDEX_BUILD_TIMING=1 \
cargo run --release -p paimon --example ivfpq_build_benchmark -- \
  <database> <table> <vector-column> --drop-existing
```

### Benchmark

Cohere 10M × 768 dimensions, cosine IVF-PQ (`nlist=4096`, `pq.m=192`, 8 bits), 10 Parquet files / 100 row groups, direct OSS, and no local cache.

#### Parallel Parquet reads

Both baseline and this PR use an explicit 512 MiB Parquet read budget. The metrics below cover only the Parquet-read work changed by this PR.

| metric | baseline | this PR | change |
|---|---:|---:|---:|
| source batch wait | 105.346s | **51.148s** | **−51.45%** |
| peak in-flight row groups | 1 | **4** | +3 |

The current-PR values are from the latest single validation run. The 512 MiB budget is a workload-specific benchmark setting and must be configured explicitly. The default remains 256 MiB.

### Tests

- Full, stitched, partial, empty, and deletion-vector row selections, including `_ROW_ID` parity.
- Shared Parquet budget accounting and peak in-flight diagnostics.
- Oversized row groups consume the full budget and emit a one-time warning; ordinary and tiny row-group accounting remains strict.
- Vector-index build timing aggregation and existing builder tests.

### API and format

No public API or file/index format changes.

### DIFF
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -910,6 +910,9 @@ fn predicate_supported_for_parquet_row_filter(op: PredicateOperator) -> bool {
             | PredicateOperator::Like
             | PredicateOperator::Between
             | PredicateOperator::NotBetween
+            | PredicateOperator::ArrayContains
+            | PredicateOperator::ArraysOverlap
+            | PredicateOperator::ArrayContainsAll
     )
 }
 
@@ -969,6 +972,27 @@ fn parquet_row_filter_literals_supported(
             for literal in literals {
                 if crate::arrow::residual::literal_scalar_for_arrow_filter(literal, file_data_type)?
                     .is_none()
+                {
+                    return Ok(false);
+                }
+            }
+            Ok(true)
+        }
+        PredicateOperator::ArrayContains
+        | PredicateOperator::ArraysOverlap
+        | PredicateOperator::ArrayContainsAll => {
+            let DataType::Array(array_type) = file_data_type else {
+                return Ok(false);
+            };
+            if matches!(op, PredicateOperator::ArrayContains) && literals.len() != 1 {
+                return Ok(false);
+            }
+            for literal in literals {
+                if crate::arrow::residual::literal_scalar_for_arrow_filter(
+                    literal,
+                    array_type.element_type(),
+                )?
+                .is_none()
                 {
                     return Ok(false);
                 }
--- a/crates/paimon/src/arrow/format/parquet.rs
+++ b/crates/paimon/src/arrow/format/parquet.rs
@@ -490,29 +490,36 @@ impl FormatFileReader for ParquetFormatReader {
         // preserving positional `_ROW_ID`, sort order, and batch backpressure. Reads
         // with predicates or an explicit row selection retain the original
         // single-stream path until their selections are split per row group.
-        let row_group_parallelism = self
-            .read_budget
-            .as_ref()
-            .filter(|_| preds.is_empty() && row_filter_factory.is_none() && row_selection.is_none())
+        let read_budget = self.read_budget.as_ref().filter(|_| {
+            preds.is_empty() && row_filter_factory.is_none() && row_selection.is_none()
+        });
+        let row_group_parallelism = read_budget
             .map(|budget| {
                 budget
                     .parallelism()
                     .min(batch_stream_builder.metadata().num_row_groups())
             })
             .unwrap_or(1);
+        let projected_bytes = read_budget
+            .filter(|budget| row_group_parallelism > 1 || budget.diagnostics_enabled())
+            .map(|budget| {
+                let projected_bytes = batch_stream_builder
+                    .metadata()
+                    .row_groups()
+                    .iter()
+                    .map(|row_group| projected_row_group_bytes(row_group, &mask))
+                    .collect::<Vec<_>>();
+                budget.record_projected_row_groups(&projected_bytes);
+                projected_bytes
+            });
         if row_group_parallelism > 1 {
             let row_group_count = batch_stream_builder.metadata().num_row_groups();
             let reader_metadata = ArrowReaderMetadata::try_new(
                 batch_stream_builder.metadata().clone(),
                 ArrowReaderOptions::new(),
             )?;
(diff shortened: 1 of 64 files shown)


shyjsarah (changes requested):
Thanks for the optimization work. I found one issue that I think should be addressed before merge:

### Concurrent multipart uploads are not bounded to 32 MiB

`async_writer_with_concurrency(4)` documents the buffering cost as `4 * 8 MiB`, but that is not the behavior of the pinned OpenDAL 0.58 implementation.

`MultipartWriter` creates `ConcurrentTasks` with `prefetch = 8192`. With concurrency greater than one, completed tasks are not continuously reaped during `write`; each completed task still owns its `WriteInput`, including the 8 MiB part buffer, until tasks are drained during `close()`. If uploads complete faster than index serialization, a ~2 GiB index can therefore retain close to the whole index size in completed upload buffers, rather than ~32 MiB. The previous concurrency-1 path executes inline and does not accumulate these buffers.

Please either keep concurrency at 1 for now, use/patch an OpenDAL implementation with strictly bounded completed-task buffering, or introduce a bounded uploader that continuously reaps completed parts. A fast fake multipart backend test asserting peak retained bytes would also help prevent regression.

Two non-blocking diagnostics/benchmark notes:

- For the target 768-dimensional workload, a ~294 MiB row group consumes the entire default 256 MiB read budget, so the parallel path can still run effectively serially unless the option is tuned. It would be useful for the benchmark to print the effective budget and observed peak concurrency, and warn when the peak is 1.
- Parquet diagnostics are collected only after `read_budget` is filtered for parallel-path eligibility. Partial row selections can perform Parquet I/O while reporting zero row-group/projected-byte diagnostics. Consider decoupling diagnostics collection from fast-path eligibility, or naming the fields as parallel-path-only metrics.

crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -490,29 +490,36 @@ impl FormatFileReader for ParquetFormatReader {
...
+            preds.is_empty() && row_filter_factory.is_none() && row_selection.is_none()
+        });
+        let row_group_parallelism = read_budget
             .map(|budget| {
                 budget
                     .parallelism()
                     .min(batch_stream_builder.metadata().num_row_groups())
             })
             .unwrap_or(1);
+        let projected_bytes = read_budget
```
JingsongLi: Because `read_budget` has already been filtered by `row_selection.is_none()`, any partial-row-range read reaches this block with `read_budget == None`. Vector-index shard boundaries can cut through a file, so these reads still perform Parquet I/O while reporting `parquet_row_group_count` and all projected-byte diagnostics as zero. Please compute and record the diagnostic sizes from `self.read_budget` independently of parallel-path eligibility, or rename these fields to make their parallel-path-only scope explicit.

JingsongLi:
[P2] Restore a bounded upload path or drop the multipart claim

At the current head aa9eb0e, vindex_index_build_builder.rs still calls async_writer(), which delegates to async_writer_with_concurrency(1). The new async_writer_with_concurrency helper has no caller requesting concurrency greater than 1; commit 23e352d reverted the only concurrency=4 call.

As a result, vector-index uploads still serialize their 8 MiB parts, so an index larger than one part continues to pay one upload round trip per part. The PR title, Changes section, and the 30.3s to 4.6s benchmark therefore no longer describe the code being reviewed.

Please either add a genuinely bounded concurrent upload implementation and use it here, or remove the unused helper and update the title, description, and benchmark scope to cover only the Parquet-read improvement. Simply restoring the previous OpenDAL concurrency setting would reintroduce the unbounded buffering issue already identified in this review.

crates/paimon/src/arrow/format/parquet.rs
```diff
@@ -490,29 +490,38 @@ impl FormatFileReader for ParquetFormatReader {
...
+            .as_ref()
+            .filter(|budget| row_group_parallelism > 1 || budget.diagnostics_enabled())
+            .map(|budget| {
+                let projected_bytes = batch_stream_builder
+                    .metadata()
+                    .row_groups()
+                    .iter()
+                    .map(|row_group| projected_row_group_bytes(row_group, &mask))
+                    .collect::<Vec<_>>();
+                budget.record_projected_row_groups(&projected_bytes);
```
shyjsarah, line 522: Non-blocking: this avoids the previous all-zero diagnostics for selected reads, but it now records metadata for **all** row groups in the file rather than only the row groups touched by the effective selection. The added test selects rows `0..9` from a two-row-group file and expects `row_group_count == 2`, even though only the first row group is involved. As a result, `parquet_row_group_count` and the projected-byte min/max/total can overstate the work performed by partial reads.

Could we either filter these values by the effective `RowSelection`, or rename/document them as file-level projected metadata rather than read-level diagnostics?

shyjsarah (approved):
+1

JingsongLi (approved):
+1

PR #739, opened 2026-08-23
feat: expose point-query primitives in paimon

JingsongLi:
### Purpose

Expose the minimal Paimon core primitives needed by snapshot-consistent point-query consumers, independently of any HTTP service implementation.

Linked issue: none yet.

### Brief change log

- Add public Blob descriptor version and range validation APIs while reusing the same validation for payload reads.
- Add a typed `SnapshotNotExist` error and a strict time-travel table copy API that propagates selector failures.
- Expose the lower-bound sum of known data-file bytes on `Plan` and `ScanTrace` for query budget enforcement.

This PR intentionally contains only changes under `crates/paimon`. The query service is tracked separately in #738.

### Tests

- `cargo test -p paimon --lib` (2376 passed, 2 ignored)
- `cargo test -p paimon spec::blob_descriptor::tests --lib` (11 passed)
- `cargo test -p paimon table::source::tests --lib` (51 passed)

### API and Format

Adds public Rust APIs for strict snapshot selection, typed missing-snapshot errors, Blob descriptor validation, and planned data-file byte statistics. No storage or serialization format changes.

### Documentation

Public APIs include rustdoc explaining strict selection, descriptor validation, and lower-bound byte accounting.

### DIFF
--- a/crates/paimon/src/table/source.rs
+++ b/crates/paimon/src/table/source.rs
@@ -1295,6 +1295,19 @@ impl Plan {
         &self.splits
     }
 
+    /// Sum of data-file bytes referenced by this plan.
+    ///
+    /// Negative file sizes are treated as unknown and do not contribute. The
+    /// result is therefore a lower bound when a connector cannot provide every
+    /// file size.
+    pub fn planned_data_file_bytes(&self) -> u64 {
+        self.splits
+            .iter()
+            .flat_map(DataSplit::data_files)
+            .filter_map(|file| u64::try_from(file.file_size).ok())
+            .sum()
+    }
+
     /// Consume this plan and return its splits without cloning their file metadata.
     #[must_use = "consuming a plan without using its splits drops the planned work"]
     pub fn into_splits(self) -> Vec<DataSplit> {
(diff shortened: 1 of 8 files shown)


crates/paimon/src/table/source.rs
```diff
@@ -1295,6 +1295,19 @@ impl Plan {
...
+    ///
+    /// Negative file sizes are treated as unknown and do not contribute. The
+    /// result is therefore a lower bound when a connector cannot provide every
+    /// file size.
+    pub fn planned_data_file_bytes(&self) -> u64 {
+        self.splits
+            .iter()
+            .flat_map(DataSplit::data_files)
+            .filter_map(|file| u64::try_from(file.file_size).ok())
+            .sum()
```
leaves12138: Could we make this accumulation overflow-safe (for example, with `saturating_add`)? `sum::<u64>()` panics with overflow checks enabled and wraps in release builds when the total known file bytes exceeds `u64::MAX`. Because this value is used for query-budget enforcement, wrapping can under-report an oversized plan and let it pass the budget check. Saturating at `u64::MAX` preserves the safety property; please also add an overflow regression test.

JingsongLi: Fixed in 1f868f61. The accumulation now uses saturating_add, preserving negative-size filtering while saturating oversized plans at u64::MAX. Added a regression test whose known sizes total exactly 2^64; it passes in both debug and release, while the old implementation would panic in debug and wrap to zero in release.

leaves12138 (approved):
LGTM. The overflow-safe byte accounting and regression test look good.

PR #742, opened 2026-08-24
fix: align manifest entry schema with Java Paimon

### LINKED ISSUE #741: Align ManifestEntry Avro field order with Java Paimon
### Description

`MANIFEST_ENTRY_SCHEMA` currently writes the top-level ManifestEntry fields in this order:

```text
_KIND, _PARTITION, _BUCKET, _TOTAL_BUCKETS, _FILE, _VERSION
```

Java Paimon's `VersionedObjectSerializer` prepends `_VERSION`, and its manifest reader expects the canonical order:

```text
_VERSION, _KIND, _PARTITION, _BUCKET, _TOTAL_BUCKETS, _FILE
```

As a result, a manifest written by Paimon Rust can fail when read by Java Paimon with an error like:

```text
Unexpected Manifest Avro field at position 0: expected _VERSION but found _KIND
```

### Proposed fix

- Move `_VERSION` to the first field in `ManifestEntry` and `MANIFEST_ENTRY_SCHEMA`.
- Add a regression test for the Java-compatible field order.
- Verify that the Rust reader still reads manifests written with the previous Rust field order. The current Rust decoder dispatches fields by the writer-schema field name, so old files should remain readable.

### Impact

This affects Java/Rust interoperability for tables whose manifests are written by Paimon Rust and later read by Java Paimon.

shyjsarah:
### Purpose

Linked issue: close #741

Paimon Rust currently writes `_VERSION` as the last top-level field of a manifest entry. Java Paimon's versioned object format requires `_VERSION` at position 0, so Java manifest readers reject Rust-written manifests with:

```text
Unexpected Manifest Avro field at position 0: expected _VERSION but found _KIND
```

The changelog commit path also used format identifier `0`, while Java's `ManifestEntrySerializer` accepts identifier `2` only.

### Brief change log

- Move `_VERSION` to the first field in `ManifestEntry` and `MANIFEST_ENTRY_SCHEMA`.
- Use format identifier `2` for both data and changelog manifest entries.
- Add a regression test for the Java-compatible top-level field order.
- Add a compatibility test proving the Rust reader can still read manifests written with the previous Rust field order.
- Add an end-to-end changelog commit assertion that reads the generated Avro manifest and verifies `_VERSION = 2`.

### Tests

- `cargo test -p paimon manifest_entry_ -- --nocapture`
- `cargo test -p paimon changelog_manifest -- --nocapture`
- `cargo test -p paimon table::table_commit::tests::test_commit_writes_changelog_manifest_list_size -- --nocapture`
- `PYO3_PYTHON=/opt/homebrew/bin/python3.11 cargo clippy --locked --all-targets --workspace --features fulltext,vortex -- -D warnings`
  - The library suite passed all 2472 executed tests.
  - Local mock REST integration tests were flaky with localhost `ConnectionReset` / `IncompleteMessage`; the initially failing token refresh test passed when rerun individually. The failures are unrelated to the manifest changes.

### API and Format

No API change.

This changes newly written ManifestEntry Avro files to the canonical Java field order and writes Java's permanent format identifier `2` for changelog entries. Existing manifests written by older Paimon Rust versions remain readable by the Rust reader because decoding follows the embedded writer schema by field name.

### Documentation

No documentation update is required.

### DIFF
--- a/crates/paimon/src/spec/manifest_entry.rs
+++ b/crates/paimon/src/spec/manifest_entry.rs
@@ -55,6 +55,10 @@ impl Hash for Identifier {
 /// Impl Reference: <https://github.com/apache/paimon/blob/release-0.8.2/paimon-core/src/main/java/org/apache/paimon/manifest/ManifestEntry.java>
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ManifestEntry {
+    // Java's VersionedObjectSerializer requires the version field at position 0.
+    #[serde(rename = "_VERSION")]
+    version: i32,
+
     #[serde(rename = "_KIND")]
     kind: FileKind,
 
@@ -69,9 +73,6 @@ pub struct ManifestEntry {
 
     #[serde(rename = "_FILE")]
     pub(crate) file: DataFileMeta,
-
-    #[serde(rename = "_VERSION")]
-    version: i32,
 }
 
 #[allow(dead_code)]
@@ -151,12 +152,12 @@ impl ManifestEntry {
         version: i32,
     ) -> Self {
         ManifestEntry {
+            version,
             kind,
             partition,
             bucket,
             total_buckets,
             file,
-            version,
         }
     }
 
@@ -186,6 +187,7 @@ pub const MANIFEST_ENTRY_SCHEMA: &str = r#"["null", {
     "name": "record",
     "namespace": "org.apache.paimon.avro.generated",
     "fields": [
+        {"name": "_VERSION", "type": "int"},
         {"name": "_KIND", "type": "int"},
         {"name": "_PARTITION", "type": "bytes"},
         {"name": "_BUCKET", "type": "int"},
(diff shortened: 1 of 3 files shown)


jerry-024:
One Java compatibility gap remains in the changelog path.

[`messages_to_changelog_entries`](https://github.com/apache/paimon-rust/blob/8ffa33bf4d8a8a3292935dab952c32e2fe04d3b3/crates/paimon/src/table/table_commit.rs#L2862-L2874) creates every changelog `ManifestEntry` with version `0`. These entries are then written through the same `write_manifest_files` path and `MANIFEST_ENTRY_SCHEMA` as regular data manifests.

Java also writes and reads changelog entries through the regular `ManifestFile`, and `ManifestEntrySerializer` accepts only format identifier `2`. As a result, Java reading a Rust-written changelog manifest will reach `checkFormatIdentifier(0)` and fail with `Unsupported version: 0`.

Could we change this value to `2` and add a regression test covering `new_changelog_files`? This PR aligns the top-level field order, but changelog manifests remain incompatible until the version value is aligned as well.

shyjsarah:
@jerry-024 Thanks for catching this. Fixed in `16da285`: changelog `ManifestEntry` now uses format identifier `2`. I also extended `test_commit_writes_changelog_manifest_list_size` to read the generated changelog Manifest Avro file and assert its serialized `_VERSION` is `2`. The test reproduced `Int(0)` before the fix and passes after it. Related manifest/changelog tests, formatting, and workspace Clippy all pass locally.

jerry-024 (approved):
+1

PR #743, opened 2026-08-24
fix: write manifest _FILE as an Avro record

shyjsarah:
## Summary

Follow up on #742 by aligning the Avro type of `ManifestEntry._FILE` with Java Paimon.

Java's `ManifestAvroReader` requires `_FILE` to be a `RECORD`, but paimon-rust currently writes it as a nullable `UNION` (`["null", record]`). This causes Java to reject Rust-written manifests with:

```text
Unexpected Manifest Avro type for field _FILE: expected RECORD but found UNION.
```

## Changes

- Write `_FILE` directly as a non-null Avro record.
- Add a schema regression test that verifies `_FILE` is a non-null record.
- Keep a compatibility test proving paimon-rust can still read manifests written with the previous Rust field order and nullable `_FILE` union.

## Testing

- [x] `cargo test -p paimon --lib` (2380 passed, 2 ignored)
- [x] `cargo clippy -p paimon --all-targets -- -D warnings`

## Notes

- No public API changes.
- Existing Rust-written manifests remain readable by paimon-rust.
- The change only affects the schema used for newly written manifest files.

### DIFF
--- a/crates/paimon/src/spec/manifest_entry.rs
+++ b/crates/paimon/src/spec/manifest_entry.rs
@@ -192,7 +192,7 @@ pub const MANIFEST_ENTRY_SCHEMA: &str = r#"["null", {
         {"name": "_PARTITION", "type": "bytes"},
         {"name": "_BUCKET", "type": "int"},
         {"name": "_TOTAL_BUCKETS", "type": "int"},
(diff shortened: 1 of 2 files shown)


QuakeWang (approved):
LGTM

PR #744, opened 2026-08-25
feat(file_index): add Java-compatible Bloom filter index

QuakeWang:
### Purpose

The file-index foundation only provided generic predicate plumbing and had no concrete Bloom implementation. Rust therefore could not read or produce Java Paimon BloomFilter64 payloads or use them for conservative file pruning.

Linked issue: N/A

### Brief change log

- Implement Java-compatible bitset layout, double hashing, xxHash64, and Thomas Wang hashing.
- Add Bloom reader/writer with conservative `Eq` and `In` evaluation and strict config/payload validation.
- Add Java golden bytes, per-type hash fixtures, and no-false-negative tests.

### Tests

- Dependency and release-license verification

### API and Format

No public API change. The serialized Bloom payload is compatible with Java Paimon. Scan and data-writer integration are not included.

### Documentation

No documentation update is required because the feature is not externally exposed yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4622,6 +4622,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tokio-util",
+ "twox-hash",
  "typed-builder",
  "unicode-segmentation",
  "url",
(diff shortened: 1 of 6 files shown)


JingsongLi (approved):
+1

PR #747, opened 2026-08-26
fix(datafusion): keep metadata queries available without table engines

shyjsarah:
## Purpose

After external table routing was made fail-closed, DataFusion metadata queries such as `information_schema.columns` could fail if a catalog contained any external table whose engine was not registered. This also surfaced through statements that inspect catalog metadata.

## Changes

- preserve catalog field metadata when classifying external tables
- return a metadata-only provider when the declared table engine is unavailable
- expose the real table schema to DataFusion metadata queries
- make `table_exist()` agree with the metadata-only provider for catalog tables without a registered engine
- bypass read-specific session option validation only for metadata-only loading; registered-engine reads still reject unsupported time travel, incremental scan, and query-auth options
- keep scans and writes fail-closed with the existing `no table engine is registered` error
- cover `SHOW TABLES`, unrelated and external `information_schema.columns` queries, direct reads, destructive statements, and `table_exist`

## Behavior

| Operation | Unregistered external table |
| --- | --- |
| `SHOW TABLES` | succeeds and lists the table |
| `information_schema.columns` | succeeds using catalog metadata |
| `SessionContext::table_exist` | returns true for the catalog table |
| direct `SELECT` | fails when the scan is planned |
| writes | fail closed |

## Tests

- `cargo +1.98.0 clippy --locked -p paimon -p paimon-datafusion --all-targets --features fulltext,vortex -- -D warnings`
- `cargo test -p paimon-datafusion --test table_type_routing`

### DIFF
--- a/crates/integrations/datafusion/src/catalog.rs
+++ b/crates/integrations/datafusion/src/catalog.rs
@@ -726,15 +726,6 @@ impl SchemaProvider for PaimonSchemaProvider {
                             identifier.full_name()
                         ));
                     }
-                    // The Paimon arm below applies these; an engine would
-                    // ignore them and answer from current data.
-                    let session_options = dynamic_options
-                        .read()
-                        .unwrap_or_else(|e| e.into_inner())
-                        .clone();
-                    paimon::spec::CoreOptions::new(&session_options)
-                        .ensure_engine_can_serve(&identifier.full_name())
-                        .map_err(to_datafusion_error)?;
                     let Some(resolver) = table_engines.get(&declared) else {
                         let schema = match external.fields() {
                             Some(fields) => crate::table::datafusion_arrow_schema(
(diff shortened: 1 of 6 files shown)


JingsongLi:
- When no engine is present, `table()` returns a metadata-only provider, yet `table_exist()` returns `false`, violating the existence semantics of the DataFusion `SchemaProvider`. 
- Consequently, the same external table can be queried via `information_schema` but appears non-existent when checked via `SessionContext::table_exist`.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -683,13 +735,20 @@ impl SchemaProvider for PaimonSchemaProvider {
                     paimon::spec::CoreOptions::new(&session_options)
                         .ensure_engine_can_serve(&identifier.full_name())
                         .map_err(to_datafusion_error)?;
-                    let resolver = table_engines.get(&declared).ok_or_else(|| {
-                        plan_datafusion_err!(
-                            "no table engine is registered for '{}' tables ('{}')",
+                    let Some(resolver) = table_engines.get(&declared) else {
```
jerry-024, line 729: **[MAJOR] Metadata-only fallback is blocked by read-option validation**

`ensure_engine_can_serve` runs before the missing-resolver branch. Because DataFusion loads every catalog table when planning `information_schema.columns`, an unrelated external table with no registered engine can be rejected by time-travel or incremental-scan options before `UnavailableEngineTableProvider` exposes its schema. This makes a metadata query for another valid table fail during planning.

Could we move the missing-resolver metadata-provider branch before read-specific session-option validation and keep validation on the registered-engine path? Actual reads and writes remain rejected by the unavailable provider.

shyjsarah:
Addressed in d3200f4: when no engine is registered, `table_exist()` now returns true to match the metadata-only provider returned by `table()`. The regression also verifies metadata loading under a session time-travel option.

crates/integrations/datafusion/src/catalog.rs
shyjsarah: Fixed in d3200f4. The missing-resolver metadata fallback now runs before session read-option validation. `ensure_engine_can_serve` remains on the registered-engine path, while scans through the metadata-only provider still fail with the unregistered-engine error.

jerry-024 (approved):
+1

JingsongLi (approved):
+1

PR #749, opened 2026-08-26
feat: add native object table read support

shyjsarah:
## Summary

Add native, read-only Object Table support to Paimon Rust and the DataFusion integration. The implementation follows Paimon Java's fixed Object Table schema and path semantics, so Object Tables no longer require an externally registered table engine for reads.

## Changes

- Add `ObjectTable` and `ObjectEntry` with the Java-compatible fixed schema: `path`, `name`, `length`, `mtime`, `atime`, and `owner`.
- Align creation with Java behavior:
  - ignore user-defined fields and persist the fixed six-column schema
  - default a missing or blank `path` to the catalog table directory
  - preserve an explicit non-empty `path`
  - load legacy filesystem metadata without `path` from the table directory
- Recursively expose object metadata with relative paths through a lazy OpenDAL stream.
- Execute DataFusion Object Table scans with `StreamingTableExec`:
  - defer recursive listing from planning to execution
  - emit bounded `RecordBatch`es using the session batch size
  - construct only columns requested by the pushed projection
  - avoid global sorting in the scan; explicit SQL `ORDER BY` remains a DataFusion operator
  - stop polling after K files for pushed `LIMIT K` queries without `ORDER BY`
  - do not preallocate memory from an untrusted SQL limit
- Route Object Tables natively from filesystem and REST catalogs.
  - Build filesystem I/O from the Object Table's own path, including cross-scheme locations.
  - Use `GetTableResponse.path` for REST Object Tables, matching Paimon Java and preserving REST data-token handling.
- Add a DataFusion table provider with projection-aware, bounded-memory streaming and limit support.
- Keep Object Tables read-only and reject branches, time travel, and writes.
- Mark the new `LoadedTable` API `#[non_exhaustive]` before its first v0.4 publication so future native table kinds do not repeatedly break downstream exhaustive matches.
- Keep non-native table types on the existing external table-engine path.

## Testing

- [x] `cargo +1.98.0 clippy --locked -p paimon -p paimon-datafusion --all-targets --features fulltext,vortex -- -D warnings`
- [x] `cargo test -p paimon-datafusion --test object_table`
- [x] `cargo test -p paimon-datafusion --test table_type_routing`
- [x] object-table filesystem creation/default-path/legacy-path unit tests
- [x] limited listing stops polling after K files
- [x] a huge limit on an empty location does not preallocate or panic
- [x] ordered `LIMIT` query regression test
- [x] recursive listing yields its first row without draining the namespace
- [x] projected scans emit multiple bounded batches at the configured session batch size
- [x] zero-column projection regression through `COUNT(*)`
- [x] `cargo test -p paimon --test rest_catalog_test test_load_table_constructs_native_object_table -- --exact`
- [x] `cargo check -p paimon -p paimon-datafusion -p paimon-rest-server --all-targets --features fulltext,vortex`
- [x] `git diff --check`

## Notes

- OpenDAL does not expose access time or owner portably, so `atime` is currently `0` and `owner` is `NULL`; modification time and length are populated when available.
- `LoadedTable` was introduced after v0.3.0 and has not appeared in a published release. It is made non-exhaustive here before the v0.4 API is published.
- The Windows REST construction test remains active; only its local recursive-listing assertion is skipped because the existing OpenDAL drive-root `StripPrefixError` is tracked separately.
- This is complementary to #747, which keeps metadata queries available when unsupported external engines are not registered; neither PR depends on the other.

### DIFF
--- a/crates/integrations/datafusion/src/table/object.rs
+++ b/crates/integrations/datafusion/src/table/object.rs
@@ -1,0 +1,153 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use datafusion::arrow::array::{
+    new_null_array, ArrayRef, Int64Array, RecordBatch, StringViewArray,
+};
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::catalog::Session;
+use datafusion::datasource::memory::MemorySourceConfig;
+use datafusion::datasource::{TableProvider, TableType};
+use datafusion::error::{DataFusionError, Result as DFResult};
+use datafusion::logical_expr::dml::InsertOp;
+use datafusion::logical_expr::Expr;
+use datafusion::physical_plan::ExecutionPlan;
+use paimon::table::ObjectTable;
+
+use crate::error::to_datafusion_error;
+
+use super::datafusion_arrow_schema;
+
+/// DataFusion provider for a native read-only Paimon object table.
+#[derive(Debug, Clone)]
+pub(crate) struct ObjectTableProvider {
+    table: ObjectTable,
+    schema: SchemaRef,
+}
+
+impl ObjectTableProvider {
+    pub(crate) fn try_new(table: ObjectTable, schema_force_view_types: bool) -> DFResult<Self> {
+        let schema = datafusion_arrow_schema(&ObjectTable::fields(), schema_force_view_types)?;
+        Ok(Self { table, schema })
+    }
+}
+
+#[async_trait]
+impl TableProvider for ObjectTableProvider {
+    fn schema(&self) -> SchemaRef {
+        Arc::clone(&self.schema)
+    }
+
+    fn table_type(&self) -> TableType {
+        TableType::Base
+    }
+
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        _filters: &[Expr],
+        limit: Option<usize>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        let mut entries = self
+            .table
+            .list_objects()
+            .await
+            .map_err(to_datafusion_error)?;
+        if let Some(limit) = limit {
+            entries.truncate(limit);
+        }
+
+        let paths = entries
+            .iter()
+            .map(|entry| entry.path().to_string())
+            .collect::<Vec<_>>();
+        let names = entries
+            .iter()
+            .map(|entry| entry.name().to_string())
+            .collect::<Vec<_>>();
+        let lengths = entries
+            .iter()
+            .map(|entry| entry.length())
+            .collect::<Vec<_>>();
+        let mtimes = entries
+            .iter()
+            .map(|entry| entry.mtime())
+            .collect::<Vec<_>>();
+        let atimes = entries
+            .iter()
+            .map(|entry| entry.atime())
+            .collect::<Vec<_>>();
+        let owners = entries
+            .iter()
+            .map(|entry| entry.owner())
+            .collect::<Vec<_>>();
+        let row_count = entries.len();
+
+        let string_array = |values: Vec<String>, index: usize| -> ArrayRef {
+            if matches!(
+                self.schema.field(index).data_type(),
+                datafusion::arrow::datatypes::DataType::Utf8View
+            ) {
+                Arc::new(StringViewArray::from(values))
+            } else {
+                Arc::new(datafusion::arrow::array::StringArray::from(values))
+            }
+        };
+        let batch = RecordBatch::try_new(
+            Arc::clone(&self.schema),
+            vec![
+                string_array(paths, 0),
+                string_array(names, 1),
+                Arc::new(Int64Array::from(lengths)),
+                Arc::new(Int64Array::from(mtimes)),
+                Arc::new(Int64Array::from(atimes)),
+                if owners.iter().all(Option::is_none) {
+                    new_null_array(self.schema.field(5).data_type(), row_count)
+                } else if matches!(
+                    self.schema.field(5).data_type(),
+                    datafusion::arrow::datatypes::DataType::Utf8View
+                ) {
+                    Arc::new(StringViewArray::from(owners))
+                } else {
+                    Arc::new(datafusion::arrow::array::StringArray::from(owners))
+                },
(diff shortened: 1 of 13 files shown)


JingsongLi:
- When `path` is not configured, Java defaults to using the table directory; Rust succeeds in creation but throws a `requires a non-empty 'path' option` error upon loading. 
- Java ignores user-defined fields and persists a fixed six-column schema; Rust persists the user-defined schema as-is, yet the reader forces the exposure of a fixed schema, resulting in an inconsistency between the metadata and the query schema.

crates/integrations/datafusion/src/table/object.rs
```diff
@@ -0,0 +1,153 @@
...
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        _filters: &[Expr],
+        limit: Option<usize>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        let mut entries = self
+            .table
+            .list_objects()
```
jerry-024: **[MAJOR]** `LIMIT` is applied only after `list_objects()` has recursively enumerated, materialized, and path-sorted the entire object namespace. Consequently, even `LIMIT 1` retains O(N) listing/memory work and O(N log N) sorting, which can cause high latency or exhaust memory on large object tables.

Please pass the limit into the listing layer and consume the recursive lister incrementally, retaining at most the path-smallest K entries. Keep the full listing only when no limit is supplied.

crates/paimon/src/catalog/mod.rs
```diff
@@ -263,14 +263,16 @@ use async_trait::async_trait;
...
 
 /// Outcome of [`Catalog::load_table`].
 #[derive(Debug)]
 pub enum LoadedTable {
     /// A constructed Paimon table (boxed: far larger than the other variant).
     Paimon(Box<Table>),
-    /// A table this reader cannot construct.
+    /// A native read-only object table.
+    Object(ObjectTable),
+    /// A table that needs a registered external engine.
```
jerry-024, line 276: **[MAJOR]** Adding `Object` to the public, exhaustive `LoadedTable` enum is a breaking API change: downstream crates with exhaustive matches will fail to compile after upgrading, even if they never use object tables.

Please treat and document this as a semver-breaking release. If more table kinds are expected, consider marking the enum `#[non_exhaustive]` in the same breaking release so subsequent variants do not repeatedly break consumers.

shyjsarah:
Addressed in bb4c346. Filesystem object-table creation now follows Java semantics: it persists the fixed six-column schema, defaults a missing/blank `path` to the table directory, preserves an explicit path, and loads legacy metadata without `path` from the table directory.

crates/integrations/datafusion/src/table/object.rs
shyjsarah: Fixed in bb4c346. DataFusion now passes the limit into `ObjectTable`; recursive listing uses OpenDAL `lister_with` incrementally and keeps only the path-smallest K statuses in a bounded max-heap. Limited reads use O(K) memory and O(N log K) selection work.

crates/paimon/src/catalog/mod.rs
shyjsarah: Fixed in bb4c346. `LoadedTable` is now `#[non_exhaustive]`, downstream matches have explicit fallback handling, and the PR description documents the API impact. `LoadedTable` was introduced after v0.3.0, so this is done before its first v0.4 publication.

crates/paimon/src/io/file_io.rs
```diff
@@ -223,23 +225,38 @@ impl FileIO {
...
+        let mut entries =
+            op.lister_with(&list_path)
                 .recursive(true)
                 .await
                 .context(IoUnexpectedSnafu {
                     message: format!("Failed to list files recursively in '{path}'"),
                 })?;
 
         let mut statuses = Vec::new();
+        let mut smallest = limit.map(|limit| BinaryHeap::with_capacity(limit.saturating_add(1)));
```
jerry-024: **[MAJOR]** `BinaryHeap::with_capacity(limit.saturating_add(1))` allocates from the SQL `LIMIT` before the lister yields any object. A query with a very large limit can therefore request gigabytes of memory—or panic on capacity overflow—even when the directory is empty.

Please use `BinaryHeap::new()` and let the heap grow only as matching files are observed. The existing `heap.len() > limit` check already keeps retained entries bounded.

crates/integrations/datafusion/src/table/object.rs
jerry-024: The update bounds retained memory to O(K) and reduces sorting to O(N log K), but the recursive lister is still drained to EOF. As a result, `LIMIT 1` still performs O(N) remote listing work and waits for the entire namespace.

Because SQL `LIMIT` without `ORDER BY` does not require the globally smallest paths, this path can stop after K files and optionally sort only those K results. A global traversal should be reserved for an explicit ordering requirement.

crates/paimon/src/catalog/mod.rs
jerry-024: Retraction after checking the published API history: `LoadedTable` is not present in the released v0.3.0 API and currently exists only on the unreleased 0.4.0 development line. Adding `Object` before the enum’s first release does not break released consumers, and `#[non_exhaustive]` prevents the same issue for future variants.

This is not a compatibility blocker; please disregard my original finding.

crates/paimon/src/io/file_io.rs
shyjsarah: Fixed in fc31528. Limited recursive listing no longer allocates from the SQL LIMIT at all: the bounded heap was removed, entries are appended as the lister yields them, and polling stops once K files have been collected. A regression with `usize::MAX` on an empty location verifies that no capacity allocation or panic occurs.

crates/integrations/datafusion/src/table/object.rs
shyjsarah: Addressed in fc31528. For a pushed `LIMIT K` without `ORDER BY`, the recursive OpenDAL lister now stops after K files instead of draining to EOF; only those K entries are sorted before materialization. Unlimited scans still traverse the full namespace. I also added an instrumented lister regression proving it is not polled past K, plus an `ORDER BY path LIMIT 1` regression to preserve ordered-query correctness.

crates/integrations/datafusion/src/table/object.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        _filters: &[Expr],
+        limit: Option<usize>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        let entries = self
```
jerry-024: **[Major] Avoid fully materializing unbounded object scans**

When `limit` is `None`, this path recursively lists and sorts the entire namespace, then duplicates its metadata into six vectors and one `RecordBatch` before the execution plan can emit any rows. This adds `O(N log N)` setup time and several `O(N)` allocations, can exhaust memory for large object stores, and still constructs columns excluded by the projection. Please stream entries into bounded `RecordBatch`es, construct only projected columns, and avoid a global sort unless ordering is explicitly required.

shyjsarah: Fixed in e50ea4f. Object Table scans now return a projection-aware `StreamingTableExec` instead of listing and materializing during planning. The partition stream lazily consumes OpenDAL, emits batches at the session batch size, builds only projected Arrow columns (including zero-column projections), and does not globally sort scan output. Pushed limits still stop the underlying lister after K files. Added regressions proving the first status is yielded before the next lister poll, projected scans produce bounded batches, and `COUNT(*)` works through an empty projection.

jerry-024 (approved):
+1

PR #751, opened 2026-08-26
fix(spec): widen index-manifest row count to i64

JunRuiLee:
## What

`IndexFileMeta._ROW_COUNT` is an Avro `long` in Java (`IndexFileMeta.SCHEMA` field 3,
`new DataField(3, "_ROW_COUNT", new BigIntType(false))`), but the Rust field is `i32` and the decoder
narrows it:

```rust
"_ROW_COUNT" => row_count = Some(read_long_field(cursor, field.nullable)? as i32),
```

An index file with more than `i32::MAX` rows therefore decodes to a silently wrong (possibly negative)
row count. The index manifest is written by Java and read here, so nothing surfaces the truncation.

Widen the field to `i64` and drop the cast. The substantive change is three lines
(`IndexFileMeta.row_count`, the decoder's local, and the cast); everything else is mechanical
adaptation of construction sites and literals across the crate, the fulltext-gated code, the tests and
the C bindings.

## Assumptions and limits, called out for reviewers

- This is a latent-overflow fix, reachable only above `i32::MAX` rows in a single index file. It is not
  triggered by any existing fixture, so the new test constructs the case directly.
- Widening a public field is a source-compatibility change for downstream code that constructs
  `IndexFileMeta` or reads `row_count` with a typed binding.

## Testing

`row_count_above_i32_max_round_trips_through_index_manifest` fails before the change and passes after.

Gates run locally from a clean HEAD: `cargo fmt --all -- --check`; `cargo clippy -p paimon
--all-targets -- -D warnings` and the same with `--features fulltext`, both clean; `cargo test -p
paimon --lib` 2406 passed / 0 failed and 2480 passed / 0 failed with `--features fulltext`.

## Note

Opened as a draft. This is the first of two independent index-manifest compatibility fixes; the second
one (index file path resolution) is stacked on top of this branch and is sent separately. There is no
semantic dependency between them — they share ~19 files of struct-literal sites, which is why they are
ordered rather than concurrent.

### DIFF
--- a/bindings/c/src/tests.rs
+++ b/bindings/c/src/tests.rs
@@ -2459,7 +2459,7 @@ fn build_pk_vector_table(path: &str, vectors: &[[f32; PK_DIM]]) -> Table {
             index_type: INDEX_TYPE.to_string(),
             file_name: index_file_name,
             file_size: i64::try_from(index_file_size).unwrap(),
-            row_count: i32::try_from(row_count).unwrap(),
+            row_count,
             deletion_vectors_ranges: None,
             global_index_meta: Some(GlobalIndexMeta {
                 row_range_start: 0,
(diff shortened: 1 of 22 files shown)


JingsongLi (approved):
+1

PR #752, opened 2026-08-26
fix(table): resolve index files by external path and bucket layout

JunRuiLee:
## What

Index files are always read from `<table>/index/<file>`, which ignores two things Java records and
honors:

- `_EXTERNAL_PATH` — `IndexFileMeta.SCHEMA` field 5. Not decoded here at all, so it is also dropped
  when this crate rewrites an index manifest.
- `index-file-in-data-file-dir` — when set, Java's `indexFileFactory(partition, bucket)` resolves an
  index file against the bucket's data-file directory instead of the table `index/` directory.

Either one makes a table's index files unreadable. Observed as a primary-key vector search:
`failed to open ANN index file '.../<table>/index/index-<uuid>-0' for range reads`, while the file is
in the bucket directory.

## How

Decode `_EXTERNAL_PATH` and add it to the write schema, add the `index-file-in-data-file-dir` option,
and route every index-file consumer through one resolver (`table/index_file_path.rs`) with two modes,
each mirroring the factory Java uses for that consumer:

| mode | consumers | Java |
| --- | --- | --- |
| global — always `<table>/index` | data-evolution global index, vector and full-text search over it | `globalIndexFileFactory()` |
| bucket-local — data-file directory when the option is set | primary-key vector ANN segments, primary-key full-text archives, deletion vectors, dynamic-bucket hash index | `pathFactories.get(partition, bucket)` |

An explicit external path wins over both layouts, as in `toPath(IndexFileMeta)`.

For the two index kinds this crate writes itself, **reads and writes move together**, so a file written
here is found again: the data-evolution writer resolves an existing deletion vector through the same
path when merging and writes a new one where the reader will look; the dynamic-bucket assigner resolves
per partition and bucket for both restore and commit (`prepare_commit_index` no longer takes an index
directory — three of its four implementations ignored it); and `TableCommit::abort` deletes where the
file was written, mirroring `FileStoreCommitImpl.abort`. Deleting is best-effort, so the old fixed path
leaked silently — and `try_commit` calls `abort` itself on failure, as do the C and Python bindings.

Bucket directories now come from one `spec::bucket_path` (Java `FileStorePathFactory.bucketPath`),
replacing four copies of the same expression: the layout is only correct while every producer and
consumer agrees byte for byte, and nothing else enforces that.

Two smaller consequences of making the option live:

- **It is now immutable**, as in Java (`@Immutable`, rejected by `SchemaManager.checkAlterTableOption`).
  It selects where index files are written, so flipping it on a populated table would hide every index
  file the table has. It was inert here before, so altering it used to be a harmless no-op.
- **`$physical_files_size`** counts an `index-` prefixed file in a bucket directory as an index file
  instead of dropping it, matching Java `FileType.classify`. Classification follows physical form, not
  the current option value. Two tests that asserted the old counts are corrected.

The BTree reader cache is keyed by the resolved path — with the bare file name, two entries sharing a
name but resolving to different locations would reuse each other's reader.

## Out of scope

- **`data-file.path-directory`**: unsupported crate-wide (bucket paths are rooted at the table for data
  files as much as index files), so honoring it belongs with data-file path handling.
- **Writing** bucket-local index files to a configured external data path: this crate does not write
  data files there either. Reading a Java-written external index file works.
- **Referenced-file identity**: index files are keyed by bare file name, as Java does in the
  higher-stakes direction (`LocalOrphanFilesClean` maps candidate *deletions* by `Path.getName()`), and
  all names are UUID-based, so collisions are not reachable.

## Behavior change

A table created with the option set but written **only by older paimon-rust** has its index files under
`<table>/index`, because this crate ignored the option — such a table is already unreadable by Java.
After this PR Rust agrees with Java. No compatibility fallback on purpose: one would diverge from Java
and mask genuine missing-file errors.

## Testing

New: `abort` deletes an index file in the bucket data-file directory, and one at an external path (both
verified to fail against the old path); ANN segment and full-text archive resolution with the option
set, through a real split whose bucket path is not derivable from the table root — the failure this PR
fixes, previously untested through a split; `directory()` agrees with `resolve()` in every mode; decode
tests for `_EXTERNAL_PATH` present, present-but-null, and absent from the writer schema; the option
cannot be altered.

Gates: `cargo fmt --all -- --check` clean; `clippy -D warnings --all-targets` clean for `paimon`,
`paimon --features fulltext`, `paimon-datafusion`, `paimon-c` (`paimon-python` needs Python ≥ 3.10,
unavailable locally); `cargo test -p paimon --lib` 2425 passed, 2500 with `--features fulltext`,
`--tests` all pass. `paimon-datafusion` 339 passed / 7 failed — all seven are `TableNotExist` for the
`/tmp/paimon-warehouse` fixtures that `make docker-up` provisions, unrelated to this change.

### DIFF
--- a/crates/paimon/src/catalog/filesystem.rs
+++ b/crates/paimon/src/catalog/filesystem.rs
@@ -30,7 +30,7 @@ use crate::spec::{
     CoreOptions, Schema, TableSchema, TableType, INDEX_FILE_IN_DATA_FILE_DIR_OPTION,
     TABLE_TYPE_OPTION,
 };
-use crate::table::{ObjectTable, SchemaManager, Table};
+use crate::table::{ObjectTable, SchemaManager, SnapshotManager, Table};
 use async_trait::async_trait;
 use bytes::Bytes;
 use opendal::raw::get_basename;
@@ -543,7 +543,26 @@ impl Catalog for FileSystemCatalog {
                 full_name: identifier.full_name(),
             })?;
 
-        reject_immutable_option_changes(current.options(), &changes)?;
+        // Only the index-layout option is snapshot-gated, so the listing a
+        // snapshot lookup costs is paid only when a change touches it. Java defers
+        // the same lookup through a `LazyField`.
+        let touches_snapshot_gated_option = changes.iter().any(|change| {
+            matches!(
+                change,
+                crate::spec::SchemaChange::SetOption { key, .. }
+                | crate::spec::SchemaChange::RemoveOption { key }
+                    if key == INDEX_FILE_IN_DATA_FILE_DIR_OPTION
+            )
+        });
+        let has_snapshots = if touches_snapshot_gated_option {
+            SnapshotManager::new(self.file_io.clone(), table_path.clone())
+                .get_latest_snapshot_id()
+                .await?
+                .is_some()
+        } else {
+            false
+        };
+        reject_immutable_option_changes(current.options(), &changes, has_snapshots)?;
 
         let new_schema = current
             .apply_changes(changes)
@@ -560,13 +579,31 @@ impl Catalog for FileSystemCatalog {
 /// this mirrors the subset this crate acts on:
 ///
 /// * `type` picks the reader, so flipping it strands a populated table behind a
-///   reader that cannot see its data. Only case-insensitive no-ops pass.
+///   reader that cannot see its data. Only case-insensitive no-ops pass, and a
+///   snapshot is not required: a format table holds data without ever writing
+///   one, so a snapshot check would not catch it. Java special-cases `type` the
+///   same way (`SchemaManager.generateTableSchema`).
 /// * `index-file-in-data-file-dir` picks the directory every bucket-local index
 ///   file is written to and read from, so flipping it hides every index file the
-///   table already has.
+///   table already has. Only files already written are at stake, so this follows
+///   Java in rejecting an actual change only once the table has snapshots, and in
+///   letting a `SetOption` that repeats the stored value through: Java compares
+///   the stored string literally, and reaches `checkAlterTableOption` only under
+///   `hasSnapshots && !unchanged`.
+///
+/// Gating on snapshot existence leaves one window open, in Java as much as here: a
+/// writer that started before the alter still holds the old layout, and a commit
+/// records the schema id it was built with without checking whether the latest
+/// schema moved, so it can publish the first snapshot after the layout was flipped
+/// under it. Closing it needs schema publication and snapshot commit to be ordered
+/// against each other — reading the latest schema once before committing is not
+/// enough, since the alter can land right after that read. A stricter alter is not
+/// the answer either: rejecting the change outright would take away the only
+/// post-creation point at which the layout can be chosen.
 fn reject_immutable_option_changes(
     current_options: &HashMap<String, String>,
     changes: &[crate::spec::SchemaChange],
+    has_snapshots: bool,
 ) -> Result<()> {
     let current_type = current_options
         .get(TABLE_TYPE_OPTION)
@@ -588,15 +625,31 @@ fn reject_immutable_option_changes(
                     message: format!("removing '{TABLE_TYPE_OPTION}' is not supported"),
                 });
             }
-            crate::spec::SchemaChange::SetOption { key, .. }
-            | crate::spec::SchemaChange::RemoveOption { key }
-                if key == INDEX_FILE_IN_DATA_FILE_DIR_OPTION =>
+            crate::spec::SchemaChange::SetOption { key, value }
+                if key == INDEX_FILE_IN_DATA_FILE_DIR_OPTION
+                    && has_snapshots
+                    && current_options.get(key.as_str()) != Some(value) =>
             {
                 return Err(Error::Unsupported {
                     message: format!(
-                        "changing '{INDEX_FILE_IN_DATA_FILE_DIR_OPTION}' is not supported: \
+                        "changing '{INDEX_FILE_IN_DATA_FILE_DIR_OPTION}' on a table with \
+                         snapshots is not supported: \
                          it selects the directory index files are written to, so the files \
                          already written would no longer be found"
+                    ),
+                });
+            }
+            crate::spec::SchemaChange::RemoveOption { key }
+                if key == INDEX_FILE_IN_DATA_FILE_DIR_OPTION && has_snapshots =>
+            {
+                // Java rejects the reset whenever the table has snapshots, whether
+                // or not the option is currently set
+                // (`SchemaManager.checkResetTableOption`).
+                return Err(Error::Unsupported {
+                    message: format!(
+                        "removing '{INDEX_FILE_IN_DATA_FILE_DIR_OPTION}' from a table with \
+                         snapshots is not supported: it selects the directory index files \
+                         are written to, so the files already written would no longer be found"
                     ),
                 });
             }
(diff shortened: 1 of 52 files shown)


JingsongLi:
- The `abort()` function in `table_commit.rs:728-735` deletes all indexes lacking an external path as if they were `BucketLocal`. 
- When `index-file-in-data-file-dir=true` is enabled, the global index is actually written to `<table>/index`, but the abort process attempts to delete it from the bucket directory; this causes a failed commit to leave behind orphan files.

JunRuiLee:
Confirmed, and reachable: the vindex index builder aborts messages holding global index files, and `try_commit` aborts whatever it was committing.

Fixed in 88f1df1. `abort` now classifies each index file rather than resolving the whole message as bucket-local:

- no `_GLOBAL_INDEX` → bucket-local (deletion vectors, the dynamic-bucket hash index);
- a `_GLOBAL_INDEX` whose `_SOURCE_META` is absent — what this crate's index builders write — or whose `_SOURCE_META` carries `DataEvolutionIndexSourceMeta`'s `DEIX` marker → global;
- anything else → bucket-local, which is what Java assumes for every index file.

Both abort tests now also plant a same-named file in the other layout and assert it survives, so neither direction can regress into deleting a path the commit does not own.

One thing I ran into while checking this against Java: `FileStoreCommitImpl.abort` resolves every `newIndexFiles()` entry through `IndexFilePathFactories.get(partition, bucket)`, while `SortedGlobalIndexWriter.flushIndex` writes through `globalIndexFileFactory()` and returns those metas in `DataIncrement.indexIncrement(...)`. So with `index-file-in-data-file-dir` set, Java leaves the same orphan behind.

crates/paimon/src/catalog/filesystem.rs
```diff
@@ -575,6 +587,18 @@ fn reject_table_type_changes(
                     message: format!("removing '{TABLE_TYPE_OPTION}' is not supported"),
                 });
             }
+            crate::spec::SchemaChange::SetOption { key, .. }
+            | crate::spec::SchemaChange::RemoveOption { key }
+                if key == INDEX_FILE_IN_DATA_FILE_DIR_OPTION =>
```
JingsongLi: [P1] Keep the index layout immutable for dynamic table copies too

This guard only covers catalog `ALTER TABLE`, but the public `Table::copy_with_options` path still merges `index-file-in-data-file-dir` through `TableSchema::copy_with_options`. Every new resolver then reads that overridden value. For example, a persisted `true` table copied with `false` looks for its bucket-local deletion/vector/full-text indexes under `<table>/index`; the inverse misses indexes stored there. More seriously, a write/commit built from such a copy can place new hash/DV index files according to the temporary value while committing name-only manifest entries to a table whose persisted schema retains the opposite value, so a normally reloaded table cannot resolve those files and abort cleanup can use the wrong layout. Please pin this option to the stored schema value (including the default `false`) or reject dynamic overrides, as is already done for other safety-sensitive options, and cover both read and write copies in a regression test.

crates/paimon/src/catalog/filesystem.rs
```diff
@@ -575,6 +587,18 @@ fn reject_table_type_changes(
...
+            {
+                return Err(Error::Unsupported {
+                    message: format!(
+                        "changing '{INDEX_FILE_IN_DATA_FILE_DIR_OPTION}' is not supported: \
```
JingsongLi: [P2] Preserve Java’s snapshot-aware/no-op immutable-option behavior

This arm rejects every set/remove of `index-file-in-data-file-dir`, even before the table has its first snapshot and even when a `SetOption` repeats the stored value. Upstream Java `SchemaManager` first computes `unchanged` and only calls `checkAlterTableOption` when `hasSnapshots && !unchanged`; this lets callers choose the layout through ALTER before the first write and makes schema reconciliation/idempotent SETs succeed. The new Rust test instead cements unconditional failure on an empty table. Please allow an exact no-op, allow an actual change while there are no snapshots, and reject only an actual change once snapshots exist; add coverage for all three cases.

crates/paimon/src/catalog/filesystem.rs
JunRuiLee: Fixed in 3df9e15. `TableSchema::copy_with_options` now pins the option to the stored value, and drops the override entirely when nothing is stored so the default stands — the shape it already uses for `type`. Java rejects such an override outright (`AbstractFileStoreTable.checkImmutability`); this copy is infallible, so pinning is the closest equivalent.

Tests: a copied-with-`false` table still writes its hash index into the bucket directory, a read through `Table::copy_with_options` still reports the stored layout, and the pin in both directions. `copy_with_replaced_options` and `Table::new` stay unpinned — they take options wholesale rather than as an override, as `type` has always been.

One correction: a writer and its committer come from the same `WriteBuilder`, so abort resolves the layout the write used. The damage that needs no such crossing is the successful commit itself.

crates/paimon/src/catalog/filesystem.rs
JunRuiLee: Fixed in 3df9e15. The guard now takes `has_snapshots` and rejects only an actual change once snapshots exist; `RemoveOption` only with snapshots, set or not. `type` stays unconditional, as in Java — a format table holds data without writing a snapshot. `alter_table` resolves the flag only when a change touches the option.

One case that looks like a no-op and is not: `false` where the option was never stored. Java compares the stored value and normalizes only `type`, `primary-key` and `partition`, so it counts as a change and is rejected once snapshots exist. Tested, along with the four you listed.

Gating on snapshots leaves a window, now in the guard's docs: a writer that started before the alter still holds the old layout, and a commit records its schema id without checking whether the latest schema moved. Java's cached lazy check has the same one; closing it belongs on the commit path.

crates/paimon/src/catalog/filesystem.rs
```diff
@@ -539,7 +542,26 @@ impl Catalog for FileSystemCatalog {
...
+        });
+        let has_snapshots = if touches_snapshot_gated_option {
+            SnapshotManager::new(self.file_io.clone(), table_path.clone())
+                .get_latest_snapshot_id()
+                .await?
+                .is_some()
+        } else {
+            false
+        };
+        reject_immutable_option_changes(current.options(), &changes, has_snapshots)?;
```
JingsongLi: [P1] Serialize this layout ALTER with first-snapshot commits

This check and `save_table_schema` are independent of snapshot publication. A writer created before the ALTER has already cached the old `index-file-in-data-file-dir` layout, so it can write HASH/DV files to the old directory and then publish the first snapshot after this ALTER succeeds. A freshly loaded table resolves the manifest file names using the new layout, making those index files unreachable (and dynamic-bucket reads/writes can fail with file-not-found). The comment below explicitly documents this interleaving, but the new tests cover only serial ALTERs. Please serialize schema publication and snapshot commit through the same lock/CAS, or keep this option immutable after creation until that ordering exists; add a stale-writer -> ALTER -> commit -> fresh-table restore regression.

JunRuiLee: Took the second option: the layout is chosen at creation and not afterwards, in 5d39ab8. Any actual change or removal is rejected whether or not snapshots exist. What still passes is what moves nothing — a `SetOption` repeating the stored value, and a `RemoveOption` for an option that is not set — so reconciliation stays idempotent. That makes Rust stricter than Java on an option Java itself marks `@Immutable`; the remove-of-an-unset-option case is the one place it is looser, and it cannot move a file.

Not the lock/CAS: a commit stamps the schema id its writer was built with and never checks whether the latest schema moved, so ordering schema publication against snapshot commit would be a commit-path change for every option, not an index-path one. Worth doing, but not here.

The regression is the sequence you described: a writer resolves the layout, both conflicting alters are refused, the writer commits, and a freshly loaded table finds the committed hash index in the bucket data-file directory. Values are compared literally, by the way — `" TRUE "` parses as `false`, so treating it as equal to `true` would let a real change through.

Also rebased onto main. #767 split the global index builders and the scanner, so the routing moved into `global_index_scanner/{entry,reader,deletion_vectors}.rs`, and #766's FM global index reader had the same hardcoded `<table>/index` path — it is routed now too.

JingsongLi (approved):
+1

PR #753, opened 2026-08-26
  fix: resolve time-travel clauses instead of answering with current rows

plusplusjiajia:
### Purpose

  `PaimonRelationPlanner` handed back every version clause it did not resolve, and DataFusion drops it. `FOR SYSTEM_TIME AS OF` on a Paimon table therefore planned against the current snapshot, so an explicitly historical query answered with today's rows.

  Reachable from a plain `SessionContext` with a dialect that parses version clauses. `SQLContext` resolves them before planning and is unaffected.

  ### Brief change log

  - Decide ownership by provider type before reading the clause. A provider this planner does not own keeps its clause untouched — relation planners are session-global, so parsing it here would take it from the engine that registered it.
  - Paimon's own object and system tables hold no snapshot to rewrite onto, so they refuse the clause instead of handing it to the default planner, which drops it. `system_tables::is_system_table_provider` answers by type rather than by the `base$name` spelling, since
  another engine may name a table with a `$`.
  - `FOR SYSTEM_TIME AS OF` carries the same timestamp expression as `TIMESTAMP AS OF`, so the two now share a resolver. `AT(...)` and `CHANGES(...)` name neither a snapshot nor a timestamp and are refused.
  - The read path classifies through `load_table`, so a table a query engine must serve reports its declared type at load time rather than deeper in the scan.

### DIFF
--- a/crates/integrations/datafusion/src/relation_planner.rs
+++ b/crates/integrations/datafusion/src/relation_planner.rs
@@ -33,7 +33,7 @@ use datafusion::sql::sqlparser::ast::{self, TableFactor, TableVersion};
 use paimon::spec::{SCAN_TIMESTAMP_MILLIS_OPTION, SCAN_VERSION_OPTION};
 
 use crate::catalog::ReadOnlyTableProvider;
-use crate::table::PaimonTableProvider;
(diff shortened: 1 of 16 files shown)


JingsongLi (approved):
+1

PR #754, opened 2026-08-27
feat(file_index): add Java-compatible Bitmap reader

QuakeWang:
### Purpose

Related issue: #34

The Rust file-index layer had predicate composition and row-level selection support, but no Bitmap reader, so it could not consume Bitmap indexes written by Java Paimon. Bitmap V2 also requires block-level lookup to avoid eagerly materializing high-cardinality dictionaries.

### Brief change log

- Decode Java-compatible Bitmap V1 and V2 metadata and value encodings.
- Support Boolean, integer, floating-point, date/time, timestamp, and string types.
- Preserve Java floating-point ordering, NaN canonicalization, and conservative signed-zero predicate semantics.
- Deserialize singleton and RoaringBitmap32 entries on demand.
- Keep V2 secondary-index blocks lazy and parse only blocks selected by predicate literals.
- Return row selections for Eq, In, NotEq, NotIn, IsNull, and IsNotNull.
- Add Java V1/V2 golden fixtures, malformed-input tests, and a high-cardinality multi-block regression test.

### Tests

- Java release-2.0 Bitmap writer/reader fixture probe

### API and Format

No public API or writer-format changes. This adds read support for existing Java Bitmap V1/V2 payloads.

### Documentation

No documentation changes.

### DIFF
--- a/crates/paimon/src/file_index/bitmap/mod.rs
+++ b/crates/paimon/src/file_index/bitmap/mod.rs
@@ -1,0 +1,1789 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #760, opened 2026-08-28
vector-search: support scalar pre-filter before Top-K

shyjsarah:
## Summary

Support scalar pre-filtering before Top-K for data-evolution/global-index vector search, including DataFusion literal and lateral `vector_search` queries.

The scalar predicate is evaluated against the same pinned snapshot as vector search, producing global row IDs that are localized per vector-index shard and passed to the vector backend as an allow-list. A normal Paimon read is used so scalar global indexes such as BTree can prune the filter read while residual evaluation, partial index coverage, and deletion visibility remain exact.

## Changes

- Enable `VectorSearchBuilder::with_filter` and `BatchVectorSearchBuilder::with_filter` on the data-evolution path.
- Pin scalar filtering, deletion-vector handling, raw fallback, refine, and materialization to one resolved snapshot.
- Apply the same global row-ID allow-list to indexed search and unindexed raw fallback.
- Localize a shared bitmap once per vector-index shard, skip empty shards, and share bitmaps across batch queries without deep copies.
- Add a reusable prepared-filter API so lateral execution evaluates a target-side scalar predicate once per physical plan.
- Push fully translated DataFusion predicates into literal `vector_search` while retaining an `Inexact` residual correctness check.
- Push fully translated target-only conjuncts into lateral vector search while preserving left-side, cross-side, unsupported, and inexact residual predicates.
- Reject NaN predicates from vector pre-filter pushdown because Paimon and Arrow comparison semantics differ.
- Show pushed scalar filters in logical and physical `EXPLAIN` output.
- Document scalar pre-filter semantics and lateral-query behavior.

## Testing

- [x] `cargo fmt --all --check`
- [x] `git diff --check`
- [x] `cargo test -p paimon --lib table::vector_search_builder::tests`
- [x] `cargo test -p paimon-datafusion --lib filter_pushdown::tests::test_vector_prefilter_rejects_nan_literals`
- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vindex_build_then_vector_search_query -- --nocapture`
- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_lateral_join_uses_query_vectors -- --nocapture`
- [x] `cargo clippy -p paimon --lib --tests -- -D warnings`
- [x] `cargo clippy -p paimon-datafusion --lib --tests -- -D warnings`

Covered cases include filter-before-Top-K ordering, empty matches, batch reuse, non-zero shard offsets, raw fallback for unindexed rows, time-travel snapshot pinning, literal SQL, lateral SQL, mixed/cross-side conjuncts, and NaN pushdown safety.

## Notes

- No on-disk format change.
- No SQL function signature change.
- Existing `VectorSearch::include_row_ids` public API remains unchanged; this PR adds `PreparedVectorSearchFilter` and shared-filter builder support.
- This is scalar pre-filtering for vector search, not the existing vector/full-text route-fusion `hybrid_search` feature.


jerry-024:
New: 2; still open: 0; fixed candidates: 0.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -398,17 +505,35 @@ impl LateralVectorSearchExec {
...
+                let prepared = self
+                    .prepared_filter
+                    .get_or_try_init(|| {
+                        self.target_table
+                            .prepare_vector_search_filter(filter.clone())
+                    })
+                    .await
+                    .map_err(to_datafusion_error)?;
+                (
+                    prepared.table(),
```
jerry-024: **[MAJOR]** The prepared scalar filter is cached for the lifetime of the physical execution plan, so re-executing the same plan after the target table advances keeps querying the snapshot selected by the first execution.

Scope the `OnceCell` to one query execution rather than storing it permanently on the reusable `ExecutionPlan`; share only an execution-scoped cache across partitions and batches.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1634,6 +1758,43 @@ async fn evaluate_batch_vector_search(
     let index_search_limit = indexed_search_limit(max_limit, refine_factor)?;
 
     let vector_entry_count = vector_entries.len();
+    let shared_include_row_ids =
+        vector_searches[0]
+            .effective_include_row_ids()
+            .filter(|include_row_ids| {
+                vector_searches
+                    .iter()
+                    .all(|search| search.effective_include_row_ids() == Some(*include_row_ids))
```
jerry-024: **[MAJOR]** Detecting the shared pre-filter compares the entire `RoaringTreemap` once for every query in the batch, even though `BatchVectorSearchBuilder` installs the same `Arc` on every query.

Inspect `shared_include_row_ids` directly and use `Arc::ptr_eq` against the first query Arc; fall back to per-query localization if any query does not share that `Arc`.

crates/integrations/datafusion/src/lateral_vector_search.rs
shyjsarah: Fixed in 3b69116. The plan-lifetime OnceCell is replaced by an execution-scoped weak cache: partitions and batches using the same live TaskContext share one prepared filter, while dropping the execution streams releases it so re-executing the physical plan prepares against a fresh snapshot. Added a regression test that reuses the same physical plan after appending and indexing a new target row.

crates/paimon/src/table/vector_search_builder.rs
shyjsarah: Fixed in dfd45ec. Shared-filter detection now inspects shared_include_row_ids and requires Arc::ptr_eq with the first query Arc. Owned bitmaps, equal-but-distinct Arcs, and mixed inputs fall back to per-query localization. Added unit coverage for all three cases.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -352,9 +450,57 @@ struct LateralVectorSearchExec {
...
+                    return prepared_filter;
+                }
+            }
+        }
+
+        let prepared_filter = Arc::new(OnceCell::new());
+        entries.push(ExecutionPreparedFilterEntry {
+            context: Arc::downgrade(context),
+            prepared_filter: Arc::downgrade(&prepared_filter),
+        });
```
jerry-024, line 563: **[MAJOR]** The execution-scoped prefilter cache only retains a weak reference to the prepared filter, so one logical execution can resolve different target snapshots across partitions. Once an earlier partition stream is exhausted and drops the last strong reference, a later partition using the same live `TaskContext` creates a new `OnceCell` and resolves the table snapshot again.

Keep the prepared filter strongly owned for the complete execution across all partitions, releasing it only when that execution finishes. Please add a sequential-partition test that updates the target between partitions.

crates/paimon/src/lumina/reader.rs
```diff
@@ -369,7 +369,7 @@ fn search_lumina_batch(
     }
     if vector_searches
         .iter()
-        .any(|vector_search| vector_search.include_row_ids.is_some())
+        .any(|vector_search| vector_search.effective_include_row_ids().is_some())
     {
```
jerry-024: **[MAJOR]** Filtered batch searches discard Lumina batch execution and rebuild the same shared allow-list once per query. Each fallback call materializes the shared `RoaringTreemap` into a new `Vec<u64>`, producing `O(Q * B)` conversion/allocation work plus Q sequential searches per shard.

When every query shares the same bitmap, materialize it once and call `search_with_filter` with the full query batch. Keep the per-query fallback only for differing filters.

crates/integrations/datafusion/src/lateral_vector_search.rs
shyjsarah: Fixed in 339321c. The execution cache now strongly owns the prepared filter while the TaskContext is alive, so sequential partitions share the same resolved snapshot. I also updated the regression test to execute two partitions sequentially, mutate/reindex the target between them, verify the second partition still sees the original snapshot, and verify a new TaskContext sees the refreshed snapshot.

crates/paimon/src/lumina/reader.rs
shyjsarah: Fixed in 66516d1. Lumina now detects a shared filter by Arc identity, materializes the bitmap once, and calls search_with_filter once for the full query batch. Differing/mixed filters and differing limits retain the scalar fallback; empty shared filters return without native calls. Added mock-backed tests for the batch call and fallback cases.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -352,9 +450,54 @@ struct LateralVectorSearchExec {
...
     output_schema: ArrowSchemaRef,
+    filter: Option<Predicate>,
+    prepared_filter_cache: Arc<ExecutionPreparedFilterCache>,
     plan_properties: Arc<PlanProperties>,
 }
 
+#[derive(Debug)]
+struct ExecutionPreparedFilterEntry {
+    context: Weak<TaskContext>,
+    prepared_filter: Arc<OnceCell<PreparedVectorSearchFilter>>,
```
jerry-024, line 465: [P1] Release prepared filters when their execution ends

This cache entry owns the OnceCell strongly, but the dead TaskContext only makes context non-upgradeable; the entry is removed only if a later execution calls for_execution. After the final use of a reusable physical plan, the last scalar-filter bitmap (and its pinned table state) therefore remains reachable for the plan's lifetime, which can retain a very large RoaringTreemap. Please tie the value to the TaskContext lifecycle or evict this exact entry when that execution completes.

shyjsarah: Fixed in f0e2b08. Each partition stream now owns a completion lease for the execution-scoped prepared filter. Sequential/concurrent partitions continue sharing the same OnceCell, and dropping the final lease for every partition removes that exact cache entry immediately, even if the TaskContext and reusable physical plan remain alive. The lease is also released on stream cancellation/error. Added tests for immediate release, exact-entry eviction across concurrent executions, and cloned in-flight leases; the sequential-partition snapshot regression remains covered.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -352,9 +450,141 @@ struct LateralVectorSearchExec {
...
+        }
+
+        let prepared_filter = Arc::new(OnceCell::new());
+        let mut active_partition_leases = HashMap::new();
+        active_partition_leases.insert(partition, 1);
+        entries.push(ExecutionPreparedFilterEntry {
+            context: Arc::downgrade(context),
+            prepared_filter: Arc::clone(&prepared_filter),
+            partition_count,
+            unfinished_partitions: (0..partition_count).collect(),
```
jerry-024: **[MAJOR]** Do not require every declared partition to start before cleanup

This pre-populates completion state for every declared partition, but an entry is cleared only when each partition has acquired and dropped a lease. A consumer may execute only a subset of partitions, or stop early, so unstarted indexes can never call finish_partition; the prepared bitmap and pinned table then remain retained for the physical plan's lifetime. Please tie eviction to execution-level completion or cancellation instead of requiring every declared partition to start.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1194,14 +1216,32 @@ impl<'a> BatchVectorSearchBuilder<'a> {
...
+    /// The builder's table must be [`PreparedVectorSearchFilter::table`] (or an
+    /// equivalent copy pinned to the same snapshot).
+    pub fn with_include_row_ids(&mut self, include_row_ids: RoaringTreemap) -> &mut Self {
+        self.include_row_ids = Some(Arc::new(include_row_ids));
+        self.filter = None;
+        self
+    }
+
+    /// Reuse a shared row-ID allow-list without copying its bitmap.
+    pub fn with_shared_include_row_ids(
```
JingsongLi: [P1] Please keep the prepared allow-list and its pinned table snapshot atomic in the public API. Today a caller can prepare at S1, let the live table advance to S2, then create this builder from the original table and pass `Arc::clone(prepared.include_row_ids())` here. `execute` independently resolves S2, so the S1 bitmap silently excludes newly matching/updated S2 rows and returns an incorrect Top-K. The internal DataFusion callers pair `prepared.table()` correctly, but this zero-copy public setter has no way to enforce that contract. Could the builder accept `PreparedVectorSearchFilter` itself and bind its table, or carry/validate the snapshot ID with the bitmap (leaving raw bitmap setters as an explicitly lower-level API)?

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -3046,7 +3313,7 @@ impl RawScoringPlan {
             .collect();
 
         for (query_index, vector_search) in vector_searches.iter().enumerate() {
```
JingsongLi: [P1] Keep shared pre-filters compact in the raw fallback

For a lateral batch, every query can intentionally share the same `Arc<Bitmap32>`, but this loop expands that bitmap into one stored query index per `(row, query)` pair. A filter with 1M included rows and 1K queries therefore retains roughly 1 billion `usize` entries (about 8 GiB before the per-row `Vec` overhead) before scoring starts. The raw path also reads the full unindexed/detail ranges before consulting this structure, so a selective pre-filter does not reduce raw I/O.

Please preserve the shared case as one bitmap plus a query-index group (and intersect the raw ranges with the shared/union filter before the Paimon read) instead of materializing `O(Q×B)` associations. A regression test with many queries sharing one sparse bitmap should assert both bounded planning memory and raw-row pruning.

crates/paimon/src/vindex/reader.rs
```diff
@@ -472,7 +472,7 @@ fn prepare_search(
         ),
     };
 
-    let filter_bytes = if let Some(include_ids) = &vector_search.include_row_ids {
+    let filter_bytes = if let Some(include_ids) = vector_search.effective_include_row_ids() {
```
JingsongLi, line 498: [P2] Serialize a shared filter only once per batch

The lateral path deliberately installs the same `Arc<Bitmap32>` on every query, but `prepare_search` serializes that bitmap separately for each query. The grouping below then compares the resulting `Vec<u8>` values byte-for-byte, so a batch of Q queries with a B-byte filter pays `O(Q×B)` allocation, serialization, and comparison even though the filter is shared.

Please detect the shared `Arc` before preparing the searches (or cache the serialized bytes by Arc identity) and reuse one serialized payload/group. A batch test with a large shared bitmap would make the intended single-serialization behavior explicit.

crates/integrations/datafusion/src/lateral_vector_search.rs
shyjsarah: Fixed in `14a22f0`. The cache now tracks only partitions that actually acquire stream leases, keeps the `TaskContext` with each lease, and wraps the child stream so cancellation/end/error drops the inner stream before releasing the cache lease. Eviction occurs when no streams remain and either the execution context is otherwise gone or all declared partitions completed. Added regressions for subset execution cleanup and cancellation drop ordering; the sequential-partition snapshot reuse integration test remains green.

crates/paimon/src/table/vector_search_builder.rs
shyjsarah: Fixed in `14a22f0`. `BatchVectorSearchBuilder::with_prepared_filter` now accepts the complete `PreparedVectorSearchFilter` and executes against its pinned table/snapshot atomically. DataFusion callers were migrated to this API; the raw bitmap setter remains documented as a lower-level, snapshot-unbound API.

crates/paimon/src/table/vector_search_builder.rs
shyjsarah: Fixed in `14a22f0`. The raw scoring plan preserves a shared filter as one `Arc<RoaringTreemap>` plus one query-index vector, avoiding `O(Q×B)` row/query associations. Before planning the raw read, filtered queries now intersect unindexed/detail ranges with the shared filter (or the union for distinct filters). Added regressions for bounded planning state and sparse range pruning.

crates/paimon/src/vindex/reader.rs
shyjsarah: Fixed in `14a22f0`. Batch preparation detects a shared bitmap by `Arc` identity, serializes it once into one `Arc<[u8]>`, reuses that allocation for every prepared query, and groups native batch searches by params plus serialized-filter identity rather than byte-vector comparison. Added a 128-query/100k-row regression asserting all prepared searches share the same serialized allocation.

crates/paimon/src/vindex/reader.rs
```diff
@@ -421,16 +421,37 @@ fn search_vindex(
...
     params: VectorSearchParams,
-    filter_bytes: Option<Vec<u8>>,
+    filter_bytes: Option<Arc<[u8]>>,
 }
 
+impl PreparedSearch {
+    fn same_batch_group(&self, other: &Self) -> bool {
+        self.params == other.params
+            && match (&self.filter_bytes, &other.filter_bytes) {
+                (Some(left), Some(right)) => Arc::ptr_eq(left, right),
```
JingsongLi: [P2] Keep batching equal filters supplied through the public API

`Arc::ptr_eq` only groups the new crate-internal shared-filter path. External callers of `VindexVectorGlobalIndexReader::visit_batch_vector_search` can only build filters with public `VectorSearch::with_include_row_ids`; two equal filters (or clones made through that API) are serialized into distinct `Arc<[u8]>` values, so this comparison puts every query in a singleton group. Those groups take `execute_scalar_search`, repeating native setup and probed-list I/O Q times. Before this commit, `PreparedSearch` equality compared the serialized bytes and these queries used the filtered native batch. The updated `filtered_batch_matches_scalar_searches_and_reuses_probed_lists` test switches from two owned public filters to the crate-private shared setter, so it no longer covers that public path. Please retain a pointer fast path but fall back to content fingerprint/equality (or expose a public shared-filter API), and restore coverage using separate `with_include_row_ids` filters while asserting native batching/I/O reuse.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1221,7 +1268,12 @@ impl<'a> BatchVectorSearchBuilder<'a> {
         // returns data-derived row ids/scores outside `TableScan`/`TableRead`,
         // so it must refuse a `query-auth.enabled` table before any fast path
         // (an empty snapshot would otherwise return empty results and bypass it).
-        let core = CoreOptions::new(self.table.schema().options());
+        let execution_table = self
+            .prepared_filter
+            .as_ref()
+            .map(PreparedVectorSearchFilter::table)
```
JingsongLi, line 1295: [P1] Reject a prepared filter from a different table

This makes the prepared filter’s table replace the table that created the builder, without checking that they identify the same physical table. A caller can build on table B, pass a public `PreparedVectorSearchFilter` prepared from table A, and—when the vector column/schema are compatible—silently search A’s snapshot, manifest, files, and options while receiving the result through B’s builder. The returned row IDs/scores can then be treated as belonging to the wrong dataset. The prepared value should pin the snapshot of the builder’s target, not retarget the builder: validate a stable table identity (at least canonical location plus branch/reference) and fail closed on mismatch, with an A/B regression test.

crates/integrations/datafusion/src/lateral_vector_search.rs
```diff
@@ -352,9 +452,198 @@ struct LateralVectorSearchExec {
...
+        let Some(active_leases) = entry.active_partition_leases.get_mut(&partition) else {
+            return;
+        };
+        *active_leases -= 1;
+        if *active_leases == 0 {
+            entry.active_partition_leases.remove(&partition);
+            entry.completed_partitions.insert(partition);
+        }
+        if entry.active_partition_leases.is_empty()
+            && (Arc::strong_count(context) == 1
```
JingsongLi, line 641: [P1] Do not make cleanup depend on the context already being dropped

This still retains a subset execution when the caller keeps its normal `Arc<TaskContext>` until after the returned stream finishes. For example: create a 4-partition execution, execute/consume only partition 0, drop its stream/lease while the caller still owns the context, then drop the caller’s context. At `finish_partition` time the strong count is greater than 1 and not all declared partitions completed, so the entry stays; the later context drop has no callback to re-run this condition. The weak entry is only pruned by a future `for_execution`, leaving the pinned table and potentially large bitmap alive for the physical plan’s idle lifetime. The new subset test avoids this order by dropping `context` before `lease`, so it does not cover the common retained-context case. Please tie the cached value to an execution/context-owned lifetime (or another actual completion hook) and add the stream-first/context-second regression.

crates/paimon/src/table/vector_search_builder.rs
shyjsarah: Fixed in `d11eaaf`. `BatchVectorSearchBuilder::execute` now validates the prepared filter against the builder target before selecting the pinned table: normalized table location (ignoring trailing `/`) and branch must match, otherwise it fails closed with `DataInvalid`. Added an A/B table regression proving a prepared filter cannot retarget another builder.

crates/integrations/datafusion/src/lateral_vector_search.rs
shyjsarah: Fixed in `d11eaaf`. Because DataFusion 54 exposes no `TaskContext` drop hook, the execution cache now runs a single cache-level reaper over weak contexts while entries exist. It removes subset executions after the caller drops its context even when the stream/lease was dropped first; all-partition completion and context-already-gone cleanup remain immediate. The reaper is started only after an entry is installed to avoid a concurrent empty-cache restart race. Added the requested stream-first/context-second regression.

crates/paimon/src/vindex/reader.rs
shyjsarah: Fixed in `d11eaaf`. Batch grouping keeps the `Arc::ptr_eq` fast path for the crate-internal shared allocation and falls back to serialized-filter content equality for equal filters supplied through the public `with_include_row_ids` API. The regression now again builds two separate public filters and verifies filtered native batching/probed-list I/O reuse against scalar execution.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -196,6 +197,33 @@ pub struct BatchVectorSearchBuilder<'a> {
...
+    pub fn table(&self) -> &Table {
+        &self.table
+    }
+
+    pub fn include_row_ids(&self) -> &Arc<RoaringTreemap> {
+        &self.include_row_ids
+    }
+}
+
+fn same_vector_search_table(left: &Table, right: &Table) -> bool {
```
JingsongLi, line 224: [P2] Bind prepared filters to storage identity, not only URI text

Two supported public `Table::new` values can have the same normalized location and branch while their `FileIO`s resolve that path to different storage. For example, separate memory `FileIO` instances are intentionally isolated (see `test_memory_operator_not_shared_between_file_io_instances`), and `with_fs_operator` lets the operator root determine the actual filesystem. Preparing on table A and attaching to table B therefore passes this check; `execute` then switches entirely to the prepared table A and silently returns A row IDs/scores. Please give a Table/FileIO clone family a stable private lineage token, preserve it across table copies, and store/validate it in the prepared filter (with a regression using two isolated FileIOs at the same location).

shyjsarah: Fixed in `45da2d0`. Each `FileIOBuilder::build` now creates a fresh private storage-lineage token, while `FileIO::clone` and all table copy paths preserve it automatically. Prepared-filter validation now requires matching FileIO lineage in addition to normalized location and branch, so equal URI text cannot retarget a builder across isolated memory/custom operators. Added a public prepare-path regression using two independent memory FileIOs at the same location, plus a unit test proving only FileIO clones share lineage. `paimon` lib tests (2482 passed, 2 ignored), DataFusion lateral tests, formatting, and clippy all pass.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -1217,11 +1270,30 @@ impl<'a> BatchVectorSearchBuilder<'a> {
...
         // returns data-derived row ids/scores outside `TableScan`/`TableRead`,
         // so it must refuse a `query-auth.enabled` table before any fast path
         // (an empty snapshot would otherwise return empty results and bypass it).
-        let core = CoreOptions::new(self.table.schema().options());
+        let execution_table = self
+            .prepared_filter
+            .as_ref()
+            .map(PreparedVectorSearchFilter::table)
+            .unwrap_or(self.table);
+        let core = CoreOptions::new(execution_table.schema().options());
```
JingsongLi, line 1297: [P1] Authorize the builder target before switching to the prepared snapshot

The lineage check still permits different schema/option copies of the same physical table, but this guard now checks only `execution_table`. A prepared filter created before `query-auth.enabled` was enabled carries the old auth-disabled schema; after the catalog reloads the table with auth enabled (the same catalog FileIO clone keeps the same lineage), attaching that prepared value passes `same_vector_search_table`, replaces the builder target here, and bypasses the fail-closed guard. I reproduced the minimal public-API form with `source.prepare_vector_search_filter(...)`, `target = source.copy_with_options({query-auth.enabled: true})`, and `target.new_batch_vector_search_builder().with_prepared_filter(prepared).execute()`: it returns `Ok([empty SearchResult])` instead of `Unsupported`. This contradicts the existing time-travel rule that current stored auth cannot be disabled by a historical/dynamic copy. Please call `ensure_read_authorized` on `self.table` before selecting the pinned prepared table (and keep the prepared-table check as defense in depth), with a stale-prepared -> auth-enabled-target regression.

shyjsarah: Fixed in `5b67f99`. `BatchVectorSearchBuilder::execute` now authorizes the live builder target before inspecting or switching to a prepared snapshot, so current auth/type policy remains authoritative even when the prepared value carries older options. The pinned execution table is still authorized separately as defense in depth. Added the public-API stale-prepared -> auth-enabled-target regression; it failed before the fix with an empty successful result and now returns the expected `query-auth.enabled` `Unsupported` error. Paimon lib tests (2527 passed, 2 ignored), DataFusion lateral tests, formatting, and clippy all pass.

JingsongLi (approved):
+1

PR #761, opened 2026-08-28
feat(go): add BlobDescriptor batch and stream readers

XiaoHongbo-Hope:
## Summary

- expose the merged Rust `BlobReader` and `BlobStream` through C and Go
- add ordered batch reads and lazy `io.ReadSeekCloser` streams
- let `table.NewBlobReader()` reuse the table FileIO, including DLF token refresh

Descriptor parsing, validation, range merging, concurrency limits, and storage I/O remain in Rust core (#762 and #763). The Go layer only manages FFI ownership and Go-friendly APIs.

## Go API

```go
reader, err := paimon.NewBlobReader(storageOptions)
reader, err := table.NewBlobReader()

values, err := reader.ReadBlobs(descriptors)
stream, err := reader.OpenBlob(descriptor) // io.ReadSeekCloser
```

`ReadBlob` delegates to `ReadBlobs`. Batch reads merge compatible ranges and preserve input order. Independent streams are lazy and are not merged.

Static options passed to `NewBlobReader` are not refreshed. `table.NewBlobReader()` retains the table's refreshing FileIO for DLF temporary data tokens.

## Validation

- `cargo test --locked -p paimon-c --lib`: 71 passed
- Go BlobReader/BlobStream runtime tests: 4 passed; local table-fixture case skipped
- `go vet`, rustfmt, clippy, dependency report, and diff checks passed

This PR adds only the C/Go binding layer; the Rust core implementation is already merged.

### DIFF
--- a/docs/src/go-binding.md
+++ b/docs/src/go-binding.md
@@ -39,6 +39,144 @@ go get github.com/apache/paimon-rust/bindings/go
 The native library is embedded and loaded automatically. Build with
 `CGO_ENABLED=1`.
 
+## Reading BlobDescriptor Values
+
+`BlobReader` reads serialized `BlobDescriptor` values without scanning a table.
+`ReadBlobs` resolves a batch in one call; `ReadBlob` handles one descriptor.
+
+```go
+package main
+
+import (
+    "database/sql"
+    "log"
+    "os"
+
+    paimon "github.com/apache/paimon-rust/bindings/go"
+    _ "github.com/go-sql-driver/mysql"
+)
+
+func main() {
+    db, err := sql.Open("mysql", os.Getenv("STARROCKS_DSN"))
+    if err != nil {
+        log.Fatal(err)
+    }
+    defer db.Close()
+
+    rows, err := db.Query("SELECT blob_descriptor FROM catalog.db.my_table")
+    if err != nil {
+        log.Fatal(err)
+    }
+    defer rows.Close()
+
+    var descriptors [][]byte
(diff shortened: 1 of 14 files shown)


docs/src/go-binding.md
```diff
@@ -39,6 +39,143 @@ go get github.com/apache/paimon-rust/bindings/go
...
+defer reader.Close()
+```
+
+The reader and its streams keep the table FileIO and refresh DLF data tokens
+before expiry. Set `dlf.oss-endpoint` when the server-provided endpoint is not
+reachable from the application. Static options passed to
+`paimon.NewBlobReader` are not refreshed.
+
+Reads are grouped by URI and nearby ranges are merged. The fixed limits are a
+64 KiB merge gap, 8 MiB merged span, 8 concurrent requests, and 64 MiB of
```
JingsongLi: [P2] Clarify the 64 MiB admission budget

This reads as a hard cap, but BlobReadLimiter::acquire_read clamps an oversized range to all permits and still reads and materializes the full range before the Go binding copies it. A single 1 GiB descriptor can therefore consume far more than 64 MiB. Please describe this as a per-reader admission threshold, call out the oversized-range exception, and recommend OpenBlob for large values.

XiaoHongbo-Hope: Updated the docs to describe 64 MiB as a per-reader admission budget, note that one oversized range may exceed it, and recommend OpenBlob for large values.

JingsongLi (approved):
+1

PR #762, opened 2026-08-29
feat(blob): add standalone BlobDescriptor reader

XiaoHongbo-Hope:
## Summary

- add a Rust `BlobReader` for serialized `BlobDescriptor` values
- reuse existing validation, FileIO, URI grouping, range merging, and bounded concurrency
- preserve input order and sanitize descriptor errors
- support an existing FileIO so catalog-provided credential refresh is retained

This is the Rust core split from draft #761. C and Go bindings will follow separately.

## Testing

- `cargo test -p paimon table::blob_resolver::tests --lib` (11 passed)
- DLF token-refresh BlobReader regression test
- rustfmt and diff checks

### DIFF
--- a/crates/paimon/src/table/blob_resolver.rs
+++ b/crates/paimon/src/table/blob_resolver.rs
@@ -57,7 +57,7 @@ impl BlobReader {
 
     /// Read a descriptor batch in input order.
     pub async fn read_blobs(&self, descriptors: &[Vec<u8>]) -> Result<Vec<Vec<u8>>> {
-        let mut by_uri = HashMap::<String, Vec<(usize, &[u8])>>::new();
+        let mut by_uri = HashMap::<String, Vec<(usize, BlobDescriptor)>>::new();
         for (index, bytes) in descriptors.iter().enumerate() {
             let descriptor = BlobDescriptor::deserialize(bytes)
                 .map_err(|error| blob_error_with_context(error, &[index], None))?;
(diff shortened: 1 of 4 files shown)


crates/paimon/src/table/blob_resolver.rs
```diff
@@ -32,6 +32,152 @@ pub(crate) const BLOB_DESCRIPTOR_READ_CONCURRENCY: usize = 8;
...
+                            })
+                            .map_err(|error| {
+                                blob_error_with_context(error, &indices, Some(&uri))
+                            })?,
+                    };
+                    let mut builder = BinaryBuilder::with_capacity(entries.len(), 0);
+                    for (_, descriptor) in &entries {
+                        builder.append_value(descriptor);
+                    }
+                    let resolved = resolve_blob_column(&builder.finish(), &file_io, limiter)
```
QuakeWang, line 100: Valid Java v1 descriptors are silently returned as raw bytes here. `read_blobs` accepts them via `BlobDescriptor::deserialize`, but `resolve_blob_column` only recognizes the v2 magic header, so the referenced range is never resolved. Please use a strict descriptor path (or normalize to v2) and add a v1 regression test.

XiaoHongbo-Hope: Fixed in 7ae349f. Parsed descriptors are normalized to the current v2 representation before entering the mixed-column resolver, and a Java v1 range-read regression test now verifies the referenced bytes are returned.

QuakeWang (approved):
LGTM

JingsongLi (approved):
+1

PR #763, opened 2026-08-30
feat(blob): add seekable descriptor streams

XiaoHongbo-Hope:
## Summary

- extend the standalone `BlobReader` from #762 with `open_blob`
- add lazy, incremental range reads and start/current/end seeking
- share the existing request and in-flight byte limiter across batch and stream reads
- preserve catalog-managed credential refresh when reusing table `FileIO`

This PR contains Rust core support only. C and Go bindings remain separate follow-up work from draft #761. Batch descriptor grouping and IO merge behavior are unchanged.

## Testing

- `cargo test -p paimon table::blob_resolver::tests --lib` (13 passed)
- `cargo test -p paimon test_blob_stream_reuses_refreshing_file_io --lib`


QuakeWang (approved):
LGTM

PR #764, opened 2026-08-31
feat(file_index): add indexer contracts and outer-format composition

### LINKED ISSUE #34: Implement FileIndex

QuakeWang:
### Purpose

Closes: #34

Bitmap and Bloom implementations already existed, but callers still had to match identifiers and construct concrete readers and writers directly. This leaked implementation details into future scan and writer integrations.

This change adds crate-private contracts and built-in dispatch so outer-format index payloads can be composed without depending on concrete index implementations.

### Brief change log

  - Add a common `FileIndexWriter` contract and built-in Bitmap/Bloom factory.
  - Build per-column readers from outer-format metadata and schema.
  - Handle nested map columns, empty indexes, type/options validation, and unknown identifiers.
  - Add component E2E coverage for multiple columns and indexes.

### Tests

### DIFF
--- a/crates/paimon/src/file_index/file_index_format.rs
+++ b/crates/paimon/src/file_index/file_index_format.rs
@@ -20,7 +20,12 @@ use std::collections::HashMap;
 use bytes::{BufMut, Bytes, BytesMut};
 
 use crate::{
+    file_index::{
+        file_index_reader::{EmptyFileIndexReader, FileIndexReader},
+        file_indexer_factory::FileIndexerFactory,
+    },
     io::{FileIO, FileRead, FileStatus, InputFile, OutputFile},
+    spec::{DataField, DataType},
     Error,
 };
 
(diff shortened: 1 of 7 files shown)


crates/paimon/src/file_index/file_index_format.rs
```diff
@@ -329,12 +334,74 @@ fn calculate_head_length(
...
+        &self,
+        fields: &[DataField],
+    ) -> crate::Result<HashMap<String, Vec<Box<dyn FileIndexReader>>>> {
+        let fields_by_name = fields
+            .iter()
+            .map(|field| (field.name(), field.data_type()))
+            .collect::<HashMap<&str, &DataType>>();
+        let mut readers = HashMap::with_capacity(self.header.len());
+
+        for (column_name, index_info) in &self.header {
```
JingsongLi: [P2] Please avoid eagerly materializing every index payload here. `create_index_readers` walks the complete header, serially reads each full range, and returns concrete readers that retain those `Bytes`, even when the query predicate uses only one column. A legal payload is already 100 MB in `test_large_data_set`, so concurrent scans can multiply unnecessary remote reads and pinned memory before evaluation. The Java `FileIndexPredicate` collects predicate field names and calls `readColumnIndex` only for those columns. Could this API accept the required columns (or keep `FileRead` plus ranges and load lazily), with a counting-reader test proving unrelated ranges are not fetched?

JingsongLi (approved):
+1

PR #769, opened 2026-09-01
refactor(vector-search): remove extra prefilter guards

JingsongLi:
## Summary

- remove DataFusion-specific NaN screening for vector pre-filters
- remove the FileIO storage-lineage token and validate prepared filters by normalized location and branch
- retain snapshot pinning and query-authorization checks

This keeps the Rust vector-search behavior and API surface closer to the Java implementation while removing edge-case-specific guards added in #760.

## Testing

- cargo fmt --all --check
- cargo test -p paimon-datafusion --lib filter_pushdown::tests
- cargo test -p paimon-datafusion --lib vector_search
- cargo test -p paimon --lib table::vector_search_builder::tests
- git diff --check

### DIFF
--- a/crates/integrations/datafusion/src/filter_pushdown.rs
+++ b/crates/integrations/datafusion/src/filter_pushdown.rs
@@ -89,20 +89,6 @@ pub(crate) fn analyze_filters(
     }
 }
 
-pub(crate) fn is_safe_vector_prefilter(predicate: &Predicate) -> bool {
-    match predicate {
-        Predicate::Leaf { literals, .. } => !literals.iter().any(|literal| {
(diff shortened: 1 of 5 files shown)


leaves12138 (approved):
Reviewed the latest commit. Removing the DataFusion-only NaN gate keeps vector pre-filtering on the normal translated Predicate/residual path, and the prepared-filter identity check still binds reuse to the normalized table location and branch while preserving snapshot pinning and both live-target and pinned-table authorization checks. The targeted filter-pushdown, vector-search, and vector-search-builder suites pass locally (143 tests), and formatting is clean.

PR #770, opened 2026-09-01
fix(spec): enforce the writer's layout in the string array decoder

JunRuiLee:
## Why

The next PR in #755 adds a C entry point that takes `BucketVectorSearchSplit` bytes from outside
the process — an engine's BE hands them in. Those bytes reach `DataFileMeta` row decoding, and one
of the decoders it goes through, `deserialize_binary_array_str`, checks only that each element's
region lies inside the buffer. It does not check how the regions relate to each other, so `n`
elements may all point at one large body and each get cloned into a `String`: output bounded around
`len²/8`. #746 already closed this for the row-array variant; this is the last variable-length
variant still open, and #755 asks for it before the C entry point, not after.

Two gaps are already wrong today, independent of that:

- **A set null bit is ignored** and the slot read as if it held a value, so the decoder returns a
  string that was never written. On `main` the new test reads `["hello"]` where Java's reader would
  return null.
- **A body addressed inside the fixed part** returns the array's own header as a value. Those bytes
  are valid UTF-8, so this decoded successfully — on `main` the new test reads
  `["\u{1}\0\0\0\0\0\0\0"]`.

## How

Java's writer pins the layout exactly, so no heuristic cap is needed — requiring what it emits
rejects everything else as a side effect. `BinaryArrayWriter` starts `cursor` after the fixed part;
`writeString` sends a body of `len <= 7` to the element's own slot and **leaves `cursor` alone**,
anything longer to `writeBytesToVarLenPart`, which advances `cursor` by the word-padded length; and
`complete()` sizes the array at exactly that cursor. So: bodies are consecutive, the buffer ends at
the last one, and inline bodies consume no variable-length space.

Walking that cursor and requiring it rejects aliasing, gaps, reordering and trailing bytes. Plus
the two above: a set null bit (all three call sites declare non-null elements — `_EXTRA_FILES`,
`_VALUE_STATS_COLS`, `_WRITE_COLS`), and an inline marker claiming more than 7 bytes when its seven
length bits allow up to 127.

`deserialize_binary_array_int` and `_long` have no variable-length part and are already exact and
bounded. The row variant is #746's.

## Testing

Seven tests; the six negative ones fail on `main`.
`binary_array_str_all_inline_is_exactly_its_fixed_part` pins the inline rule so the cursor walk
cannot over-reject a legal all-inline array. The existing goldens (`split_v1_data.bin`,
`split_v1_indexed.bin`, `bucket_vector_search_split_v1*.bin`, `datasplit_v8/v9.bin`) decode
`DataFileMeta` rows through this function and still pass, which is the evidence the new strictness
matches real writer output.

Gates: `cargo fmt --all -- --check`; `cargo clippy -p paimon --all-targets -- -D warnings` and with
`--features fulltext`, both clean; `cargo test -p paimon --lib` 2586 passed (2579 on `main`, +7,
nothing skipped) and 2661 with `--features fulltext`; `cargo build -p paimon-datafusion` and
`-p paimon-c`.

### DIFF
--- a/crates/paimon/src/spec/binary_row.rs
+++ b/crates/paimon/src/spec/binary_row.rs
@@ -765,6 +765,11 @@ pub fn serialize_binary_array_long(values: &[Option<i64>]) -> Vec<u8> {
     data
 }
 
+/// Largest string body the writer keeps in an element's own 8-byte slot
+/// (Java `BinarySection.MAX_FIX_PART_DATA_SIZE`): seven bytes of content plus the
+/// marker byte that carries the length.
+const MAX_INLINE_STRING_LEN: usize = 7;
+
 /// Reverse of [`serialize_binary_array_str`].
 pub fn deserialize_binary_array_str(data: &[u8]) -> crate::Result<Vec<String>> {
     let n = read_binary_array_len(data)?;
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #771, opened 2026-09-01
feat(table): execute a primary-key vector search over engine-planned splits

JunRuiLee:
### Purpose

Linked issue: part of #755 (final step).

Execute a primary-key (bucket-local ANN) vector search that an engine planned
elsewhere. Paimon Java plans `BucketVectorSearchSplit`s on a planner node and
ships each to a worker, which calls this entry point. Planning stays in Java;
execution comes here.

```rust
let mut search = table.new_vector_search_builder();
search
    .with_vector_column("v")
    .with_query_vector(q)
    .with_limit(k)
    .with_projection(&["id", "v"]);
let rows = search.execute_read_for_bucket_splits(&split_bytes).await?;
```

Split bytes in, Arrow rows out, in one call — option 1 from the 2026-09-09 scope
review, and the boundary #755 step 4 specifies. Rows come back best-first with
`__paimon_search_score`; query parameters and projection reuse the existing
`with_*` methods, so everything after planning is the ordinary primary-key vector
read.

The splits are the plan: their payload files, per-file row ranges and pinned
snapshot are used as given, and this table's index manifest is not read. Top-K is
local to the supplied splits — a caller distributing one call per bucket merges
the results itself. Only a primary-key vector column can be read this way; the
data-evolution route plans through the global index and is rejected.

The staged `VectorRead` / `TableRead` API is out of scope here and parked in #804
until it has a consumer, as #755 specifies. The two are independent: this PR does
not need it, and it can re-express this entry point as a thin wrapper whenever it
lands.

30 files, +2,316 −343. Three hardening fixes found along the way were split out
and are merged: #801, #802, #803. This branch uses #801's `take_only_result` for
its own single-query unwrap, and inherits #802's deletion-vector position bound,
which turned out to be load-bearing here: a split names its own deletion files,
so without it a forged position past its own file silently drops a row from the
next one.

### Brief change log

The entry point is the small part. The rest follows from the trust boundary: the
plan, and the row counts inside it, come from outside the process.

**A file the split does not list means "unrestricted"**, as Java's
`rowRangesByFile` specifies. `FileRowSelection` is the per-file three-state
selection: absent unrestricted, empty excluded, non-empty restricted. The kernel
previously read a missing entry as "no rows allowed" — the opposite of Java — so
planning normalized every omitted file into an explicit `[0, row_count - 1]`
range.

That expansion is gone. Ranges stay ranges from planning through the ANN mask,
the ANN result check and the exact fallback, and a split that narrowed nothing
(`rangeFileCount == 0`, both the committed fixture and the default shape an
engine ships) now reaches the ANN backend with no filter at all, instead of an
all-permitting mask that Lumina turns into one `u64` per live row.

Endpoints are validated where they enter:

- A listed range is checked against its source file's row count, as Java checks
  it, catching a range that disagrees with its own segment.
- Row counts are checked non-negative at decode for **every** data file, not only
  the ones the message lists ranges for.
- `MAX_LIVE_ROW_IDS` bounds a segment's mask, charged before each insertion.
  Without it, one restricted sibling file or one deletion vector inserts an
  unrestricted file wholesale at whatever size its wire-supplied count claims.
  The value is Java's own for this quantity
  (`LuminaVectorGlobalIndexReader.toScopedIds` refuses above `Integer.MAX_VALUE`).
- `lumina.search.max-filter-bytes` (default 64 MiB) bounds the dense-filter
  allocation under concurrent searches. A Rust reader budget, not passed to
  native Lumina.

`paimon_vector_search_builder_execute_read_for_bucket_splits` is one new C symbol
returning a `paimon_result_record_batch_reader`. It marshals the split array,
decodes at that boundary, and runs the same Rust entry point.

### Tests

The fixture is Java-generated: table directory written by Java's `ivf-flat`
indexer, split bytes planned by Java's `PrimaryKeyVectorScan` over that same
table. Provenance and regeneration command are in the test file's module doc.

Rust integration (6):

- `reads_java_planned_bucket_split` — end to end over Java's own bytes.
- `restricts_the_read_to_the_splits_row_ranges` — narrowing the split's ranges
  must drop a row from the result. This is what separates "used the split" from
  "quietly re-planned from the manifest".
- `agrees_with_the_manifest_route` — the two routes return the same rows and
  scores.
- `honors_a_narrower_limit`, `rejects_an_empty_split_list`,
  `rejects_corrupt_split_bytes`.

Unit tests cover the three-state selection, each `MAX_LIVE_ROW_IDS` charge site,
the range/count checks and the dense-filter budget, each mutation-checked against
the code it covers.

C (3): argument marshalling, corrupt bytes, one happy path over the same Java
bytes. The semantic contract stays on the Rust side, where a failure names what
broke.

Gates: `cargo test -p paimon --lib` 2,645 passed / 2 ignored; bucket-split suite
6 passed; `cargo fmt --all -- --check` clean. CI is green on this head (14/14),
which is where the workspace clippy gate actually runs.

### API and Format

One new Rust method, one new C ABI symbol; no existing signature changes.

No storage-format change — the `BucketVectorSearchSplit` byte form is Java's
`PKVSPLIT` v1, decoded by #746 (merged), and nothing new is serialized.

Behaviour change to an existing path: a split that omits a file from
`rowRangesByFile` now means "unrestricted" rather than "no rows". That is the
Java semantic; the previous reading meant a valid split could silently return
nothing.

### Documentation

`docs/src/sql.md` gains the `lumina.search.max-filter-bytes` option.


crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -416,6 +417,131 @@ impl<'a> VectorSearchBuilder<'a> {
...
+            &[query_vector.as_slice()],
+            limit,
+        )?;
+        let plan = PkVectorScan::new(
+            self.table,
+            params.field_id,
+            params.index_type.clone(),
+            self.filter.clone(),
+        )
+        .plan_for_bucket_vector_splits(splits)?;
```
JingsongLi, line 536: [P1] Bound or preserve decoded row ranges before bitmap expansion

This new entry point forwards engine-supplied DataFileMeta.row_count values into plan_for_bucket_vector_splits without a resource bound. When a file omits its range entry (which is valid in the Java encoding), planning synthesizes [0, row_count - 1], and the search path materializes that range into a RoaringTreemap. Both roaring 0.11.4 and the current 0.11.5 implementation iterate every high-u32 shard in insert_range and construct a full u32 bitmap for each intermediate shard; row_count = 2^40 already creates 256 full bitmaps (about 16.8 million containers), while a value near i64::MAX can hang or OOM the worker before any storage I/O. The later i32 row-position validation runs only after this expansion. Please reject unreasonable row counts/endpoints and, preferably, preserve unrestricted or interval-form selections through the ANN layer instead of eagerly materializing every allowed position.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -416,6 +417,131 @@ impl<'a> VectorSearchBuilder<'a> {
...
+            self.table,
+            &self.options,
+            self.filter.as_ref(),
+            &core,
+            &pk_col,
+            &[query_vector.as_slice()],
+            limit,
+            &plan,
+            &params,
+        )
```
JingsongLi, line 552: [P2] Keep unrestricted files off the filtered ANN path

A Java split with rangeFileCount == 0 is the normal no-prefilter form (including the committed fixture), but the planner normalizes every omitted file into an explicit full-file range. This call then converts those ranges into RoaringTreemaps and passes Some(filter) to ANN even though every row is allowed. Lumina subsequently collects the bitmap into a Vec<u64> and invokes search_with_filter, adding O(live_rows) setup and at least 8 * live_rows bytes for the ID vector per segment search. Please preserve an unrestricted sentinel (with a per-file unrestricted/restricted/excluded state for mixed splits), and add a backend-facing test asserting that the no-prefilter fixture uses the unfiltered ANN path.

crates/paimon/src/table/vector_search_builder.rs
JunRuiLee: Fixed in 57ad142.

Both comments trace to one line: planning normalized an unlisted file into `[0, row_count - 1]`, because our kernel read a missing entry as "no rows" — the opposite of Java. Now mirrored: `whole_file_range` and `positions_in_ranges` are gone, and ranges stay ranges through the mask, the result check and the exact fallback.

Endpoints are rejected in three places: a listed range against its source file's row count (as Java checks it), non-negative row counts at decode for *every* data file, and `MAX_LIVE_ROW_IDS` charged before each insertion into the mask.

That last one turned out to be necessary: removing the planning-side expansion alone is not enough, because the mask spans the segment's *source* row counts, which also come off the wire. One restricted sibling file or one deletion vector is enough to insert an unrestricted file wholesale — the added tests OOM-kill the process without the bound. The limit is Java's own for this quantity (`LuminaVectorGlobalIndexReader.toScopedIds` refuses above `Integer.MAX_VALUE`); our dense conversion had no such guard, so `to_scoped_ids` mirrors it.

One thing worth flagging: bounding the claimed span by the index file's byte length looks reasonable but is unsound — `PkVectorAnnSegmentFile` advances the ordinal space by each source's full physical row count while writing vectors only for non-null, non-excluded rows, so a legitimate mostly-null source has a small payload.

Unrelated bug found in the same function: deletion-vector positions were added without checking them against their own source file, so a position past that file's rows deleted a row in the *next* file.

crates/paimon/src/table/vector_search_builder.rs
JunRuiLee: Fixed in 57ad142.

`FileRowSelection` is the per-file state, with absence carrying the unrestricted case exactly as Java's `rowRangesByFile.get(file) == null` — absent unrestricted, empty excluded, non-empty restricted. Mixed splits are handled per file through the ANN mask, the result check and the exact fallback.

A split that narrowed nothing now reaches the backend with `include_row_ids` unset. Two tests:

- `the_no_prefilter_shape_reaches_the_backend_with_no_filter_at_all` — a recording scorer asserts `effective_include_row_ids()` is `None`;
- `the_java_fixture_reaches_the_ann_layer_with_no_mask_at_all` — the committed fixture through planning into `build_live_row_ids`. Result equality against the manifest route cannot see this difference.

The early "no mask needed" return deliberately keeps Java's whole-map `rowRangesByFile.isEmpty()` rather than a segment-local test: the narrower one would leave more segments unfiltered, but it changes which backend entry point they take, and `search` / `search_with_filter` can differ in recall.

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -416,6 +418,131 @@ impl<'a> VectorSearchBuilder<'a> {
...
+    /// search, optional refine, local Top-K and materialization stay identical to
+    /// [`execute_read`](Self::execute_read): output is the projected user columns
+    /// plus `__paimon_search_score`, best-first. The Top-K is local to the supplied
+    /// splits; a caller distributing one call per bucket merges the per-bucket
+    /// results itself.
+    ///
+    /// Only a primary-key vector column can be read this way. The data-evolution
+    /// route plans through the global index rather than through bucket splits, so
+    /// it is rejected rather than silently answered from a different plan.
+    pub async fn execute_read_for_bucket_splits(
```
JingsongLi, line 468: Why not use VectorRead to read and use TableScan and TableRead to read real data?

JingsongLi:
Please show the API and let us discuss it.

JunRuiLee:
> Please show the API and let us discuss it.

Thanks for remind. I've updated the pr description to show the API.

The API now follows Java's two-step search/read model without adding a generic Split abstraction to Rust:

```
let mut search = table.new_vector_search_builder();
search.with_vector_column("v").with_query_vector(q).with_limit(k);
let splits = search.search_for_bucket_splits(&split_bytes).await?;

let mut read_builder = table.new_read_builder();
read_builder.with_projection(&["id", "v"])?;
let read = read_builder.new_read()?;
let schema = read.indexed_read_type()?;
let rows = read.to_arrow_indexed(&splits)?;
```

crates/paimon/src/table/vector_search_builder.rs
```diff
@@ -416,6 +439,160 @@ impl<'a> VectorSearchBuilder<'a> {
...
+    /// The splits are NOT in best-first order and the read does not reorder them:
+    /// rows come back in physical order carrying `__paimon_search_score`, and a
+    /// caller wanting them ranked sorts on that column. Java does the same -- its
+    /// read emits physical order and `VectorSearchProcedure` sorts afterwards.
+    ///
+    /// Only a primary-key vector column can be searched this way. The
+    /// data-evolution route plans through the global index rather than through
+    /// bucket splits, so it is rejected rather than silently answered from a
+    /// different plan.
+    pub async fn search_for_bucket_splits(
```
JingsongLi: Why we need to introduce this?

JunRuiLee: Thanks. I removed `search_for_bucket_splits` from `VectorSearchBuilder` and moved the execution to `VectorRead`.

  The API is now:

  ```rust
  let vector_read = builder.new_vector_read()?;
  let indexed_splits = vector_read.read(bucket_splits).await?;
  let rows = table_read.to_arrow_indexed(&indexed_splits)?;

  The externally planned BucketVectorSearchSplits are used directly without re-planning from the manifest. The C API only decodes the split bytes and calls the same VectorRead implementation.

```
I deliberately did not introduce a public VectorScan::Plan or new_vector_scan() in this PR, because this integration already receives the externally planned bucket splits.

**Comment by JingsongLi** [comment]
Thanks for restructuring the API. I took another pass focused specifically on scope and reviewability.

The PR is now **38 files** and roughly **+4,298/-383**. My main concern is that it now includes the staged public read API that #755 explicitly left for a follow-up until there is a concrete consumer. This includes:

- `VectorRead` / `new_vector_read`
- public `PkVectorIndexedSplit`
- `TableRead::{to_arrow_indexed, indexed_read_type}`
- the C split-result handle plus count/free/read functions
- their supporting builder state and tests

Commit `048eb800` alone is about **+2,153/-154**.

Unless this PR deliberately supersedes the scope recorded in #755, please either:

1. Keep #771 on the original one-shot boundary: externally planned bucket splits in, Arrow rows out; or
2. Extract the generic staged `VectorRead` / `TableRead` API into a separate PR with its own consumer and contract discussion.

That should remove roughly half of the current review surface while preserving the core feature.

There are also several useful but independent hardening changes that should be separate PRs:

- Rejecting a deletion-vector position that crosses into the next source file.
- Existing Lumina result-path changes for NaN ordering, sentinel-filled label buffers, and tie-breaking.
- Lumina empty-index error-precedence changes.
- Scalar batch-result arity checks in existing search wrappers.

For #771 itself, I would keep the changes required by the new trust boundary:

- The three-state `FileRowSelection` semantics.
- Keeping unrestricted files off the filtered ANN path.
- Direct planning from the supplied split.
- Range/count validation.
- The Lumina dense-filter memory budget.

The tests can also be reduced without losing meaningful coverage:

- `bucket_split_decoder_rejects_corrupt_bytes` no longer exercises `VectorRead`, because the Rust API accepts already-decoded splits.
- Projection/filter rejection is repeated across Rust integration tests, `TableRead` unit tests, and C integration tests. Keep the semantic contract at the Rust level; the C layer should focus on one Java-fixture happy path plus pointer/buffer/handle safety.

The earlier safety fixes look valuable. This is a request to keep this PR focused and move independent fixes into their own reviews, not to drop those fixes entirely.

**Comment by JunRuiLee** [comment]
Thanks — restructured along the scope review.

### #771 is back on the one-shot boundary

30 files / +2,236−343, down from 38 / +4,298−383. Moved out:

- The staged read API → #804 (draft), up now so it can be seen rather than only described.
- A deletion-vector position bound → #802
- Lumina's empty-index error precedence → #803
- The scalar batch arity check → #801

The three fixes are on `main`, independent of each other and of this PR. #802 and #803 were mutation-checked: reverting each fix individually fails exactly its own test.

### On the two options

I took both, because they compose: this PR returns to option 1's boundary, and option 2 becomes the follow-up. That keeps the answer to your 2026-09-03 comment here — *"Why not use VectorRead to read and use TableScan and TableRead to read real data?"* — on the table with its own contract discussion, rather than dropped to shrink the diff.

#804 keeps the one-shot terminal instead of replacing it, which is what the earlier revision did. Once this PR is in, that method and its C ABI symbol are published surface, so removing them would be a breaking change rather than the cleanup it would have been while the branch was unmerged. That is what makes #804 additive (+2,168/−30 over this PR) — roughly the other half of the surface you were seeing.

### Test convergence

- `bucket_split_decoder_rejects_corrupt_bytes` is now a decoder assertion. The entry point decodes serialized bytes at that boundary, so that is where corrupt input is refused; it no longer drives a read behind a query that cannot execute either way.
- Projection/filter rejection is asserted once, at the Rust level. The C suite is argument marshalling, corrupt bytes, and one happy path over the Java fixture — plus handle and pointer safety in #804, where the new handles are.

`vector_search_append_filter_returns_invalid_input` still fails under `-p paimon-c`. It fails identically on `origin/main`; unrelated, left alone.

**Comment by JunRuiLee** [comment]
Rebased onto #801, #802 and #803, now that all three are merged. 30 files / +2,271−343.

One thing worth recording, because it changes how one of them should be read: **#802 was load-bearing for this PR's trust boundary, not an independent hardening fix.**

A bucket split names its own deletion files, so the deletion positions come off the same untrusted bytes as everything else in the plan:

```
BucketVectorSearchSplit  ->  embedded DataSplit.data_deletion_files()
  -> build_split_dv_factory(&split.data_split)      pk_vector_orchestrator.rs
    -> DeletionVectorFactory::new(io, data_files, data_deletion_files)
      -> positions -> build_live_row_ids -> global = file_offset + position
```

Without #802's bound, a position past its own file lands in the **next** file's ordinal range and drops one of that file's rows — silently, because the mask stays well formed. Had this PR merged first, that would have been reachable from a forged split. It read as one of three unrelated fixes and one of them was not; it is on main now, so this branch inherits it.

Also new: `an_empty_index_outranks_a_filter_too_large_to_densify`, covering where this PR's dense-filter byte budget meets #803's step order. Neither pins it alone — without the ordering the budget rejects the query, without the budget there is nothing to order against — so it could only be written after #803 landed.

This branch also uses #801's `take_only_result` for its own single-query unwrap rather than repeating the check inline.

JingsongLi (approved):
+1

PR #781, opened 2026-09-02
datafusion: cast lateral vector search rows to provider schema

shyjsarah:
## Summary

Fix lateral `vector_search` queries that return string columns. Paimon materializes strings as Arrow `Utf8`, while DataFusion may expose the table-provider schema as `Utf8View`; constructing the lateral result batch without aligning those types caused an Arrow schema mismatch at execution time.

## Changes

- Align each materialized target column with the DataFusion provider schema before constructing the lateral result batch.
- Resolve target columns by name rather than relying on positional layout.
- Extend the existing string-column vector search regression to cover the lateral path and validate returned string values.

## Testing

- [x] `cargo test -p paimon-datafusion --test read_tables vector_search_tests::test_vector_search_casts_string_column -- --exact`
- [x] `cargo test -p paimon-datafusion --lib lateral_vector_search::tests`
- [x] `git diff --check`
- [x] `cargo clippy -p paimon-datafusion --lib --tests -- -D warnings`

## Notes

This change does not affect the public API or vector-search ranking semantics. It applies the same schema-alignment behavior already used by the non-lateral vector search materialization path.

### DIFF
--- a/crates/integrations/datafusion/src/lateral_vector_search.rs
+++ b/crates/integrations/datafusion/src/lateral_vector_search.rs
@@ -31,6 +31,7 @@ use datafusion::arrow::array::{
     new_empty_array, Array, ArrayRef, FixedSizeListArray, Float32Array, Int64Array, ListArray,
     RecordBatch, UInt32Array,
 };
+use datafusion::arrow::compute::cast;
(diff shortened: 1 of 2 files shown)


jerry-024 (approved):
+1

PR #782, opened 2026-09-02
feat(datafusion): support the $consumers system table

jerry-024:
### Purpose

Support querying Java-compatible Paimon consumer progress through the DataFusion `$consumers` system table.

### Brief change log

- Add a branch-aware `ConsumerManager` for reading `consumer/consumer-*` progress files.
- Match Java master’s consumer JSON format and retry behavior for files being overwritten.
- Register `$consumers` with the columns `consumer_id` and `next_snapshot_id`.
- Push down `consumer_id` equality and `IN` filters to avoid reading unrelated consumer files.
- Preserve backslashes in POSIX `file://` paths so consumer IDs round-trip correctly.
- Read consumer files with bounded concurrency.
- Document the new system table.

### Tests

- `cargo +1.94.0 test --locked -p paimon consumer_manager --lib`
- `cargo +1.94.0 test --locked -p paimon fs_relative_path_tests --lib`
- `cargo +1.94.0 test --locked -p paimon-datafusion --test system_tables`
- `cargo +1.94.0 clippy --locked -p paimon-datafusion --tests -- -D warnings`

### API and Format

Adds `ConsumerManager` and `Table::consumer_manager()` to the Rust API. No storage format changes; the implementation reads the existing Java-compatible `consumer-<id>` files and `nextSnapshot` JSON field.

### Documentation

Adds `$consumers` usage and column documentation to `docs/src/sql.md`.

### DIFF
--- a/crates/integrations/datafusion/src/system_tables/consumers.rs
+++ b/crates/integrations/datafusion/src/system_tables/consumers.rs
@@ -75,16 +75,10 @@ impl TableProvider for ConsumersTable {
         let manager = self.table.consumer_manager();
         let requested_ids = requested_consumer_ids(filters);
         let consumers = crate::runtime::await_with_runtime(async move {
-            let Some(ids) = requested_ids else {
-                return manager.list_all().await;
-            };
-            let mut consumers = Vec::with_capacity(ids.len());
-            for id in ids {
-                if let Some(next_snapshot) = manager.get(&id).await? {
-                    consumers.push((id, next_snapshot));
-                }
+            match requested_ids {
+                Some(ids) => manager.list_by_ids(&ids).await,
+                None => manager.list_all().await,
             }
-            Ok(consumers)
         })
         .await
         .map_err(to_datafusion_error)?;
--- a/crates/integrations/datafusion/src/system_tables/consumers.rs
+++ b/crates/integrations/datafusion/src/system_tables/consumers.rs
@@ -1,0 +1,196 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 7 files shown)


shyjsarah:
Thanks for adding `$consumers`. The filter pushdown and Java-compatible JSON handling look good overall. I found two minor compatibility/correctness issues:

1. **The overwrite retry window is shorter than Java's** (`crates/paimon/src/table/consumer_manager.rs:27-28,66-84`). Java `release-1.3` retries mismatched consumer JSON 10 times with a 1-second delay, while this implementation uses 200 ms, exhausting the retries after about 1.8 seconds. I reproduced a case where the file becomes valid after 2.1 seconds: Rust returns `DataInvalid`, while Java would still be retrying. Could we align the interval/window with Java and add a delayed-overwrite test?

2. **A POSIX consumer ID containing a backslash is silently omitted by `list_all`** (`crates/paimon/src/table/consumer_manager.rs:103-113`). The listed filename preserves `\\`, but the local `file://` path conversion normalizes it to `/` when `get(id)` reconstructs the path, so a file such as `consumer-id\\part` is listed and then read from a different path. Reading the listed `FileStatus.path` directly, or consistently validating such IDs, would avoid this.

Non-blocking: consumer files are currently read serially. Bounded concurrency may be worth considering for object-store tables with many consumers.

I also ran the PR's targeted tests and clippy commands locally; they all passed.

shyjsarah:
Correction to my earlier review: the retry-window finding was based on Java `release-1.3`, which used 1-second retry intervals. This PR explicitly targets Java **master**, and current Java master uses the same 10 attempts with 200-ms intervals as this Rust implementation. Therefore, the retry behavior is compatible with the stated target and that finding should be disregarded.

I rechecked commit `d095349`: the POSIX backslash-ID issue is fixed, the bounded-concurrency change looks good, and the targeted tests/clippy pass. No remaining blocking findings from my review. Sorry for the baseline mismatch.

shyjsarah (approved):
+1

JingsongLi (approved):
+1

PR #784, opened 2026-09-02
feat(blob): support MAP<K, BLOB> reads in Go

XiaoHongbo-Hope:
## Summary

- read Java/PyPaimon `MAP<K, BLOB>` layout as values or descriptors in Rust
- expose per-read options and `MAP<STRING, BLOB>` descriptors as a native Go map
- keep copied keys and descriptors valid after the Arrow record is released
- reuse the existing `BlobReader`: `OpenBlob` streams and seeks; `ReadBlobs` batches with range merging

`NewReadBuilderWithOptions` exposes the existing C API and is equivalent to Java/PyPaimon's `table.copy(options).newReadBuilder()`.

This is read support only. Blob storage IO, descriptor parsing, merge limits, concurrency, and DLF data-token refresh remain in Rust core. Null map keys cannot be represented by Arrow maps and return an error, matching PyPaimon's Arrow reader.

## Go usage

```go
builder, err := table.NewReadBuilderWithOptions(map[string]string{
    "blob-as-descriptor": "true",
})
// scan/read an Arrow record, then:
descriptors, err := paimon.StringBlobMapDescriptors(record.Column(mapColumn), row)
record.Release()
for key, descriptor := range descriptors {
    if descriptor == nil {
        continue
    }
    stream, err := blobReader.OpenBlob(descriptor)
    // io.Copy(destinationFor(key), stream)
}
```

Use `OpenBlob` for large values. Pass descriptors for small values to one `ReadBlobs` call to use range merging.

## Verification

- `cargo test -p paimon --lib blob`: 128 passed
- `cargo build -p paimon-c --release`
- Go writer-generated table fixture: `NewReadBuilderWithOptions` → Rust decoder → C Data → native Go map → `ReadBlobs`
- focused Go BLOB tests and `go vet ./...`
- real DLF E2E in regression `morax_test`: PyPaimon wrote a `MAP<STRING, BLOB>` table; Go traversed descriptors after releasing Arrow records, streamed and checksummed a 10,485,883-byte BLOB, sought a 4 KiB range, and batch-read multi-file, empty, and duplicate descriptors


crates/paimon/src/arrow/format/blob.rs
```diff
@@ -764,6 +995,277 @@ fn build_blob_array_descriptors(
...
+                        })?;
+                        Ok(Bytes::from(
+                            BlobDescriptor::new(file_path.to_string(), offset, length).serialize(),
+                        ))
+                    })
+                    .transpose()
+            })
+            .collect::<crate::Result<Vec<_>>>()?
+    } else {
+        let payload = read_blob_entry(reader, blob_entry_range(&payload_range)).await?;
```
QuakeWang, line 1319: Please reject oversized inline payloads before this read. total_value_length is a u64, but these values are later collected into an Arrow BinaryArray, whose 32-bit offsets panic once the total byte length exceeds i32::MAX. ARRAY<BLOB> already preflights this case; MAP should do the same and also validate the flattened batch total.

bindings/go/tests/blob_reader_test.go
```diff
@@ -101,6 +104,61 @@ func TestBlobReaderReadBlobAndBatch(t *testing.T) {
 	}
 }
 
+func TestStringBlobMapDescriptors(t *testing.T) {
```
QuakeWang, line 105: Could this test cover the actual read path instead of starting from a hand-built Arrow map? As written, it only tests descriptor extraction and BlobReader. A writer-generated fixture read through NewReadBuilderWithOptions would also cover the Rust decoder, C Data conversion, and Go API.

docs/src/go-binding.md
```diff
@@ -147,8 +147,8 @@ catalog, err := paimon.NewCatalog(map[string]string{
     "token.provider":             "dlf",
     "dlf.region":                 os.Getenv("DLF_REGION"),
     "dlf.oss-endpoint":           os.Getenv("DLF_OSS_ENDPOINT"),
-    "dlf.token-loader":           "ecs",
-    "dlf.token-ecs-role-name":    os.Getenv("DLF_ECS_ROLE"),
+    "dlf.access-key-id":          os.Getenv("DLF_ACCESS_KEY_ID"),
```
QuakeWang: This changes the DLF example from the ECS RAM-role loader to static credentials, which is unrelated to MAP<K, BLOB> support and drops automatic credential refresh. Please revert it here or move it to a separate PR with its own rationale.

crates/paimon/src/arrow/format/blob.rs
XiaoHongbo-Hope: Fixed in cd19225. Inline MAP values are now rejected before the complete entry read, and flattened key/value batches enforce Arrow i32 offset limits. Added sparse >2 GiB regression coverage.

bindings/go/tests/blob_reader_test.go
XiaoHongbo-Hope: Fixed in cd19225. The Go test now reads a PyPaimon-generated table through NewReadBuilderWithOptions, Rust decoding, C Data import, Go map extraction, and ReadBlobs.

docs/src/go-binding.md
XiaoHongbo-Hope: Fixed in cd19225. Restored the ECS RAM-role loader example with dlf.token-loader and dlf.token-ecs-role-name.

QuakeWang (approved):
LGTM

PR #785, opened 2026-09-02
fix(deletion_vector): read Java bitmap64 deletion vectors

jackylee-ch:
A deletion vector written by Java in the 64-bit format (`Bitmap64DeletionVector`)
cannot be read: `read_from_bytes` accepts only the v1 magic, so every scan touching
such a file fails with `Invalid magic number`, naming the v1 magic alone.
Java's `DeletionVector.read` dispatches on both.

Two things differ beyond the value. The v2 magic is written **little-endian**, and
`DeletionFile.length()` counts the length prefix and the CRC for v2 but neither for
v1. The second also breaks the read range: the factory asked for `length + 8`, past
the end of a v2 entry, so a vector ending its index file was rejected by the storage
layer before parsing.

**Fix**: detect either magic, apply each format's length convention, and decode the
payload's outer layer here rather than through `RoaringTreemap::deserialize_from` —
that decoder validates each inner roaring32 bitmap but inserts bucket keys straight
into a map, so a duplicate key silently replaces a bucket and its deletes vanish.
Java rejects that in `OptimizedRoaringBitmap64.readBitmapCount` / `readKey`, and the
same checks are applied: count range, key bounds, strictly ascending. A duplicate
key, a descending key, a bogus count or a truncated payload now error instead of
decoding partially. Java run-length encodes before writing and emits every bucket
key densely, so tests cover both shapes.

Read side only; the writer still emits v1. Positions above `u32::MAX` are rejected
rather than truncated — every API here is already roaring32-bound, and reaching one
needs a data file with over 4.29e9 rows. A negative `bitmapLength` read from the
file now errors instead of wrapping the size guard into passing.

### DIFF
--- a/crates/paimon/src/deletion_vector/core.rs
+++ b/crates/paimon/src/deletion_vector/core.rs
@@ -32,7 +32,15 @@ pub struct DeletionVector {
 /// Magic number for BitmapDeletionVector serialization format
 /// Same as Java: 1581511376
 const MAGIC_NUMBER: u32 = 1581511376;
-const MAGIC_NUMBER_SIZE_BYTES: usize = 4;
+/// Magic number for Java `Bitmap64DeletionVector` (v2). Written little-endian,
+/// unlike the v1 magic, so the two are told apart by byte order as well as value.
+const MAGIC_NUMBER_64: u32 = 1681511377;
+const MAGIC_NUMBER_SIZE_BYTES: u64 = 4;
+/// Size of the `bitmapLength` prefix and of the trailing CRC. Mirrors Java
+/// `Bitmap64DeletionVector.LENGTH_SIZE_BYTES` / `CRC_SIZE_BYTES`; a v2 entry
+/// counts both in `DeletionFile.length()` while a v1 entry counts neither.
+const LENGTH_SIZE_BYTES: u64 = 4;
+const CRC_SIZE_BYTES: u64 = 4;
 
 impl DeletionVector {
     /// Create a new empty DeletionVector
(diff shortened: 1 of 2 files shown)


crates/paimon/src/deletion_vector/core.rs
```diff
@@ -139,59 +156,117 @@ impl DeletionVector {
...
+    ///
+    /// Row positions are offsets inside a single data file, and every existing API
+    /// here is already roaring32-bound -- [`Self::is_deleted`] documents that
+    /// positions above `u32::MAX` cannot be present, and `to_bitmap` hands a
+    /// `RoaringBitmap` to the writer. A position that does not fit is therefore
+    /// rejected rather than truncated, so a 64-bit vector can never silently lose
+    /// deletes; a data file with more than `u32::MAX` rows would be needed to
+    /// reach it.
+    fn from_bitmap64_bytes(bitmap_data: &[u8]) -> crate::Result<Self> {
+        let treemap = roaring::RoaringTreemap::deserialize_from(bitmap_data).map_err(|e| {
```
JingsongLi: [P2] Reject duplicate or out-of-order bitmap64 bucket keys before they can drop deletes

`roaring::RoaringTreemap::deserialize_from` validates each inner roaring32 bitmap, but its outer decoder simply inserts each `u32` key into a `BTreeMap`; it does not enforce Java `OptimizedRoaringBitmap64.deserialize`'s requirement that keys are strictly increasing. A duplicate key therefore overwrites the earlier bucket silently. If an index is corrupted so two buckets carry key 0, deleted positions from the first bucket reappear as live rows instead of the vector being rejected.

I verified this with a two-bucket payload containing key 0 / bitmap {1} followed by key 0 / bitmap {2}: this method succeeds and returns only {2}; the equivalent Java reader rejects the second key. Please decode/validate the outer count and keys explicitly (including nonnegative/range/order checks) before assembling the treemap, and add duplicate plus descending-key cases.

jackylee-ch:
Confirmed. Outer layer is decoded here now, mirroring `readBitmapCount`/`readKey`: count range, key bounds, strictly ascending. Duplicate or descending keys error instead of dropping deletes.

JingsongLi (approved):
+1

PR #786, opened 2026-09-02
fix(python): render field types as SQL strings, not Rust Debug

jackylee-ch:
`PyDataField.field_type()` returns Rust `Debug` output — a column declared
`VARCHAR(10)` reads back as `Varchar(VarCharType { length: 10, nullable: true })`.
The in-place `TODO(#284 follow-up)` names the cause: `DataType` has no enum-level
`Display`, so there was nothing to call.

Only 10 of the 23 variants carried a leaf `Display`. This adds the other nine
scalars and the four constructed types, then the dispatch. The constructed formats
fill Java's `FORMAT` with the children's own SQL strings (`ARRAY<%s>`,
`MAP<%s, %s>`, `MULTISET<%s>`, `ROW<%s>`), so a child's `NOT NULL` sits inside the
brackets and the outer one after them.

`ROW` renders fields like `DataField.asSQLString`: the escaped name, the type, then
`COMMENT '...'`. So a nested row reads ``ROW<`id` INT, `name` VARCHAR(10)>``.

`spec::escape_identifier` was not Java's `EncodingUtils.escapeIdentifier` — it only
doubled `"` and added no delimiter — so it is aligned here rather than worked
around. It had no production caller before this PR (`Display for DataField` was
never written), and its test pinned the wrong shape.

The dispatch also retires `VectorType::element_sql_name`, a private table of seven
primitive names with `unreachable!()` for the rest.

`cargo test -p pypaimon_rust` cannot link libpython here, so the binding is
compile-checked locally and its behaviour left to `integration (python)`. The
rendering is covered by unit tests in `paimon`.

### DIFF
--- a/crates/paimon/src/spec/types.rs
+++ b/crates/paimon/src/spec/types.rs
@@ -1838,6 +1838,16 @@ pub struct RowType {
     fields: Vec<DataField>,
 }
 
+/// Quote a row field name the way Java `EncodingUtils.escapeIdentifier` does:
+/// wrap it in backticks and double any backtick it contains.
+///
+/// `spec::escape_identifier` is *not* this function -- it only doubles `"` and adds
+/// no delimiters -- so it cannot be used here. Aligning it instead would be the
+/// tidier fix, but it lives in `schema.rs`, which two open PRs are rewriting.
+fn escape_row_field_name(name: &str) -> String {
+    format!("`{}`", name.replace('`', "``"))
+}
+
 /// Render one row field the way Java `DataField.asSQLString` does: the escaped
 /// name, a space, the field type's SQL string, then `COMMENT '...'` when the
 /// field carries a description. Java also appends `DEFAULT <value>`, which has no
@@ -1846,7 +1856,7 @@ fn write_row_field(f: &mut Formatter<'_>, field: &DataField) -> std::fmt::Result
     write!(
         f,
         "{} {}",
-        crate::spec::escape_identifier(field.name()),
+        escape_row_field_name(field.name()),
         field.data_type()
     )?;
     if let Some(description) = field.description().filter(|text| !text.is_empty()) {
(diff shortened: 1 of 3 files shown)


crates/paimon/src/spec/types.rs
```diff
@@ -1734,6 +1838,79 @@ pub struct RowType {
...
 
+/// Render one row field the way Java `DataField.asSQLString` does: the escaped
+/// name, a space, the field type's SQL string, then `COMMENT '...'` when the
+/// field carries a description. Java also appends `DEFAULT <value>`, which has no
+/// counterpart on this struct.
+fn write_row_field(f: &mut Formatter<'_>, field: &DataField) -> std::fmt::Result {
+    write!(
+        f,
+        "{} {}",
+        crate::spec::escape_identifier(field.name()),
```
JingsongLi: [P2] Quote ROW field names with Java's identifier syntax

This helper does not implement Java `EncodingUtils.escapeIdentifier`: the Rust function only doubles `"` characters and does not add delimiters, while Java returns a backtick-wrapped identifier and doubles embedded backticks. Consequently this renders even `id` as `ROW<id INT>` instead of Java's backtick-wrapped `id`; names containing spaces or reserved words become ambiguous, and a name containing a backtick is not escaped at all. That also means the new Python `field_type()` contract is not the claimed Java `asSQLString()` for nested rows.

Please use a ROW-specific helper that wraps the name in backticks and replaces each embedded backtick with two backticks (or deliberately align the existing helper if all its callers need Java syntax), and update the tests to cover a normal name, a reserved/space-containing name, and an embedded backtick.

jackylee-ch:
Right — `spec::escape_identifier` only doubled `"`. Aligned it with Java's backtick syntax instead of adding a ROW-local escape: it had no production caller and its test pinned the wrong shape.

JingsongLi (approved):
+1

PR #790, opened 2026-09-03
build: prepare the 0.4.0 release

XiaoHongbo-Hope:
## Summary

- update the release status, 0.3.0 history, and 0.4.0 highlights
- protect the repository's actual `release-*` branch naming convention
- require source archives to come from a signed release tag whose version matches the workspace
- align Rust, Python, and Go RC naming and reject `rc0`/leading-zero RC numbers
- validate Go manual inputs against the source workspace and pass them safely into Bash
- make Go tag/Release publication resumable while binding retries to the original annotated tag object and source commit
- mark Go RCs as pre-releases and keep Go convenience releases from replacing the root project's Latest release
- replace the yanked `chacha20` 0.10.1 lock entry with 0.10.2 and regenerate affected dependency reports

## Validation

- `PYO3_PYTHON=python3.11 cargo +1.94.0 check --locked --workspace --all-targets --all-features`
- `cargo +1.94.0 package --locked -p paimon -p paimon-datafusion --all-features`
- `cargo +1.94.0 publish --locked -p paimon --all-features --dry-run`
- `python3.11 scripts/dependencies.py verify`
- `python3.11 scripts/dependencies.py check`
- `python3.11 -m unittest discover -s scripts/tests -p 'test_*.py'`
- `python3 scripts/validate_asf_yaml.py`
- parsed all workflow YAML and syntax-checked explicit Bash steps
- simulated Go retry recovery against `bindings/go/v0.3.0-rc1`, including tag object, parent, path allowlist, and all four packaged library verifiers
- verified the latest `main` CI run is green

No release tag or external package publication is performed by this PR.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -47,7 +47,7 @@ github:
       branches:
         includes:
           - "~DEFAULT_BRANCH"
-          - "release/*"
+          - "release-*"
           - "rel/*"
         excludes: []
       bypass_teams:
--- a/.github/workflows/release-go-binding.yml
+++ b/.github/workflows/release-go-binding.yml
@@ -38,6 +38,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ inputs.version || github.ref_name }}
+  cancel-in-progress: false
+
 jobs:
   validate:
     runs-on: ubuntu-latest
@@ -46,6 +50,7 @@ jobs:
       source_ref: ${{ steps.meta.outputs.source_ref }}
       tag: ${{ steps.meta.outputs.tag }}
       source_sha: ${{ steps.meta.outputs.source_sha }}
+      tag_object: ${{ steps.meta.outputs.tag_object }}
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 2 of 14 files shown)


JunRuiLee:
@XiaoHongbo-Hope Could the 0.4.0 cut wait for #771?

 #771 is close: the two review comments on it are fixed and pushed, and the API shape was settled with the reviewer offline today — the search returns the selected splits and an ordinary read consumes them, mirroring Java's IndexedSplit. I'm finishing that rework now.

The reason it matters for 0.4.0: on main, plan_for_bucket_vector_splits and the BucketVectorSearchSplit decoder have no caller outside their own tests — #771 is the entry point that reaches them. So as it stands 0.4.0 would ship the plumbing for engine-planned primary-key vector search with no way to use it, and the usable API would only arrive in the next release.

XiaoHongbo-Hope:
> @XiaoHongbo-Hope Could the 0.4.0 cut wait for #771?
> 
> #771 is close: the two review comments on it are fixed and pushed, and the API shape was settled with the reviewer offline today — the search returns the selected splits and an ordinary read consumes them, mirroring Java's IndexedSplit. I'm finishing that rework now.
> 
> The reason it matters for 0.4.0: on main, plan_for_bucket_vector_splits and the BucketVectorSearchSplit decoder have no caller outside their own tests — #771 is the entry point that reaches them. So as it stands 0.4.0 would ship the plumbing for engine-planned primary-key vector search with no way to use it, and the usable API would only arrive in the next release.

Sure.

JingsongLi:
**[P2] Support Go release recovery when rerunning only failed jobs**

Location: [`.github/workflows/release-go-binding.yml:321–323`](https://github.com/apache/paimon-rust/blob/9a5775c784ca21992f9cbb7d66a14005c657f1ba/.github/workflows/release-go-binding.yml#L321-L323)

On the first run, `validate` outputs an empty `tag_object` because the Go release tag does not exist yet. If `publish` successfully pushes the tag but then fails while creating the GitHub Release, selecting **Re-run failed jobs** reruns `publish` while retaining the successful `validate` job's outputs. The new checkout now contains the tag, but `EXPECTED_TAG_OBJECT` is still empty, so this comparison fails with `tag ... changed after validation` before reaching the recovery branch.

This leaves a common retry path unsupported by the new recovery logic. Rerunning all jobs or starting a new manual run works around it, but rerunning only the failed `publish` job cannot recover.

Could tag creation and GitHub Release publication be split into separate jobs, with the latter consuming the successful tag job's object SHA? That would allow the failed Release job to resume while preserving the strict tag-object binding.

I reproduced the guard failure using the actual tag step in an isolated local Git repository with a local bare remote. The artifact verifier was stubbed; the failure occurs before it is invoked.

JingsongLi (approved):
+1

PR #792, opened 2026-09-03
fix(table): match signed zero keys in scalar bitmap index equality

jackylee-ch:
`-0.0` and `+0.0` are separate dictionary keys — the sign bit survives
serialization and the comparator orders them apart — but equality consults only the
literal's own key. The scalar row-level filter compares through Arrow's IEEE
kernel, where the two are equal, so on a file holding `-0.0` the index answers
`WHERE d = 0.0` with an empty candidate set, and that is `Some(empty)` rather than
a decline: every row is pruned before the filter that would have kept it runs.
`In` and the bitmap file-level min/max check share the flaw.

**Scalar equality only.** `arrow::residual` compares array *elements* bitwise,
matching Java (`CompareUtils.compareLiteral` → `Float.compareTo`), so `-0.0` is
genuinely not an occurrence of `+0.0` there and `ArrayContains` / `ArraysOverlap` /
`ArrayContainsAll` stay exact — pinned by
`test_array_membership_uses_java_nan_and_signed_zero_semantics`. The sorted (BTree)
kind splits the two zeros the same way in its own key serde; that is a separately
persisted contract and is left to a follow-up.

Note this makes the index agree with Rust's scalar filter, not with Java: Java's
`Equal` also goes through `Float.compareTo`, so `WHERE f = 0.0` matching `-0.0`
rows is a pre-existing Rust-only behaviour. Happy to file that separately if you
would rather converge on Java and drop this widening instead.


crates/paimon/src/table/global_index_scanner/entry.rs
```diff
@@ -169,10 +171,21 @@ pub(super) fn bitmap_meta_may_match(
...
     }
+    if meta.may_match(op, serialized_literals, cmp) {
+        return true;
+    }
+    // A `±0.0` literal also matches rows holding the other signed zero, but the
+    // comparator orders `-0.0` below `+0.0`, so a key range that stops at one of
+    // them would prune a file the reader would have found. `Eq` looks at the first
+    // literal only, so the alternates are probed as a call of their own rather
+    // than appended.
+    let alternate_zeros: Vec<Vec<u8>> = serialized_literals
```
JingsongLi, line 197: [P1] Apply signed-zero expansion to multivalue metadata pruning too

This expansion only affects `GlobalIndexFileKind::Bitmap`. Array predicates are routed to `GlobalIndexFileKind::Multivalue`, whose file-level check still goes through `multivalue_meta_may_match` below using only the original literal. For a multivalue file whose min/max is `-0.0`, `ARRAY_CONTAINS(+0.0)` (and similarly `ARRAYS_OVERLAP` / `ARRAY_CONTAINS_ALL`) is therefore rejected before `BitmapGlobalIndexReader::equal` gets a chance to union the opposite zero. The scanner then treats the indexed range as covered and returns no candidates, so matching rows are lost.

I verified this by adding the same `[-0.0, -0.0]` metadata case to `test_signed_zero_survives_meta_prune` through `multivalue_meta_may_match`; it fails on `ARRAY_CONTAINS` for `Float`. Please apply the alternate-zero logic inside the multivalue Eq/In checks as well and cover all three array operators.

jackylee-ch:
Java compares array elements with `Float.compareTo` (`CompareUtils.compareLiteral`), so `-0.0` is not an occurrence of `+0.0` — `residual.rs` pins that. Narrowed to scalar Eq/In; arrays stay exact.

crates/paimon/src/table/bitmap_global_index_reader.rs
```diff
@@ -270,11 +276,31 @@ impl BitmapGlobalIndexReader {
...
+    /// deliberately *bitwise* in `arrow::residual` (Java compares elements with
+    /// `Float.compareTo`), so the array operators must not widen.
+    async fn equal_including_signed_zero(
+        &self,
+        key: &[u8],
+        data_type: &DataType,
+    ) -> io::Result<RoaringTreemap> {
+        let mut result = self.equal(key, data_type).await?;
+        if let Some(other_zero) = opposite_zero_key(key, data_type) {
+            result |= self.equal(&other_zero, data_type).await?;
```
QuakeWang, line 299: Scalar `Eq`/`In` residuals use Arrow 58.4.0's `eq`, which distinguishes signed zeros. This widening only adds candidates the residual rejects. Could you reproduce the claimed false negative with a test that actually runs the residual filter?

jackylee-ch:
You're right — measured it: `evaluate_exact_leaf_predicate` on `[0.0, -0.0]` with `Eq(+0.0)` gives `[true, false]`. Arrow's `eq` is bitwise, so there is no false negative. My IEEE claim was wrong.

jackylee-ch:
Closing: the premise does not hold, see above.

PR #793, opened 2026-09-04
fix(c): align append vector filter with core behavior

QuakeWang:
### Purpose

The C binding test still expected scalar filters on data-evolution vector searches to be rejected, although the Rust core now supports applying them before Top-K.

### Brief change log

  - Update the C vector filter documentation to match the current contract.
  - Replace the obsolete error assertion with a filtered append-table read test.
  - Verify that the C result matches the Rust core and refills the requested Top-K result.

### Tests

  - `cargo test -p paimon-c --lib`
  - `cargo clippy -p paimon-c --all-targets -- -D warnings`
  - `cargo +1.94.0 check --locked --workspace --all-targets --all-features`

### Documentation

Updated the inline C API documentation.

### DIFF
--- a/bindings/c/src/tests.rs
+++ b/bindings/c/src/tests.rs
@@ -3263,32 +3263,48 @@ fn vector_search_pk_filter_excludes_neighbor() {
 }
 
 #[test]
-fn vector_search_append_filter_returns_invalid_input() {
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #796, opened 2026-09-07
feat(go): support case-insensitive column names on reads

jackylee-ch:
### Purpose

Follow-up to #496, which added this read-time switch to the core, the C ABI and
Python and left "Go's own opt-in case-insensitive API" to a follow-up. Measured
from Go, fixture columns lowercase:

```
NewRead() after WithProjection([ID NAME]) -> Column ID does not exist in table ...
pb.Eq("ID", 1)                            -> Column 'ID' not found in schema fields ["id", "name"]
```

Neither half is reachable from Go, and the first message names the column absent
when only its case differs.

### Brief change log

One switch name, `WithCaseSensitive`, on each object that owns a name resolution:
`ReadBuilder` for projection, `PredicateBuilder` for predicates. The split is
core's and the C ABI's, not Go's: the core resolves a column when the predicate is
built, so a read-builder flag cannot reach an already-built one. Python needs a
single switch only because its filter is a dict converted at `with_filter` time,
while Go's `WithFilter` takes a resolved handle.

`Table` gains no method; `PredicateBuilder.WithCaseSensitive` returns a copy. The
ten predicates Go exposes now call the additive
`paimon_predicate_*_with_case_sensitive` entry points, the default passing `true`.
The six operators C gained in #523 but Go never bound stay out of scope.

### Tests

Asserted in both directions: unset and `true` still reject uppercase names,
`false` resolves them, records come back with the schema's own spelling, and both
call orders agree. Each of the three predicate argument shapes has a case, and one
test pins that neither switch reaches the other.

### API and Format

Two new Go methods, one name. No ABI, wire or storage change; the C symbols
already exist.

### Documentation

One `go-binding.md` section per half.

### DIFF
--- a/bindings/go/predicate.go
+++ b/bindings/go/predicate.go
@@ -339,7 +339,7 @@ func (pb *PredicateBuilder) NotIn(column string, values ...any) (*Predicate, err
 
 // buildLeafPredicate is a helper for comparison predicates that take (table, column, datum).
 func (pb *PredicateBuilder) buildLeafPredicate(
-	ffiVar *leafPredicateFFI,
+	ffiVar *FFI[func(*paimonTable, *byte, paimonDatumC, bool) (*paimonPredicate, error)],
 	column string, datum Datum,
 ) (*Predicate, error) {
 	t := pb.table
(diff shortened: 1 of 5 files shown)


bindings/go/predicate.go
```diff
@@ -253,8 +253,14 @@ var errConsumedPredicate = fmt.Errorf("paimon: predicate already consumed or nil
...
 // so there is no Close() method.
+//
+// caseSensitive is baked into every predicate this builder produces: the core
+// resolves a column when the predicate is constructed, so it cannot be changed
+// afterwards. Use Table.PredicateBuilder for exact matching or
+// Table.PredicateBuilderWithCaseSensitive to opt out.
 type PredicateBuilder struct {
-	table *Table
+	table         *Table
+	caseSensitive bool
```
JingsongLi, line 258: Why we need to add this API? Only for GO?

jackylee-ch: Rust/C support this since #496. Go needs the flag at predicate construction, before WithFilter.

jackylee-ch: Not Go-only: the split is core's and the C ABI's, both merged in #496, whose description left Go's half to a follow-up. Python needs a single switch only because its filter is a dict converted at `with_filter` time; Go's `WithFilter` takes an already-resolved handle, so the flag has to sit on the builder that produces the predicate.

Collapsed it into `PredicateBuilder.WithCaseSensitive` in 23e53fa so both halves share one switch name and `Table` gains no method. `go test ./...` in bindings/go green, including a new test pinning that neither switch reaches the other.

JingsongLi (approved):
+1

PR #798, opened 2026-09-09
perf(vindex): enable 8-way multipart uploads for index files

jerry-024:
### Purpose

Reduce the vector-index serialization/upload bottleneck by allowing up to eight concurrent multipart uploads, while keeping each chunk at 8 MiB.

This is a standalone upload-only change. The existing single remote scan, training, and `add` flow remain unchanged.

### Brief change log

- Extract an internal `async_writer_with_concurrency()` helper. Existing `async_writer()` callers retain concurrency 1 and 8 MiB chunks.
- Use concurrency 8 for vector-index output directly, without an MVP environment switch.
- Add round-trip, out-of-order multipart completion, and index-build failure tests through the existing `FileIOProvider` test entry point. Reuse the existing build failure handling.

### Tests

Coverage added or updated; runtime acceptance is still pending:

- [x] `cargo fmt --all -- --check`.
- [x] `git diff --no-ext-diff --check`.
- [ ] Memory/fs round trips at concurrency 1 and 8: write nine full 8 MiB chunks plus a 12,345-byte tail, then verify length and every byte after shutdown.
- [ ] Exercise OpenDAL's multipart scheduler with deterministic out-of-order part completion and verify the final content.
- [ ] Run existing FileIO, cache-invalidation, and vector-index build tests. The cache-invalidation test now exercises concurrency 8.
- [ ] Inject both a part-write failure and a final close failure: require a build error, unchanged committed index metadata, unchanged existing index bytes and query results, and cleanup of newly completed shard files.
- [ ] Run the ignored `streaming_upload_oss` test against real OSS multipart storage and verify scratch-object cleanup. It requires `PAIMON_CATALOG_OPTIONS`, `PAIMON_UPLOAD_TEST_DATABASE`, and `PAIMON_UPLOAD_TEST_TABLE`.
- [ ] Complete three interleaved runs per variant on vdb2: default upload versus 8-way upload, with the same source baseline, PR90 core, 10M data, index parameters, and queries. Record source/binary checksums, total build and serialization/upload time, peak memory, index size, and Recall; retain all runs and check an absolute Recall difference of at most 0.005.

The vdb2 verification attempt stopped during dependency download, before compilation, because the Cargo registry connection timed out. Runtime correctness and performance acceptance have not been established; no measured speedup is claimed here.

### API and Format

No public API or storage-format changes. Upload concurrency changes only for vector-index output.

Known existing limitation: the `AsyncWrite` adapter does not expose multipart abort. The failure test explicitly records one unfinished upload and zero abort calls, separately from completed-object cleanup. This patch does not add multipart-abort support; real OSS failure cleanup still needs verification.

### Documentation

No new user-facing configuration. No documentation changes are included.

### DIFF
--- a/crates/paimon/src/io/file_io.rs
+++ b/crates/paimon/src/io/file_io.rs
@@ -838,11 +838,18 @@ impl OutputFile {
 
     /// Get an async streaming writer for format-level writes (e.g. parquet).
     pub(crate) async fn async_writer(&self) -> crate::Result<Box<dyn AsyncFileWrite>> {
+        self.async_writer_with_concurrency(1).await
+    }
+
+    pub(crate) async fn async_writer_with_concurrency(
+        &self,
+        concurrency: usize,
+    ) -> crate::Result<Box<dyn AsyncFileWrite>> {
         let (op, relative_path, cache_path) = self.resolve().await?;
         let writer: Box<dyn AsyncFileWrite> = Box::new(
             op.writer_with(&relative_path)
                 .chunk(8 * 1024 * 1024)
-                .concurrent(1)
+                .concurrent(concurrency)
                 .await?
                 .into_futures_async_write()
                 .compat_write(),
@@ -859,6 +866,9 @@ impl OutputFile {
         }))
     }
 }
+
+#[cfg(test)]
+pub(crate) mod multipart_test;
 
 #[cfg(test)]
 mod file_action_test {
@@ -1937,7 +1947,7 @@ mod input_output_test {
         let mut writer = file_io
             .new_output(&location)
             .unwrap()
-            .async_writer()
+            .async_writer_with_concurrency(8)
             .await
             .unwrap();
         writer.write_all(b"new metadata").await.unwrap();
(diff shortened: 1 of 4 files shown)


jerry-024:
This PR should remain on hold until paimon-rust can use an official OpenDAL release containing [apache/opendal#8224](https://github.com/apache/opendal/pull/8224) (or an equivalent backport).

The change from concurrency 1 to 8 exposes a memory regression in the currently used OpenDAL implementation. Concurrent task results retain their original retry input, including the upload payload, even after a part succeeds. The multipart writer uses a hard-coded prefetch of 8192, so completed payloads can accumulate while more parts are submitted. Eight concurrent network uploads therefore do not bound retained payloads to eight 8 MiB parts, and large index builds can run out of memory. The previous concurrency-1 path does not accumulate completed inputs this way.

The upstream fix releases successful task inputs before result collection while retaining failed inputs for retry, preserving concurrent uploads. The upstream PR is merged, but merging upstream is not sufficient for this PR: the selected published dependency must actually contain the fix. Once that release is available, update the dependency and verify buffer release, retry correctness, and peak memory during a large 8-way upload before merging.

PR #801, opened 2026-09-10
fix(table): check batch arity instead of asserting it in debug builds

JunRuiLee:
### Purpose

Three single-query terminals in `vector_search_builder.rs` call a batch entry point with one query and then take `results.remove(0)` behind a `debug_assert_eq!(results.len(), 1)`.

The assert is compiled out of release builds, which leaves two silent failure modes there:

- an **empty** vector panics on an index out of bounds, with nothing indicating what went wrong;
- a **longer** vector returns the FIRST result — pairing the caller's single query with another query's answer, which no caller can detect.

A wrong length means the batch ran the wrong number of searches. That is a programming error in this crate rather than bad input, so the error kind stays the same in spirit; only its enforcement moves from debug-only to always.

### Brief change log

- Add `take_only_result`, which reports both a missing and an extra result as `UnexpectedError`, naming the operation.
- Use it at the three pre-existing sites: `VectorSearchBuilder::execute`, `plan_and_search_pk_candidates`, and the test-only `evaluate_vector_search`.

### Tests

- `take_only_result_rejects_bad_batch_arity` — the one-element, empty, and two-element cases.

### API and Format

No public API or storage-format change. `take_only_result` is private to the module.

### Documentation

None needed.

### DIFF
--- a/crates/paimon/src/table/vector_search_builder.rs
+++ b/crates/paimon/src/table/vector_search_builder.rs
@@ -226,6 +226,33 @@ fn same_vector_search_table(left: &Table, right: &Table) -> bool {
         && left.branch() == right.branch()
 }
 
+/// Unwrap a single-query result from a batch entry point that must return exactly
+/// one element per input query.
+///
+/// The batch terminals below are handed one query, so their result vector holds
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #802, opened 2026-09-10
fix(vindex): bound a deletion-vector position by its own source file

JunRuiLee:
### Purpose

An ANN segment's source files share one ordinal space — a position is mapped to `global = file_offset + position`. `build_live_row_ids` bounds the residual predicate's positions against the source file they belong to, but not the deletion vector's.

A deletion-vector position past its own file's row count therefore does not fall *outside* the mask. It lands inside the **next** source file's ordinal range and deletes one of THAT file's rows. The mask comes back well formed, so nothing downstream can tell that a row of an unrelated file was dropped.

### Brief change log

- Bound each deletion-vector position by its own source file's row count in `build_live_row_ids` (`crates/paimon/src/vindex/pkvector/ann.rs`), with the same message shape the residual path already uses.

The check is `>= row_count`, so a file's last valid position still deletes its own row.

### Tests

- `a_deletion_vector_position_past_its_own_file_is_refused` — two files of 3 and 2 rows; a deletion vector on the first naming position 3 would otherwise delete the second file's first row. Also pins the boundary in the other direction: position 2 still deletes the first file's own last row and leaves the second file untouched.
- Mutation-checked: replacing the new condition with `false` fails this test and nothing else.

### API and Format

No public API or storage-format change. A malformed deletion vector that used to corrupt the mask silently is now refused as `DataInvalid`.

### Documentation

None needed.

### DIFF
--- a/crates/paimon/src/vindex/pkvector/ann.rs
+++ b/crates/paimon/src/vindex/pkvector/ann.rs
@@ -121,6 +121,18 @@ pub(crate) fn build_live_row_ids(
         if active {
             if let Some(dv) = deletion_vectors.get(source_file.file_name()) {
                 for position in dv.iter() {
+                    // Bound the position against ITS OWN source file, as the residual
+                    // path above already does. Source files share one ordinal space
+                    // (`global = file_offset + position`), so a position past this
+                    // file's rows does not fall out of the mask -- it lands inside the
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #803, opened 2026-09-10
fix(lumina): decide on the index size before parsing the metric

JunRuiLee:
### Purpose

Both Lumina search paths read `index_meta.metric()` before asking the index how many vectors it holds. An index holding nothing has an empty answer whatever the metric says, so a segment whose metadata names an unknown metric reports a malformed-metric error for a query that could not have been affected by it.

Java runs these in the other order (`LuminaVectorGlobalIndexReader`, both paths): `index.size()` decides first, the filter is built next, and the metric is read only on the way out.

### Brief change log

- Scalar path (`search_lumina`): `get_count()` / `effective_k == 0` now decide before `index_meta.metric()`.
- Batch path (`search_lumina_batch`): same reordering, and the empty include-set short-circuit now reads the set directly instead of the list built from it, so a filter that permits nothing makes no native call at all — which it already did, one step later.
- `if count == 0` returning early on the batch path is a step the `effective_k == 0` check below already covers; it is there so an empty index does not build the dense filter first, and it is what the metric ordering is stated against. It has no separately observable behaviour on this tree, so it carries no test of its own — noted rather than left implicit.

### Tests

Two tests, one per path, each driving an index of size 0 whose metadata names a metric Lumina does not define, and asserting the call reports **no hits** rather than the parse error:

- `an_empty_index_reports_empty_before_the_metric_is_parsed` (scalar)
- `an_empty_index_reports_empty_before_the_filter_is_densified` (batch; also asserts the index size is the only native call made)

Both were mutation-checked by restoring the old order individually — each mutation fails exactly its own test.

### API and Format

No public API or storage-format change. A query against an empty index whose metric metadata is unparseable now returns an empty result instead of an error.

### Documentation

None needed.

### DIFF
--- a/crates/paimon/src/lumina/reader.rs
+++ b/crates/paimon/src/lumina/reader.rs
@@ -363,12 +363,16 @@ fn search_lumina<S: LuminaSearch + ?Sized>(
     }
 
     let limit = vector_search.limit;
-    let index_metric = index_meta.metric()?;
+    // Java's order: `index.size()` decides first, and the metric is parsed only once
+    // there is something to search. An empty index has no hits to rank, so reading
+    // the metric ahead of the size would report a malformed metric for a query whose
+    // answer is empty whatever the metric says.
     let count = searcher.get_count()? as usize;
     let effective_k = std::cmp::min(limit, count);
     if effective_k == 0 {
         return Ok(None);
     }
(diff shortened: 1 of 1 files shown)


JingsongLi (approved):
+1

PR #806, opened 2026-09-10
fix(file_index): skip pruning for narrowing integer schema changes

QuakeWang:
### Purpose

Follow-up to #783. FileIndex predicate remapping checked literal conversion but allowed narrowing integer schema changes. An old INT value such as 383 becomes NULL when read as TINYINT, so pushing IS NULL into the old index can incorrectly discard matching rows.

### Brief change log

- Reject index predicate remapping for narrowing integer changes, matching Java's fallback rule.
- Preserve safe integer widening and add regression tests for predicate composition and index-enabled/disabled result equivalence.

### Tests

### API and Format

No changes.

### Documentation

No changes needed.


JingsongLi (approved):
+1

PR #812, opened 2026-09-11
feat(rest): add partition registration, lookup and statistics APIs

sundapeng:
### Purpose

Split out of #591 so it can be reviewed on its own. This adds the REST catalog partition APIs that the catalog-managed Format Table work builds on: registering partitions with optional statistics, unregistering them, looking them up by name, and listing them by predicate or by name pattern. Nothing reachable from SQL changes.

### Brief change log

- `RESTApi`:
  - `create_partitions` / `create_partitions_with_statistics` (`POST .../partitions`, optional `partitionStatistics` and `replaceStatistics`) and `drop_partitions` (`POST .../partitions/drop`), wire-compatible with Java.
  - `list_partitions_by_names` (`POST .../partitions/list-by-names`, body field `specs` as in Java).
  - `list_partitions_by_filter` and `list_partitions_by_filter_paged` (`POST .../partitions/list-by-filter`).
  - `list_partitions_by_name_pattern`, backed by a new `partition_name_pattern` argument on `list_partitions_paged`.
  - Paging follows a non-empty token past an empty page, stops on an empty token, and fails on a repeated token instead of looping.
- `Predicate::to_rest_json`, the inverse of `from_rest_json`, checked against the Java `PredicateJsonSerdeTest` wire strings. It returns `None` for what Java cannot read back: `NOT`, and temporal, decimal or binary literals.
- `Catalog::create_partitions`, `create_partitions_with_statistics` (default `Unsupported`) and `list_partitions_by_names` (default filters `list_partitions`). `RESTCatalog`:
  - sends an idempotent create in batches of 1000, each with its own statistics, and keeps a strict create in one request;
  - rejects statistics for a partition that is not being created, or reported twice;
  - looks partitions up by name in batches of 1000, and uses the plain listing when the endpoint answers 501.
- `Partition` decodes absent statistics as unknown (-1) instead of failing, as Java does since apache/paimon#9379.
- `PartitionStatistics::last_file_creation_time` is signed, like Java `long`, so an unknown value can be sent.

Dropping partitions through the `Catalog` trait is left to the partition commands PR, which needs it.

### Tests

- Unit tests for the request wire shapes and defaults, the drop path, absent statistics decoding, and `Predicate::to_rest_json` round trips, Java wire strings and refusals.
- `rest_api_test.rs`: create and drop requests, lookup by names, list-by-filter request fields and 501, name-pattern listing, and paging over empty pages, empty tokens and repeated tokens.
- `rest_catalog_test.rs`: strict and idempotent create batching, per-batch statistics and their validation, conflict, bad request and missing table mapping, by-names batching and the 501 fallback.
- The mock REST server gains the partition endpoints.

### API and Format

- `RESTApi::list_partitions_paged` takes a fourth `partition_name_pattern: Option<&str>` argument, matching Java `RESTApi.listPartitionsPaged`.
- `PartitionStatistics::last_file_creation_time` changes from `u64` to `i64`, and `PartitionStatistics` derives `PartialEq` and `Eq`.
- New request types `CreatePartitionsRequest`, `DropPartitionsRequest`, `ListPartitionsByNamesRequest` and `ListPartitionsByFilterRequest`.
- New `Catalog` methods with default implementations: `create_partitions`, `create_partitions_with_statistics` and `list_partitions_by_names`.

### Documentation

None: nothing here is reachable from SQL yet. The catalog-managed Format Table scan, the partition commands (#591) and `ANALYZE TABLE` build on these APIs in follow-up PRs.

### DIFF
--- a/crates/paimon/src/catalog/rest/rest_catalog.rs
+++ b/crates/paimon/src/catalog/rest/rest_catalog.rs
@@ -578,18 +578,32 @@ fn spec_key(spec: &HashMap<String, String>) -> Vec<(String, String)> {
     entries
 }
 
-/// Index reported statistics by the spec they describe, rejecting a report for a partition that
-/// is not being created and a partition reported twice: the catalog cannot tell which of two
-/// reports is meant.
+/// Index reported statistics by the spec they describe, before anything is sent.
+///
+/// A partition registered twice is rejected: every occurrence would carry the same report, so a
+/// spec repeated across batches would have an additive report applied once per batch. A report
+/// for a partition that is not being created and a partition reported twice are rejected too: the
+/// catalog cannot tell which of two reports is meant.
+///
+/// Mirrors Java `CatalogFormatTablePartitionManager.validateAndIndexStatistics`.
 fn index_statistics_by_spec(
     identifier: &Identifier,
     partition_specs: &[HashMap<String, String>],
     statistics: Vec<PartitionStatistics>,
 ) -> Result<HashMap<Vec<(String, String)>, PartitionStatistics>> {
-    let requested = partition_specs
-        .iter()
-        .map(spec_key)
-        .collect::<std::collections::HashSet<_>>();
+    let mut requested = std::collections::HashSet::with_capacity(partition_specs.len());
+    for spec in partition_specs {
+        if !requested.insert(spec_key(spec)) {
+            return Err(Error::DataInvalid {
+                message: format!(
+                    "Partition {spec:?} of table {} is registered twice in one request that \
+                     reports statistics; report each partition once",
+                    identifier.full_name()
+                ),
+                source: None,
+            });
+        }
+    }
     let mut by_spec = HashMap::with_capacity(statistics.len());
     for statistic in statistics {
         let key = spec_key(&statistic.spec);
--- a/crates/paimon/src/catalog/rest/rest_catalog.rs
+++ b/crates/paimon/src/catalog/rest/rest_catalog.rs
@@ -34,9 +34,11 @@ use crate::catalog::{
 use crate::common::{CatalogOptions, Options};
 use crate::error::Error;
 use crate::io::cache::{create_local_cache_with_namespace, LocalCache};
-use crate::spec::{Partition, Schema, SchemaChange};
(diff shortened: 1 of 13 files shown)


crates/paimon/src/catalog/rest/rest_catalog.rs
```diff
@@ -411,6 +413,79 @@ impl Catalog for RESTCatalog {
...
+            PARTITION_BATCH_SIZE
+        } else {
+            partition_specs.len()
+        };
+        for batch in partition_specs.chunks(batch_size) {
+            let batch_statistics = statistics.as_ref().map(|by_spec| {
+                batch
+                    .iter()
+                    .filter_map(|spec| by_spec.get(&spec_key(spec)).cloned())
+                    .collect::<Vec<_>>()
```
JingsongLi, line 442: [P2] Reject duplicate partition specs before batching statistics

Each occurrence of a spec clones the same statistics report here, so a spec repeated across batches causes one additive report to be applied multiple times. For example, with 1,001 specs whose first and last entries are identical, a single report containing `record_count = 7`, `ignore_if_exists = true`, and `replace_statistics = false`, the two requests each send that report. I reproduced a stored record count of **14 instead of 7** using this PR's RESTCatalog and mock server.

Java's [`CatalogFormatTablePartitionManager.validateAndIndexStatistics`](https://github.com/apache/paimon/blob/651b9b99db778be471ae07b142d1658efad17bbd/paimon-core/src/main/java/org/apache/paimon/table/format/CatalogFormatTablePartitionManager.java#L232-L256) explicitly rejects duplicate registration specs when statistics are provided to prevent this batching ambiguity. Could we apply the same validation before sending any requests and add a regression test with a duplicate spanning two batches?

sundapeng: Thanks, good catch. Fixed in c4e635e: when statistics are given, `create_partitions_with_statistics` now rejects a spec registered twice before sending any request, as `validateAndIndexStatistics` does. `test_rest_catalog_refuses_a_partition_registered_twice_with_statistics` covers a duplicate spanning two batches (it fails without the fix) and checks that nothing is sent.

JingsongLi (approved):
+1

PR #813, opened 2026-09-11
fix(table): skip staging files and list format table partitions concurrently

sundapeng:
### Purpose

Split out of #591 so it can be reviewed and merged on its own. It changes how a format table scan lists data files, for every format table, not only catalog-managed ones.

- **Staging files were planned as data.** The scan listed everything below its root and only checked the file's own name. A file a committer stages below a partition directory, such as `dt=a/_temporary/0/_temporary/attempt_0/part-1.parquet` or `dt=a/__magic_job/.../part-2.parquet`, was planned and read even though its job may never commit. The same happened below the root of an unpartitioned table.
- **Partition directories were listed one after another.** A scan planned over many partitions paid one sequential listing per partition.

### Brief change log

- `fix(table): skip committer staging files under format table partitions`: a listed file is skipped once any directory or file name below the partition directories starts with `.` or `_`, the rule apache/paimon#8861 gives the Java reader. Partition directories stay exempt, so a value-only layout still reads `__DEFAULT_PARTITION__`. Paths are compared by segments without the scheme, since a listing may spell `file:/` where the root says `file:///`.
- `fix(table): list format table partitions concurrently`: scan roots are listed through a bounded, ordered stream sized by the new `format-table.scan.list-parallelism` option (default 64, clamped to [1, 1000] like Java). Java reads the option only for catalog-managed partitions; here it bounds every format table scan. Plan order is unchanged and the first listing failure still fails the scan.

### Tests

- `format_table_scan::tests`, on an in-memory warehouse:
  - `test_scan_skips_committer_staging_files_of_an_unpartitioned_table`
  - `test_scan_skips_committer_staging_files_below_a_partition_directory` (full scan and a partition-set scan)
  - `test_scan_still_reads_a_value_only_default_partition_directory` (full scan and the null partition)
  - `test_hidden_names_count_only_below_the_partition_levels` (table, leading-prefix and partition roots, mixed scheme spellings)
  - `test_concurrent_listing_keeps_the_plan_order` (eight partitions, parallelism 1 and 4)
- `core_options::tests::test_format_table_scan_list_parallelism`

The three staging tests fail on `main` without the fix: every staging file shows up in the plan.

Commands run:

```
cargo fmt --all -- --check
cargo test --locked -p paimon-datafusion --all-targets
```

Locally, 39 `paimon-datafusion` tests fail because they read fixture tables provisioned by `make docker-up` (two also need the lumina native library). The same 39 fail by name on `main` in this environment; everything else passes.

### API and Format

Adds the table option `format-table.scan.list-parallelism` and `CoreOptions::format_table_scan_list_parallelism`. No storage format change.

### Documentation

No documentation change; the option carries its default and bounds in its doc comment.

### DIFF
--- a/crates/paimon/src/spec/core_options.rs
+++ b/crates/paimon/src/spec/core_options.rs
@@ -44,6 +44,9 @@ const PARTITION_DEFAULT_NAME_OPTION: &str = "partition.default-name";
 const PARTITION_LEGACY_NAME_OPTION: &str = "partition.legacy-name";
 const FORMAT_TABLE_PARTITION_PATH_ONLY_VALUE_OPTION: &str =
     "format-table.partition-path-only-value";
+const FORMAT_TABLE_SCAN_LIST_PARALLELISM_OPTION: &str = "format-table.scan.list-parallelism";
+const DEFAULT_FORMAT_TABLE_SCAN_LIST_PARALLELISM: usize = 64;
+const MAX_FORMAT_TABLE_SCAN_LIST_PARALLELISM: i64 = 1000;
 pub(crate) const BUCKET_KEY_OPTION: &str = "bucket-key";
 const BUCKET_FUNCTION_TYPE_OPTION: &str = "bucket-function.type";
 const BUCKET_OPTION: &str = "bucket";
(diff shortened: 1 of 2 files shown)


JingsongLi (approved):
+1

PR #816, opened 2026-09-11
feat(datafusion): add SHOW, ADD and DROP PARTITION for catalog-managed format tables

sundapeng:
### Purpose

Child of #591, which stays open as a draft that shows the full feature; its remaining changes reach `main` through child PRs.

A Format Table loaded from a REST catalog with `metastore.partitioned-table=true` has its partitions registered in the catalog, and since #814 a scan reads exactly those registrations. DataFusion could not list or change them. This PR adds `SHOW PARTITIONS` and `ALTER TABLE ... ADD / DROP PARTITION` for such tables.

**Depends on:** nothing. It is based on `main`, where #811, #813, #812 and #814 have merged. `MSCK REPAIR TABLE` (#817) and `ANALYZE TABLE` (#815) build on this PR.

### Brief change log

- `SHOW PARTITIONS table [PARTITION (...)]` lists the registered partitions in escaped `key=value/...` form, sorted. Values are read with the column type: a DATE registered as epoch days prints as `dt=2026-07-22`, `month=01` and `month=1` both print as `month=1`, and the default partition prints as `null`. The optional `PARTITION (...)` clause filters by any subset of partition values, compared the same way.
- `ALTER TABLE table ADD [IF NOT EXISTS] PARTITION (...) [PARTITION (...)]` registers complete specifications with the catalog, then creates their directories. `LOCATION` is refused.
- `ALTER TABLE table DROP [IF EXISTS] PARTITION (...)` follows Java `resolveFormatTablePartitionsForDrop`:
  - several specifications per statement; a specification that fixes only some keys, leading or not, expands to every registered partition it matches, and may match none;
  - a complete specification that matches nothing fails the whole statement unless `IF EXISTS` is given;
  - values are compared as the catalog holds them, so `month = 1` does not select a partition registered as `month=01`;
  - a statement made only of complete specifications looks them up through `list-by-names`; one with a partial specification reads the registry once;
  - registrations are removed first and directories deleted afterwards; a partition at a custom location (`options.path`) is only unregistered and its data is left in place;
  - it cannot be combined with other `ALTER TABLE` operations.
- Partition literals are read the way Java reads partition strings. A SQL literal becomes the string Spark would hand to Paimon (quoted strings and `DATE '...'` as written, booleans as `true`/`false`, numbers in canonical form, `NULL` as the default partition), and `parse_format_partition_value` reads it with the column type, as `TypeUtils.castFromString` does. So `month = '01'` and `month = 01` both name `month=1`, a BOOLEAN accepts `t/true/y/yes/1` and `f/false/n/no/0`, and a DATE accepts `yyyy-MM-dd`. An invalid value fails with its column named.
- ADD and DROP PARTITION refuse an empty or whitespace-only string for a string partition column, which would otherwise address the default partition, as Java `PaimonFormatTable.requireNameablePartitionValues` does.
- `Catalog::drop_partitions`, with a default that returns `Unsupported`. `RESTCatalog` sends it in batches of 1000 and refuses tables without catalog-managed partitions.
- `FormatTablePartitionPaths`, `format_partition_value` and `parse_format_partition_value` become public in `paimon::table`.
- The statements live in `crates/integrations/datafusion/src/format_partition_ddl.rs`, the way DELETE, UPDATE and MERGE INTO have their own modules. `sql_context.rs` parses and dispatches them, and keeps DROP PARTITION on a Paimon table, which stays a snapshot commit.

### Tests

  - `test_partition_commands_update_rest_metadata_and_directories`: ADD of several partitions, SHOW with and without a filter, DROP, and their directories;
  - `test_partition_literals`: DATE, INT, BOOLEAN and VARCHAR literals, including Java boolean spellings, unquoted, zero-padded and negative numbers and NULL, with the names and directories they produce, and invalid literals rejected with the column named;
  - `test_drop_partition_specifications`: partial and non-leading specifications, several specifications, a missing complete specification with and without `IF EXISTS`, an empty partial match, a failing specification leaving the statement unapplied, and the refusal to combine with other operations, each checking the remaining registrations and directories;
  - `test_drop_partition_matches_values_as_the_catalog_holds_them`: SHOW reads `month=01` and `month=1` with the column type, while DROP compares the values as registered;
  - `test_drop_partition_looks_up_complete_specifications_by_name`;
  - `test_partition_ddl_refuses_blank_string_values`: ADD and DROP with an empty or whitespace-only string are refused, and the NULL partition and its directory stay;
  - `test_drop_partition_leaves_a_custom_location_in_place`;
  - a compile-time check that the future of `SQLContext::sql` stays `Send`.
- `crates/paimon/tests/rest_catalog_test.rs`: `test_rest_catalog_skips_empty_and_batches_drop_partitions`, `test_rest_catalog_refuses_to_drop_partitions_it_does_not_manage` and `test_rest_catalog_drop_partitions_maps_missing_table`.
- `sql_context` unit test `test_drop_if_exists_partition_does_not_ignore_missing_table`.

```
cargo fmt --all -- --check
```

fmt and clippy pass. `paimon`: 2960 tests pass. `paimon-rest-server`: 10 tests pass. `paimon-datafusion`: 772 pass, including the 7 tests in `rest_format_partition_sql.rs`, and 39 fail. The failures are the 39 tests that also fail by name on `main` in this environment, because they read fixture tables provisioned by `make docker-up` (two also need the lumina native library).

- `format_partition_ddl.rs` turns SQL partition literals into partition strings itself (`partition_literal_to_string`): in Java, Spark resolves the partition spec before Paimon sees it, and DataFusion has no such step for these statements. The strings are then read by `parse_format_partition_value`, the counterpart of `TypeUtils.castFromString`.
- `format_partition_ddl.rs` parses `SHOW PARTITIONS` itself, since sqlparser 0.62 has no such statement.
- `RESTCatalog::drop_partitions` refuses tables without catalog-managed partitions, where Java commits `truncatePartitions`: turning string specs into typed values for every Paimon partition type is not available yet, and the endpoint alone would report success without touching data.
- The compile-time `sql_future_is_send` test keeps the future of `SQLContext::sql` `Send`, which Rust callers need for `tokio::spawn` and `#[async_trait]`.

### API and Format

- New `Catalog::drop_partitions`, with a default implementation that returns `Unsupported`.
- `paimon::table::FormatTablePartitionPaths`, `format_partition_value` and `parse_format_partition_value` become public.

### Documentation

`docs/src/sql.md` gains a Format Table Partitions section: the `metastore.partitioned-table` prerequisite, `SHOW` / `ADD` / `DROP PARTITION`, how partition values are read, the ordering guarantees of ADD and DROP, and how a partition at a custom location is treated. The SQL support scope lists the new statements.

### Scope and limitations

- Several `DROP PARTITION` specifications are written as repeated clauses, `DROP PARTITION (...), DROP PARTITION (...)`; sqlparser 0.62 cannot parse the Hive form `DROP PARTITION (...), (...)`.
- Reading a partition from its custom location is not implemented yet; such partitions fail closed.
- A TIME partition value must be given as an integer. Partition columns typed DECIMAL, TIMESTAMP, FLOAT, DOUBLE or BINARY remain unsupported on this path.

### DIFF
--- a/crates/integrations/datafusion/src/sql_context.rs
+++ b/crates/integrations/datafusion/src/sql_context.rs
@@ -66,9 +66,8 @@ use datafusion::sql::sqlparser::ast::{
     AlterColumnOperation, AlterTableOperation, BinaryLength, CharacterLength, ColumnDef,
     ColumnOption, CreateFunction, CreateFunctionBody, CreateTable, CreateTableOptions, CreateView,
     Delete, Expr as SqlExpr, FromTable, FunctionBehavior, FunctionReturnType, Ident, Insert, Merge,
-    ObjectName, ObjectType, Partition as SqlPartition, RenameTableNameKind, Reset, ResetStatement,
-    Set, ShowCreateObject, SqlOption, Statement, TableFactor, TableObject, Truncate, Update, Use,
-    Value as SqlValue,
+    ObjectName, ObjectType, RenameTableNameKind, Reset, ResetStatement, Set, ShowCreateObject,
+    SqlOption, Statement, TableFactor, TableObject, Truncate, Update, Use, Value as SqlValue,
 };
 use datafusion::sql::sqlparser::dialect::GenericDialect;
 use datafusion::sql::sqlparser::keywords::Keyword;
@@ -82,9 +81,6 @@ use paimon::spec::{
     DecimalType, DoubleType, FloatType, IntType, LocalZonedTimestampType, MapType as PaimonMapType,
     RowType as PaimonRowType, SchemaChange, SmallIntType, TimestampType, TinyIntType,
     VarBinaryType, VarCharType, VariantType,
-};
-use paimon::table::{
-    format_partition_value, parse_format_partition_value, FormatTablePartitionPaths,
 };
 
 use crate::error::to_datafusion_error;
@@ -458,11 +454,17 @@ impl SQLContext {
             // Time-travel queries are not DDL; skip our own parsing and handle directly.
             return self.handle_time_travel_query(&rewritten_sql).await;
         }
-        if let Some(show_partitions) = parse_show_partitions(&rewritten_sql)? {
-            return self
-                .handle_show_partitions(&show_partitions, enable_ident_normalization)
-                .await;
+        if let Some(show_partitions) =
+            crate::format_partition_ddl::parse_show_partitions(&rewritten_sql)?
+        {
+            return crate::format_partition_ddl::execute_show_partitions(
+                self,
+                &show_partitions,
+                enable_ident_normalization,
+            )
+            .await;
         }
+
         let statements = parse_sql_statements(&rewritten_sql)?;
 
         if statements.len() != 1 {
@@ -1323,7 +1325,7 @@ impl SQLContext {
                 ));
             }
             return self
-                .handle_drop_partition(
+                .handle_drop_partitions(
                     catalog,
                     &identifier,
                     &drop_partition_requests,
@@ -1402,16 +1404,16 @@ impl SQLContext {
                     if_not_exists,
                     new_partitions,
                 } => {
-                    return self
-                        .handle_add_partitions(
-                            catalog,
-                            &identifier,
-                            new_partitions,
-                            *if_not_exists,
-                            if_exists,
-                            enable_ident_normalization,
-                        )
-                        .await;
+                    return crate::format_partition_ddl::execute_add_partitions(
+                        self,
+                        catalog,
+                        &identifier,
+                        new_partitions,
+                        *if_not_exists,
+                        if_exists,
+                        enable_ident_normalization,
+                    )
+                    .await;
                 }
                 other => {
                     return Err(DataFusionError::Plan(format!(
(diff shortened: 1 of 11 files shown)


crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -3033,6 +3450,222 @@ fn is_table_not_exist(e: &paimon::Error) -> bool {
...
+                    _ => text.trim().to_string(),
+                };
+                parse_format_partition_value(&text, data_type)
+                    .and_then(|datum| {
+                        format_partition_value(
+                            &datum,
+                            data_type,
+                            options.partition_default_name(),
+                            options.legacy_partition_name(),
+                        )
```
QuakeWang: `format_partition_value` maps empty and whitespace-only strings to the default partition name. As a result, `DROP PARTITION (dt = '')` deletes the NULL partition, including its files—I reproduced this locally. Please reject these literals before formatting and add a regression test.

sundapeng: Agreed, fixed in c0fe18c.

JingsongLi (commented):
Can you extract codes from sql_context.rs to a dedicated rs file?

sundapeng:
Agreed, moved to `crates/integrations/datafusion/src/format_partition_ddl.rs` in f18d972.

JingsongLi (approved):
+1

PR #817, opened 2026-09-11
feat(datafusion): add MSCK REPAIR TABLE for catalog-managed format tables

sundapeng:
### Purpose

Child of #591, which stays open as a draft that shows the full feature; its remaining changes reach `main` through child PRs.

A Format Table with catalog-managed partitions reads only the partitions its catalog registers. A partition directory written by a tool that does not register partitions stays invisible to a scan, and a registration whose directory was removed points at nothing. This PR adds `MSCK REPAIR TABLE` for such tables, following Java `FormatTablePartitionRepair`, which backs the same statement in Spark.

#816 has merged, so this PR now applies to `main` on its own. `ANALYZE TABLE` (#815) builds on this PR.

### Brief change log

- `MSCK REPAIR TABLE table [{ADD|DROP|SYNC} PARTITIONS]` on a Format Table with catalog-managed partitions:
  - `ADD`, which plain `MSCK REPAIR TABLE` also means, registers every partition directory the catalog does not hold; `DROP` unregisters every registration whose directory is gone; `SYNC` does both. New registrations are created, ignoring any that already exist, before any is dropped.
  - Only registrations change; no directory or data file is created or deleted.
  - Directory discovery and the catalog listing both finish before any change, so a listing failure changes nothing.
  - Partitions are matched by their escaped `key=value/...` name and registered with the values their directories spell, so a directory `month=01` is registered as `month=01`, as in Java.
  - A partition registered at a custom location (`options.path`) is never unregistered, since its data does not live under the table directory.
  - `MSCK` without `REPAIR` and tables without catalog-managed partitions are refused.
- `FormatTablePartitionPaths::discover` finds the complete partition specs under the table directory, the counterpart of Java `PartitionPathUtils.searchPartSpecAndPaths` as repair calls it. Hidden directories are skipped except the value-only default partition directory, segments outside the configured layout and paths shallower than the partition depth are ignored, a missing root holds no partitions, and results are sorted and deduplicated by partition name.
- `unescape_path_name` moves from the Format Table scan to `spec::partition_utils`, beside `escape_path_name`, so the scan and discovery share it.
- The statement lives in `crates/integrations/datafusion/src/format_partition_repair.rs`, beside the repair it runs; `sql_context.rs` only dispatches it.

### Tests

  - `test_msck_repair_reconciles_registrations_with_directories`: ADD registers a directory the catalog does not know, and SYNC unregisters a partition whose directory was removed while the other directory stays;
  - `test_msck_repair_keeps_a_partition_at_a_custom_location`: SYNC keeps a partition registered at a custom location even though no directory for it exists under the table;
  - `test_msck_repair_refuses_to_drop_without_a_table_directory`: DROP and SYNC fail and leave the registrations unchanged when the table root was moved away or replaced by a file.
- `crates/paimon/tests/format_partition_test.rs`: `discover_format_partitions_returns_complete_sorted_specs`, `discover_format_partitions_treats_missing_root_as_empty`, `discover_value_only_partitions_rejects_parent_traversal_value` and `discover_value_only_partitions_keeps_hidden_default_directory`.
- Unit tests `spec::partition_utils::tests::test_unescape_path_name` and `table::format_partition::tests::test_storage_not_found_matches_only_not_found`.

```
cargo fmt --all -- --check
```

fmt and clippy pass. `paimon`: 2966 tests pass. `paimon-rest-server`: 10 tests pass. `paimon-datafusion`: 775 pass, including the 10 tests in `rest_format_partition_sql.rs`, and 39 fail. The failures are the 39 tests that also fail by name on `main` in this environment, because they read fixture tables provisioned by `make docker-up` (two also need the lumina native library).

- `DROP` and `SYNC` first require the table root to be a directory and fail if it is missing or a file. On a store without directory entries, a root that holds at least one object counts. Discovery, like Java's, reads a missing root as a table without partitions, and listing a missing local directory returns nothing. Java's repair has no such check, so a missing table directory would make `DROP` or `SYNC` unregister every partition.
- Discovery fails on a matching directory name whose escaping is malformed or not canonical, because the spec registered for it would not resolve back to the same directory. Java's `unescapePathName` keeps a malformed escape as written.

### API and Format

- New public method `paimon::table::FormatTablePartitionPaths::discover`.

### Documentation

`docs/src/sql.md` gains an `MSCK REPAIR TABLE` subsection under Format Table Partitions and lists `MSCK REPAIR TABLE` in the SQL support scope.

### Scope and limitations

- Repair always covers the whole table, and there is no dry run.
- Java's `format-table.repair.collect-statistics`, which also measures the partitions a repair finds, is not supported; `ANALYZE TABLE` (#815) measures them afterwards.

### DIFF
--- a/crates/integrations/datafusion/src/format_partition_repair.rs
+++ b/crates/integrations/datafusion/src/format_partition_repair.rs
@@ -90,9 +90,8 @@ async fn repair(
         )
         .await?;
     let registered_partitions = catalog.list_partitions(identifier).await?;
-    // A partition registered at a location of its own does not live under the table directory,
-    // so discovery never finds it there. Repair leaves it registered rather than reading that as
-    // a directory gone missing.
+    // A partition registered at a location of its own is never discovered under the table
+    // directory, so repair leaves it registered.
     let custom_located = registered_partitions
         .iter()
         .filter(|partition| {
--- a/crates/integrations/datafusion/src/format_partition_repair.rs
+++ b/crates/integrations/datafusion/src/format_partition_repair.rs
@@ -1,0 +1,156 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! MSCK REPAIR TABLE for Format Tables with catalog-managed partitions.
+
+use std::collections::{BTreeMap, HashMap, HashSet};
+
+use datafusion::error::{DataFusionError, Result as DFResult};
+use datafusion::prelude::DataFrame;
+use datafusion::sql::sqlparser::ast::{AddDropSync, Msck};
+use paimon::catalog::{Catalog, Identifier};
+use paimon::spec::CoreOptions;
+use paimon::table::{FormatTablePartitionPaths, Table};
+
+use crate::error::to_datafusion_error;
+use crate::format_partition_ddl::ensure_catalog_managed_format_table;
+use crate::sql_context::{ok_result, SQLContext};
+
+pub(crate) async fn execute_msck(ctx: &SQLContext, msck: &Msck) -> DFResult<DataFrame> {
+    if !msck.repair {
+        return Err(DataFusionError::Plan(
+            "MSCK requires the REPAIR keyword".to_string(),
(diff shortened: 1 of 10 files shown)


crates/integrations/datafusion/src/format_partition_repair.rs
```diff
@@ -0,0 +1,155 @@
...
+    let partition_paths = FormatTablePartitionPaths::new(
+        table.schema().partition_keys().iter().cloned(),
+        core_options.format_table_partition_only_value_in_path(),
+    );
+    let table_path = table.location();
+
+    if matches!(mode, RepairMode::Drop | RepairMode::Sync) {
+        // Discovery treats a missing root as empty. Destructive repair must fail
+        // instead, or it could unregister every catalog partition.
+        table.file_io().list_status(table_path).await?;
```
QuakeWang: `list_status` returns an empty list when the directory is missing on the local filesystem, so this check doesn't protect against a missing table root. I reproduced this: after moving the root directory, `MSCK ... DROP PARTITIONS` succeeds and unregisters the existing partitions. Please check that the root exists and is a directory, and cover this case for DROP and SYNC.

sundapeng: Thanks for the repro. DROP and SYNC now refuse a table root that is missing or not a directory, and a new test covers both modes (2491663).

JingsongLi (approved):
+1

PR #819, opened 2026-09-12
feat(file_index): generate indexes for append writes

### LINKED ISSUE #818: Support FileIndex generation for append writes
### Search before asking

### Description

Rust can read Bitmap and Bloom Filter FileIndexes, but normal append writes do not generate them: `DataFileWriter` leaves `extra_files` empty and `embedded_index` unset.

This is a write-side follow-up to #34 and #780. Reuse the existing index writers and format to:

- Parse index column declarations, per-column options, and `file-index.in-manifest-threshold`, with explicit errors for invalid or unsupported configurations.
- Build indexes per physical data file, preserving row order and NULL positions across batches and resetting state on file rolling.
- Store the complete serialized index in `embedded_index` when its size is at most the threshold; otherwise write a `.index` sidecar and register it in `extra_files`. Match Java's default of 500 B.
- Handle both foreground and background close paths. Return metadata only after data and index completion succeeds, and clean up newly created sidecars on failure.

Tests should cover unconfigured writes, Bitmap/Bloom, multiple columns, NULLs, empty batches, rolling, threshold boundaries, and failure/abort cleanup. After commit, verify identical query results with indexes enabled and disabled, actual FileIndex pruning, and residual filtering for Bloom false positives.

Scope is limited to ordinary append writes and supported top-level columns. KV writes, COW/DataEvolution rewrites, dedicated Blob/Vector paths, nested indexes, additional index types, and public API expansion are excluded. Java-generated table fixtures remain a separate read-compatibility follow-up.

### Willingness to contribute

QuakeWang:
### Purpose

Closes #818.

Rust can read FileIndexes, but ordinary append writes leave index metadata empty. This change generates configured indexes during append writes so committed files can benefit from FileIndex pruning.

### Brief change log

- Validate top-level Bitmap/Bloom Filter configuration and build indexes per physical file, including rolling files.
- Store complete indexes embedded or as sidecars using the configured threshold (default 500 B), and clean up failed writes.

### Tests

- Cover commit/reload, actual pruning, Bloom false positives, NULLs, rolling, threshold boundaries, failure cleanup and commit abort.
- Formatting, workspace Clippy and core tests with `fulltext,vortex` passed (2968 passed, 2 ignored).

### API and Format

No public API or index format changes. Reuses existing FileIndex serialization and metadata; generation is limited to ordinary append writes.

### Documentation

Added append FileIndex configuration and scope to the getting-started guide.

### DIFF
--- a/crates/paimon/src/table/table_write.rs
+++ b/crates/paimon/src/table/table_write.rs
@@ -34,6 +34,7 @@ use crate::table::bucket_assigner_dynamic::DynamicBucketAssigner;
 use crate::table::bucket_assigner_fixed::FixedBucketAssigner;
 use crate::table::bucket_function::validate_bucket_function;
 use crate::table::commit_message::CommitMessage;
+use crate::table::data_file_index_writer::FileIndexOptions;
 use crate::table::data_file_writer::DataFileWriter;
 use crate::table::dedicated_format_file_writer::AppendDedicatedFormatFileWriter;
 use crate::table::kv_file_writer::{KeyValueFileWriter, KeyValueWriteConfig};
@@ -148,6 +149,8 @@ pub struct TableWrite {
     has_dedicated_vector_fields: bool,
     row_kind_generator: Option<RowKindGenerator>,
     row_kind_filter: Option<RowKindFilter>,
+    file_index_options: Option<Arc<FileIndexOptions>>,
+    indexed_write_failed: bool,
 }
 
 impl TableWrite {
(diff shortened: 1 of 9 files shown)


crates/paimon/src/table/table_write.rs
```diff
@@ -148,6 +149,8 @@ pub struct TableWrite {
     has_dedicated_vector_fields: bool,
     row_kind_generator: Option<RowKindGenerator>,
     row_kind_filter: Option<RowKindFilter>,
+    file_index_options: Option<Arc<FileIndexOptions>>,
+    indexed_write_failed: bool,
```
JingsongLi: Why need this?

QuakeWang: This was an unnecessary defensive check to prevent writer reuse after a failure. Removed in 2fa63e3; error propagation and file cleanup are preserved.

JingsongLi (commented):
[P2] Preserve append writes for unimplemented upstream index types

`FileIndexOptions::parse` now rejects every `file-index.<identifier>.columns` property except `bitmap` and `bloom-filter`, and `TableWrite::new` invokes it for every append writer. Apache Paimon supports, for example, `file-index.bsi.columns` and `file-index.range-bitmap.columns`; a Rust writer previously left those table options alone, and the existing reader deliberately skips unknown index payloads. Consequently, appending to a Java-created table with one of these valid properties now fails before any file is written, even though Rust can conservatively leave that index ungenerated/unread.

Only select supported identifiers for generation, and skip unsupported identifier groups (including their per-column options) so unsupported upstream indexes retain the reader's conservative behavior. Keep validation for the index types Rust does generate.

JingsongLi (approved):
+1

PR #820, opened 2026-09-12
feat(rest): add permission management APIs

plusplusjiajia:
## Purpose

Ports the client half of the experimental REST management API added in apache/paimon#9393 (object permissions; wire contract in apache/paimon#9406) to Rust.

## What it adds

    GET  /v1/{prefix}/permissions?resourceType&database&table&function&view&principal&access&maxResults&pageToken
    POST /v1/{prefix}/permissions/grant
    POST /v1/{prefix}/permissions/revoke

`api/management.rs` mirrors `org.apache.paimon.management` (`ResourceType`, `PermissionResource`, `PermissionAccess`, `PermissionColumns`, `PermissionAssignment`, `ListPermissionsRequest`), and `RESTApi` gains `list_permissions_paged`, `grant_permission` and `revoke_permission`, with thin delegates on `RESTCatalog`. Wire names, query parameters and client-side validation follow Java. No new dependencies.

## Notes for review

- **Responses are not validated on deserialize**, only requests. This matches Java's non-validating Jackson constructor and is deliberate: a server may list a value a client could not have submitted, and one such row must not fail a whole page.
- The management methods stay **off the `Catalog` trait**, on `RESTCatalog` only, as in Java — no other catalog implementation speaks this contract. They are exposed as plain methods rather than behind a `permissionManagement()` wrapper.
- A grant posts `PermissionAssignment` directly; Java's separate `GrantPermissionRequest` has the identical wire shape, so a second type would carry nothing.
- `PermissionResource`'s locator getters are `database_name()`, `table_name()`, `function_name()`, `view_name()` — Rust cannot give a constructor and a getter the same name, and the constructors take the plain names.
- `expireTime` must end in `Z`, matching `Instant.parse` on JDK 8/11. JDK 12 widened `ISO_INSTANT` to accept other offsets, so against a server on a newer JDK this client is the stricter of the two.


JingsongLi (approved):
+1

PR #821, opened 2026-09-12
fix(rest): sign the query string the client actually sends

plusplusjiajia:
## The bug

Java's `RESTAuthParameter` constructor URL-encodes every parameter value, so the signer sees the encoded form:

    this.parameters.put(entry.getKey(), encodeString(entry.getValue()));

The Rust port of that type stores the values raw, while the HTTP client still percent-encodes them on the way out. The client therefore signs one string and sends another, and DLF answers HTTP 401 for any value containing a character outside `[A-Za-z0-9*-._]`. Measured against a real DLF catalog, `databaseNamePattern` of `%`, `db%`, `a:b`, `a/b`, `a b`, `a+b`, `a=b` and `dt=2026-07-22/%` all failed, while `plain` and `a*b` passed — and the Java client passed on every one of them. Every `*NamePattern` filter takes a `%` wildcard, so this reaches shipping call sites.

The mock server used by the test suite does not verify signatures, which is why no existing test catches it.

## The fix

Restore the encoding in `RESTAuthParameter::new`, where Java does it, so every signer inherits it. Values are encoded; keys are left alone, as in Java. The DLF4 canonical-query construction is untouched — only its input was wrong.

`DLFOpenApiSigner` decodes those values back out, so its two `urlencoding::decode` calls become `RESTUtil::decode_string` — `urlencoding` left the `+` that `encode_string` writes for a space in place. `decode_string` itself was built on `form_urlencoded::parse`, which splits on `=` and `&`, so it is rewritten here as the whole-string decoder Java's `RESTUtil.decodeString` (`URLDecoder.decode`) actually is; `URLDecoder` resolves `%XX` and maps `+` to a space and leaves `=` and `&` alone. Malformed percent bytes now decode lossily rather than yielding an empty string, so a bad path signs as itself and fails diagnosably. That swaps the `urlencoding` dependency for `percent-encoding`, which was already in the resolved graph.

`RESTApi` also stops pre-encoding the `warehouse` value, which would otherwise now be encoded twice.

One divergence found along the way and fixed for parity rather than for the 401: `ResourcePaths::new` interpolated the catalog `prefix` raw, where Java encodes it (`ResourcePaths.java:64-66`), including a multi-segment prefix that `ResourcePathsTest` pins as `catalog/id` to `/v1/catalog%2Fid/permissions`. With the decoder corrected this is no longer load-bearing, but the paths this client builds should match the ones Java builds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4692,6 +4692,7 @@ dependencies = [
  "paimon-mosaic-core",
  "paimon-vindex-core",
  "parquet",
+ "percent-encoding",
  "pretty_assertions",
  "rand 0.8.7",
  "regex",
@@ -4713,7 +4714,6 @@ dependencies = [
  "typed-builder",
  "unicode-segmentation",
  "url",
- "urlencoding",
  "uuid",
  "vortex",
  "zstd",
@@ -7379,12 +7379,6 @@ dependencies = [
  "percent-encoding",
  "serde",
 ]
-
-[[package]]
-name = "urlencoding"
-version = "2.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "daf8dba3b7eb870caf1ddeed7bc9d2a049f3cfdfae7cb521b087cc33ae4c49da"
 
 [[package]]
 name = "utf8-ranges"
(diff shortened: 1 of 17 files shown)


JingsongLi (approved):
+1

PR #824, opened 2026-09-14
feat(catalog): add tag lifecycle APIs

XiaoHongbo-Hope:
### Summary

- add Catalog APIs to create, get, and delete table tags
- support explicit snapshot IDs and latest-snapshot tag creation
- keep tag operations unsupported by default; FileSystemCatalog uses TagManager and RESTCatalog uses server endpoints
- return the full tag response, including metadata written by Java/PyPaimon
- reject path-unsafe filesystem tag names
- add typed errors plus filesystem and REST lifecycle tests

### Scope

This PR adds the Rust core API only. C/Go bindings will follow separately. Tag listing and retention creation are outside this initial API. REST tag deletion follows the server implementation. FileSystemCatalog deletes tags while their snapshots still exist, but returns Unsupported for expired snapshots until Java-compatible TagDeletion cleanup is implemented.

### Testing

- `cargo test -p paimon tag`
- `cargo test -p paimon --test rest_catalog_test test_catalog_tag_lifecycle`
- `cargo fmt --all --check`
- `git diff --check`


JingsongLi (approved):
+1

PR #826, opened 2026-09-14
fix(fm_index): store blocks in Java's codec envelope, not the SST one

jackylee-ch:
`write_block` sent FM blocks through `compress_block`, which prepends an
uncompressed-size varint. The varint belongs to the SST and bitmap block
formats, not the codec: Java's `FMIndexFile#writeBlock` stores
`BlockCompressor` output verbatim with `storedLength = compressedLength`, and
`decodeStoredBlock` decompresses it directly. So paimon-rust cannot read a
Java-written FM global index, failing with `Compressed block size does not match
its metadata`, and Java cannot read ours. Both default to `lz4`, and ZSTD is
affected too since the varint precedes the codec branch.

**Fix**: add a codec-only pair, `compress_codec_block` and
`decompress_codec_block`, and point FM at it. `compress_block` keeps the varint
for SST and bitmap, where Java does prepend it.

The V1 goldens store blocks uncompressed, so no test covered this; the new one
asserts the stored bytes against Java's envelope.


JingsongLi (commented):
[P1] Preserve reads of previously written Rust V1 indexes

`decode_stored_block` now always treats a compressed V1 block as codec-only. However, the existing V1 writer called `compress_block`, which placed the outer uncompressed-size varint before the same codec payload. Its metadata and footer still declare version 1, so a table whose FM index was written by the current main branch before this change will pass the checksum and then fail to read. The new `codec_block_rejects_the_varint_prefixed_form` test demonstrates that exact legacy encoding is rejected.

Keep a bounded legacy decode fallback (the removed expected-size decoder already provided it), or introduce a readable on-disk version distinction. Please add a regression fixture that is written with the pre-change V1 encoding and read through this path.

JingsongLi (approved):

PR #827, opened 2026-09-14
fix(table): support all Java-comparable types in min/max aggregators

jackylee-ch:
Java's `FieldMinAggFactory` and `FieldMaxAggFactory` accept any type
`TypeCheckUtils#isComparable` allows, which excludes only MAP, MULTISET, ROW,
ARRAY, VECTOR, VARIANT and BLOB. `MinMaxState` covered eleven roots and omitted
four that `InternalRowUtils#compare` handles: BOOLEAN (`Boolean.compare`),
TIMESTAMP WITH LOCAL TIME ZONE (same branch as TIMESTAMP), BINARY and VARBINARY
(`byteArrayCompare`). Configuring `max` on such a column made an aggregation
table unreadable with `Aggregate function 'max' does not support data type ...`.

The four are added. BLOB stays rejected even though Arrow maps it to the same
`Binary`, and the accumulator now carries the Arrow timezone, read back from
`paimon_type_to_arrow` so it cannot drift — a result array without it fails
`RecordBatch::try_new` against the field.

An outdated comment claimed BOOLEAN has no ordering; that test now pins BLOB.


JingsongLi (commented):
Requirement fit: SUPPORTED. This completes min/max aggregation support for value types already represented in Paimon Rust, providing expected aggregate behavior for tables using booleans, local-zoned timestamps, binary, and varbinary columns.

Implementation: CLEAN. I traced the comparable-value handling and accumulator update paths and found no actionable regressions. Targeted aggregation tests pass (30 tests).

JingsongLi (approved):
+1

PR #828, opened 2026-09-14
feat(go): expose snapshot and tag APIs

XiaoHongbo-Hope:
## Purpose

Expose the existing Rust snapshot and catalog tag APIs to Go.

## Changes

- add `Table.LatestSnapshot`
- add `Catalog.CreateTag`, `Catalog.GetTag`, and `Catalog.DeleteTag`
- expose the shared `GetTagResponse` metadata shape through the C bridge
- preserve not-found and already-exists error categories
- document creating, inspecting, and reading a tag

A nil snapshot ID selects the latest snapshot. Tag retention remains out of scope because the Rust catalog API does not expose it yet.

## Tests

- `cargo test -p paimon-c` (77 passed)
- `cargo clippy -p paimon-c --all-targets -- -D warnings`
- `go vet ./...`
- `go test -run TestCatalogTagLifecycle -v`
- Shanghai regression DLF: create table and snapshot, create/get/delete tags, and read with `scan.tag-name`


JingsongLi (approved):
+1

PR #830, opened 2026-09-14
fix(rest-server): use Java's uppercase resourceType in error responses

jackylee-ch:
The five `resourceType` values this server emits are lowercase, but Java's
`ErrorResponse` declares them as `DATABASE` / `TABLE`, and a Java client
dispatches on the field: `RESTCatalog#alterTable` raises
`TableNotExistException` only when it equals `RESOURCE_TYPE_TABLE`, comparing
char by char through `StringUtils#equals`. With `table` neither branch matches,
so `alterTable(ident, changes, false)` on a missing table silently returns
instead of throwing. `createBranch` rethrows in the same position, so the
symptom varies by call.

They are now the uppercase constants. No effect on the Rust client, which keys
off the status code alone.

Left alone deliberately: the two column errors still answer 400 with
`column:<table>`. Making them Java's 404 plus `COLUMN` needs the Rust client
changed too, which would otherwise read a missing column as a missing table.


JingsongLi (commented):

resourceType is a wire-level discriminator for Java REST clients, so the existing lowercase values prevent expected exception dispatch despite the correct HTTP status. The change aligns all affected database/table errors and intentionally leaves column status/resource handling untouched until both server and Rust client can move together. I found no regression in the changed path. Verification: cargo test -p paimon-rest-server (all unit, integration, and doctests passed). Review scope: REST error serialization and client-visible status/resource-type contracts.

JingsongLi (approved):
+1

PR #831, opened 2026-09-14
fix(python): align native scan planning with PyPaimon

JingsongLi:
### Purpose

Related: #413 (follow-up to the Python scan-planning bindings).

Make native scan plans usable by PyPaimon for incremental and distributed reads while preserving snapshot identity, branch isolation, deletion vectors, and primary-key merge groups. For example, versions of the same primary key in two APPEND snapshots must be planned together; planning each commit separately can return both versions.

### Brief change log

- Add combined APPEND-delta planning for an explicit snapshot range, using the ending snapshot for metadata and deletion vectors.
- Add Data Evolution row-position slices and shards before manifest/entry row-range pruning, group statistics, projection, and deletion-vector filtering. Intersect explicit/global-index ranges only after candidate positions are assigned, and preserve group order when a later limit selects rows.
- Preserve snapshot IDs for empty plans and resolve branch-qualified table handles and their metadata managers.
- Decode deletion-vector arrays using their actual Avro writer schema and read legacy Python bucket-local index references from `table/index` when the canonical path is absent.
- Accept Python `bytes`/`bytearray` binary literals.
- Keep overlapping composite floating-point key ranges together, preserving signed-zero ordering and conservatively grouping files with NaN key bounds.

### Tests

Passed locally:

- Complete Rust core unit suite: 2,775 passed, 2 ignored.
- Rust incremental batch-scan tests: 34.
- New Rust planner-parity integration suite: 15.
- Python-binding Rust unit tests: 73.
- `cargo clippy --offline -p paimon -p pypaimon_rust --all-targets -- -D warnings`, `cargo fmt --all -- --check`, and `git diff --check`.

The full workspace and Spark-warehouse integration suites were not run locally.

The new Rust cases cover full/combined slice and shard intersections across files and manifests; B-tree/bitmap search modes; row-ID gaps and balanced shards; projection, updates, limits, and deletion vectors; repeated and historical deletes; branch isolation; compaction and missing snapshots; primary-key residual filters; FLOAT/DOUBLE signed-zero merge planning and conservative NaN-bound grouping; and Java/Python DV schema and path compatibility. A small enumerated row-ID oracle checks positional boundaries independently of the range implementation.

### API and Format

Adds `Plan.snapshot_id()`, `Table.branch()`, `ReadBuilder.new_incremental_scan(start, end)`, and DE row-position selection methods to the binding, plus the corresponding combined-scan APIs in the core.

Existing per-commit `IncrementalScan.plan()` behavior is preserved. The newly added row-position configuration requires combined planning. Split/storage format versions are unchanged.

NaN key bounds use conservative overlapping groups. Legacy bucket-local DV resolution can add filesystem existence checks; explicit external paths and existing canonical files retain priority. Complete parity for all PyPaimon modes remains follow-up work.

### Documentation

Updated `docs/src/python-binding.md` and the Python type stubs with the new APIs, range semantics, and compatibility guarantees.


crates/paimon/src/table/table_scan.rs
```diff
@@ -1827,11 +1949,26 @@ impl<'a> PaimonTableScan<'a> {
         let open_file_cost = core_options.source_split_open_file_cost();
         let partition_keys = self.table.schema().partition_keys();
 
+        // Assign row positions using the full candidate file ranges, before
+        // group stats, projection or DVs change visible rows. Intersect explicit
+        // and index-selected ranges only after assigning the positional range.
+        let effective_row_ranges = if let Some(selection) = self.row_position_selection {
+            Some(selection.select(&entries, effective_row_ranges.as_deref())?)
```
leaves12138, line 1962: [P1] Assign row positions before row-range pruning

`entries` is already pruned by explicit/global-index row ranges here: `plan_snapshot_from_lists` passes `manifest_row_ranges` into `plan_manifest_entries_with_trace`, which applies `retain_manifest_row_ranges` and `retain_manifest_entry_row_ranges`. If that removes a preceding file, `selection.select` renumbers the remaining rows, so intersecting the ranges afterward cannot preserve the requested positions.

For example, persist a DE snapshot with two three-row files covering row IDs `[0,2]` and `[3,5]`, then run:

```python
rb.with_row_ranges([(4, 4)])
plan = rb.new_scan().with_row_position_slice(3, 6).plan()
```

The positional slice should select IDs `[3,5]`, and its intersection with `[4,4]` should retain row 4. Instead, early pruning removes the first file, the remaining three rows are assigned positions `[0,3)`, and the returned plan has no splits. I reproduced this through the public Rust `new_scan().plan()` path against this head; the expected `[RowRange { from: 4, to: 4 }]` was actually `[]`.

Please preserve the complete candidate position space when positional selection is enabled, and only then intersect explicit/index-selected ranges. Disabling the early row-range pruning for positional scans made the reproducer and the existing row-position tests pass locally. Add multi-file coverage where the explicit range removes an entire preceding file, including combined incremental scans and shard selection; the current intersection tests do not exercise that case.

leaves12138 (changes requested):
Reviewed commit `71bbcf72108a5d968635942e6f8de755baa23070`.

One blocking correctness issue: early row-range pruning changes the position space used by the new DE positional selection, causing valid rows to be omitted. See the inline comment for a reproduced two-file example.

Local validation on the unmodified PR head:
- `cargo test --offline -p paimon --lib`: 2,773 passed, 2 ignored.
- `cargo test --offline -p paimon --test incremental_batch_scan_test`: 31 passed.

A separate regression test reproduces the issue; a temporary change deferring early row-range pruning makes it pass along with the existing row-position tests. These experimental changes were reverted after validation. I did not rebuild the Python binding or run the full workspace/Spark integration suites.

leaves12138 (approved):
Re-reviewed `877595f3a2c11b5dab06f56843ffb6dbae8cbc56`. My previous row-position pruning blocker is fixed: both scan paths preserve the candidate position space before intersecting explicit/global-index ranges. The two-file reproducer now returns row 4, and the snapshot/combined-delta slice and shard regression tests verify the actual read output.

Also reviewed the added floating-point key normalization in writer sorting, same-file deduplication, and cross-file merging, including rejection of legacy non-monotonic NaN key ordering. No further blocking findings in the changes reviewed.

Local validation on this exact head:
- `cargo test --offline -p paimon --lib row_position`: 13 passed.
- `cargo test --offline -p paimon --test scan_planning_parity_test`: 15 passed, including global-index position intersection, deletion vectors, branches, and floating-point keys.
- `cargo test --offline -p paimon --test incremental_batch_scan_test`: 34 passed.
- `cargo test --offline -p paimon --lib`: 2,777 passed, 2 ignored (includes the row-position unit tests above).

I did not rebuild the Python binding or run the full workspace/Spark integration suites.

PR #833, opened 2026-09-15
fix(core): align timestamp partition paths with Java

JingsongLi:
### Purpose

Align timestamp partition directories with Java for both writes and reads. Rust currently omits the `.000` suffix for whole-second non-legacy `TIMESTAMP(3)` values and shortens legacy fractions such as `.100` to `.1`. These differences produce incompatible data directories and prevent native scans from locating Java-written bucket-local deletion vectors. The DV failure was reproduced while fixing the review in apache/paimon#9825.

### Brief change log

- Use Java timestamp formatting in the shared partition-path calculation used by data files and indexes. Non-legacy formatting keeps the declared fractional precision; legacy formatting uses 3, 6, or 9 fractional digits, matching `LocalDateTime.toString()`.
- Remove the `java_timestamps` switch and separate DV-only Java path calculation. The existing `partition.legacy-name` table option still selects Java's legacy or non-legacy format.
- Preserve explicit DV paths and the existing fallback for older Python files under `table/index`. Canonical bucket-local files take precedence over same-name table-index copies.

### Tests

- Partition utilities: 32 tests passed, covering whole-second precision, significant fractions at precision 0, and legacy fractional widths.
- Index-file paths: 14 tests passed, including timestamp partitions through the table-level resolver, same-name decoys, missing files, and authoritative explicit paths.
- Python binding writes: 19 tests passed. Five new cases verify physical timestamp directories, two commits, fresh-table planning, and SQL reads. Four of those cases failed against the previous implementation.
- With a wheel rebuilt from this final change, PyPaimon passed 259 related tests and exercised 84 native plans, including all 91 DV-path cases.
- `cargo fmt --all -- --check`, `cargo clippy --locked -p paimon --lib --tests -- -D warnings`, Python Flake8, and `git diff --check` passed.

### API and Format

No public API or binary serialization changes. Generated timestamp directory names change for values previously formatted incorrectly by Rust. Files under those old Rust-only directory names are not automatically moved, and the shared path calculation no longer resolves those spellings. Existing explicit external paths remain authoritative.

### Documentation

Updated timestamp formatting comments to describe Java fractional precision and legacy widths.


crates/paimon/src/spec/partition_utils.rs
```diff
@@ -460,26 +461,28 @@ fn format_timestamp_legacy(dt: NaiveDateTime) -> String {
...
 fn format_timestamp_non_legacy(dt: NaiveDateTime, precision: u32) -> String {
     let nano = dt.nanosecond();
     let ymdhms = dt.format("%Y-%m-%d %H:%M:%S").to_string();
 
-    if precision == 0 || nano == 0 {
-        return ymdhms;
-    }
-
-    // Pad nano to 9 digits, then strip trailing zeros but keep at least up to `precision` digits.
+    // Whole seconds keep the declared precision; nonzero fractions are never truncated.
```
leaves12138, line 485: [P1] Preserve reads of timestamp tables written by the released 0.3.0

This shared formatter changes existing-file lookup as well as new writes. The format note acknowledges old Rust directories, but these spellings are also produced by the published `pypaimon-rust==0.3.0`, not just development builds. I installed that wheel and committed two `TIMESTAMP(3)`-partitioned tables: non-legacy `2026-09-15 12:00:00`, and legacy `2026-09-15 12:00:00.100`. Both SQL reads succeed with 0.3.0. Reading the same unchanged warehouses using a wheel built from this head fails with filesystem `NotFound`: the first file exists under `ts=2026-09-15 12%3A00%3A00/` but is sought under `...00.000/`; the second exists under `ts=2026-09-15T12%3A00%3A00.1/` but is sought under `...00.100/`.

The committed metadata does not record an explicit path for these local data files, so an upgrade cannot recover their original locations from the new formatter alone. Please keep Java-canonical paths for new writes while adding a read-side fallback for the old Rust spelling when the canonical file is absent, preserving explicit-path precedence. Add an upgrade test that writes with the released wheel and reads with the new one; fresh-table round trips do not exercise this failure. The same compatibility requirement applies to the legacy fractional-width change above.

leaves12138 (changes requested):
The new formatting agrees with Java in the cases I tested. Validation at 5c5acff: 32 partition utility tests and 14 index-path tests passed; a locally rebuilt binding passed all 19 write tests. I also generated 540 timestamp partition paths using Java's InternalRowPartitionComputer/PartitionPathUtils (precisions 0 through 9, both legacy modes, fractional-width boundaries and pre-epoch values), then verified physical directories and SQL readback in 20 additional binding tests; all passed.

However, I reproduced an upgrade regression using tables actually written by the published pypaimon-rust 0.3.0 wheel. Both tables remain readable with 0.3.0 but fail with this PR because existing data files are looked up under the new directory spelling. This affects committed data, not just newly generated paths. Please retain read compatibility for those files; details are inline. These are targeted checks, not the full CI suite.

crates/paimon/src/spec/partition_utils.rs
leaves12138: With the requirement clarified, Java is the authoritative format and compatibility with the incorrect directory spellings written by Rust 0.3.0 is intentionally out of scope. I am withdrawing this blocking request; no fallback for those spellings is needed. The Java parity and new-write/read validations passed, and I have no other blocking findings at 5c5acff.

leaves12138 (approved):
Approved with Java as the authoritative format. My previous request to preserve the incorrect Rust 0.3.0 directory spellings is withdrawn under the clarified scope. At this unchanged head, the previously completed Java parity checks (540 cases), 46 Rust tests, and 19 binding write tests passed. No remaining blocking findings.

PR #834, opened 2026-09-15
fix(core): preserve local timezone semantics for TIMESTAMP_LTZ schemas

JingsongLi:
### Purpose

Preserve local-timezone semantics when reading PyPaimon schemas containing `TIMESTAMP_LTZ(p)`. Rust currently rejects that spelling in `LocalZonedTimestampType` and then accepts it as an ordinary `TimestampType` because both share the `TIMESTAMP` prefix. In a non-UTC process, native planning consequently computes the wrong partition directory and can select a stale table-level deletion vector.

For example, with `partition.legacy-name=false` and `TZ=Asia/Shanghai`, an LTZ instant at `2026-09-15 20:00:00 UTC` belongs under `2026-09-16 04:00:00.000`, but the mistaken type keeps `2026-09-15 20:00:00.000`. This was reproduced while fixing LTZ paths in apache/paimon#9825.

### Brief change log

- Recognize `TIMESTAMP_LTZ(p)` as `LocalZonedTimestampType`, preserving precision and nullability.
- Reject LTZ spellings in the ordinary timestamp parser so they cannot silently lose timezone semantics.
- Keep serialization in Java's canonical `TIMESTAMP(p) WITH LOCAL TIME ZONE` form.

### Tests

- Type tests: 15 passed. The new regression covers precisions 0/3/6/9, nullable and non-nullable fields, both LTZ spellings, canonical serialization, and ordinary timestamp behavior. It failed before the fix.
- Partition utilities: 32 tests passed.
- Python binding writes: 21 passed, including two new cases that reload PyPaimon's schema alias, write actual partitioned files, inspect directory names, plan, and read through SQL. Both new cases failed with the old wheel.
- With a rebuilt wheel and the corresponding Python path fix, 299 related PyPaimon tests passed, including isolated UTC/Asia-Shanghai LTZ cases for Python and native planning, repeated deletes, and historical reads.
- Rust fmt/clippy, Python Flake8, and `git diff --check` passed.

### API and Format

No public API or binary encoding changes. LTZ aliases now retain their intended timezone semantics instead of being interpreted as ordinary timestamps. Canonical serialized type names are unchanged. Files previously written using the mistaken type's partition directory are not automatically moved.

### Documentation

Added a parser comment explaining PyPaimon's LTZ schema spelling.


leaves12138 (approved):
LGTM. Reviewed commit `2cdab04916891e1486e80d5c9a55a85d4cb53637` against the Java implementation as the reference.

The alias now resolves to `LocalZonedTimestamp` without losing precision or nullability, while serialization retains the canonical `TIMESTAMP(p) WITH LOCAL TIME ZONE` spelling. Ordinary timestamps remain separate.

Validation on this commit:
- Rust type tests: 15 passed; partition utility tests: 32 passed; `cargo fmt --all -- --check` passed.
- Built the Python wheel and ran all 21 write tests in both UTC and Asia/Shanghai: all passed.
- Ran 48 additional end-to-end cases covering alias/canonical schemas, legacy/non-legacy partition names, millisecond/microsecond precision, whole seconds, fractional values, cross-day timezone conversion, and pre-epoch values. Physical partition paths matched output from Java `InternalRowPartitionComputer`, and fresh scan planning and SQL readback succeeded.

No blocking issues found.

PR #836, opened 2026-09-15
fix(io): flush disk cache blocks before publishing

### LINKED ISSUE #835: Disk cache blocks can be published before writes complete
`DiskCache::put_block` drops and renames a Tokio file immediately after `write_all().await`, without waiting for the background write to finish. A reader can observe an empty published block, discard it as invalid, and return a cache miss.

The persistence test failed at its first read in [PR #819's CI run](https://github.com/apache/paimon-rust/actions/runs/34704396500/job/103581714472): expected `Some(b"persistent block")`, got `None`.

On macOS arm64, delaying only the real cache write by 50 or 500 ms reproduced the same assertion in 10/10 runs at each delay. The recorded order was: write started, rename succeeded, read returned zero bytes, cache file removed, write completed. No read results or I/O errors were fabricated. The historical Ubuntu run has no I/O trace, so its exact scheduling remains unconfirmed.

Cache blocks should be published only after the write completes successfully. Writing or flushing failures should leave no published cache entry.

QuakeWang:
### Purpose

Closes #835.

Tokio background writes could finish after a cache block was published, causing empty reads and intermittent persistence test failures.

### Brief change log

- Flush before publishing; reuse cleanup for write and flush errors.
- Verify complete on-disk contents immediately after writing.

### Tests

- 30 controlled runs with 0/50/500 ms write delays.

### API and Format

No changes.

### Documentation

Not needed.


JingsongLi (commented):

Publishing a renamed cache file before its asynchronous writes have crossed Tokio file I/O creates a real empty/partial-read window. The flush is correctly sequenced before publication, and the same cleanup covers write and flush failures. I found no regression in the changed path. Verification: cargo test -p paimon --lib cache::disk::tests (15 passed). Review scope: write ordering, error cleanup, and immediate on-disk visibility.

JingsongLi (approved):
+1

PR #281, opened 2026-04-23
feat: add custom Avro OCF reader for manifest parsing with filtered decoding

JingsongLi:
### Purpose

Replace apache-avro's read path (Value intermediate representation) with a zero-copy custom decoder that reads Avro binary directly into target structs. Key optimizations:

- Custom AvroCursor for zero-copy Avro binary primitive decoding
- OCF parser with snappy/deflate/zstd decompression support
- Writer schema parsing with field index mapping for compatibility
- Two-pass filtered ManifestEntry decoding: lightweight fields first, skip expensive DataFileMeta for filtered-out entries
- SharedSchemaCache for cross-task schema reuse
- Zero-copy entry consumption (into_identifier, into_parts)
- Parallel base+delta manifest list reads via futures::try_join!
- Remove redundant exists() checks before file reads

Benchmark for TableScan: 
1. Full Scan: consistent ~4x speedup across all sizes. Each manifest has ~22,369 entries at ~7.66 MB. The custom Avro OCF decoder with zero-copy cursor and two-pass filtered decoding is paying off nicely.
2. Partial Scan: partition filtering scenarios is highly significant — 14.2x.

PR #842, opened 2026-09-15
fix(spec): support deflate for Avro manifest files

jackylee-ch:
`avro_codec` rejected `deflate` and the OCF reader's codec whitelist did too, so a
table whose schema carries `manifest.compression=deflate` could be neither
written nor read: every scan failed with `avro ocf: unsupported codec: deflate`.
The option comes from the persisted schema, so a table created by Java brings it
along to every paimon-rust process. Java accepts deflate through
`CodecFactory.fromString`.

Read support is a regression: `553e4a3` (#281) advertised "snappy/deflate/zstd"
in its own message while shipping only three arms, replacing a reader that did
handle deflate. Write support is parity work. At HEAD the crate already reads
deflate-compressed Avro *data* files, so only manifests were refused.

No upstream report; found by inspection. `bzip2` and `xz` are the same gap and
stay rejected. Drive-by: the snappy mismatch message said CRC32C where the code
computes CRC-32.


JingsongLi (commented):

A Java-created table can persist manifest.compression=deflate, so read and write support is needed for cross-engine operation. The raw RFC 1951 decoder is scoped to the Avro OCF codec, preserves rejection of unsupported codecs, and the writer/reader codec sets remain compatible. I found no regression in the changed paths. Verification: cargo test -p paimon --lib test_deflate_manifest_compression_round_trips (passed). Review scope: OCF framing, strict decode failure, manifest-list and manifest commit/read-back.

JingsongLi (changes requested):
-1 No need to do this. No use case.

PR #843, opened 2026-09-15
fix(spec): decode _GLOBAL_INDEX fields by name, not by position

jackylee-ch:
`decode_nullable_global_index` read five of the six `_GLOBAL_INDEX` fields
positionally, guarding only `_SOURCE_META` by name. #831 gave the sibling
deletion-vector record in the same function a name-based walk over the writer
schema and left this one behind.

Nothing in a manifest says how many fields the record has. Java tried a runtime
field-count check when `_SOURCE_META` arrived (#8549), replaced it with a
serializer version (#8952), reverted that (#9004) and deleted the versioned
serializer (#9039) — shape compatibility is delegated to schema resolution, which
positional decoding cannot do.

Latent, not live: `_GLOBAL_INDEX` is the last field, so an unconsumed seventh
would corrupt the *next* record in the block, and single-entry manifests decode
fine today. The arms also now honour each field's declared nullability instead of
assuming it.


QuakeWang (approved):
+1

PR #844, opened 2026-09-15
fix(go): prevent native crashes from concurrent record reader access

XiaoHongbo-Hope:
### Purpose

Concurrent calls to NextRecord on one Go RecordBatchReader can poll the same Rust Arrow stream simultaneously. A concurrent Close can also free the stream while it is being polled, causing a native panic or segmentation fault.

### Changes

- Serialize NextRecord and Close per reader.
- Keep separate readers fully parallel.
- Cover concurrent reads and read/close races with regression tests.

### Verification

- Concurrent regression tests: 100 iterations passed.
- Race-enabled regression tests: 10 iterations passed.
- go vet passed for the binding and test modules.


QuakeWang (approved):
LGTM

PR #848, opened 2026-09-16
fix(spec): preserve manifest extra files

jianguotian:
## Purpose

Recent Paimon Java manifests can own auxiliary files through the nullable `_EXTRA_FILES` field. Rust must preserve that field when reading and rewriting manifest metadata, otherwise a cross-language rewrite can lose lifecycle ownership of manifest sidecars.

## Changes

- Add nullable `extra_files` to `ManifestFileMeta` using the Java field name and schema position.
- Preserve the distinction between a legacy missing/null value and an explicitly empty list.
- Include manifest-owned auxiliary files in referenced-file collection and file-size accounting.
- Add Avro round-trip, legacy compatibility, and referenced-file lifecycle tests.

## Verification

- Targeted manifest metadata, manifest-list, and referenced-file tests

## Related Java work

- apache/paimon#9746


QuakeWang (approved):
LGTM

PR #849, opened 2026-09-16
feat(file-index): read range bitmap indexes

jianguotian:
## Purpose

Paimon Java can write `range-bitmap` file indexes, but the Rust reader currently ignores them. As a result, selective predicates cannot turn those indexes into row selections, including for Mosaic row-group scheduling.

## Changes

- Decode the Java V1 range-bitmap header, chunk dictionaries, and BSI bitmaps.
- Evaluate equality, set, range, between, and null predicates into `FileIndexResult::Selection`.
- Support Java-compatible primitive, decimal, date/time, timestamp, and string encodings.
- Preserve Rust predicate semantics for signed zero and fail open for NaN cases that cannot be proven safely.
- Correctly select every row for all-null chunks and validate offsets, bitmap cardinality, and payload bounds.
- Fail open on malformed or newer payloads rather than incorrectly pruning data.

This PR is query-side only: it reads Java-written range-bitmap indexes. A Rust writer can be added independently.

## Verification

- `cargo test -p paimon --lib`: 2,821 passed, 2 ignored
- Java-generated golden payloads for integer, string, and float columns
- Regression coverage for all-null chunks, truncation, signed zero, NaN, and each supported predicate

## Related Java work

- apache/paimon#9140
- apache/paimon#9141
- apache/paimon#9742


crates/paimon/src/file_index/range_bitmap.rs
```diff
@@ -0,0 +1,1491 @@
...
+            PredicateOperator::IsNotNull if literals.is_empty() => self.bsi.existing.clone(),
+            PredicateOperator::Eq if literals.len() == 1 => {
+                self.eq(&self.codec.value(&literals[0])?)
+            }
+            PredicateOperator::NotEq if literals.len() == 1 => {
+                let value = self.codec.value(&literals[0])?;
+                if value.is_nan() {
+                    self.bsi.existing.clone()
+                } else {
+                    self.not(&self.eq(&value))
```
QuakeWang, line 942: `eq()` matches both signed zeros, but the Arrow row filter treats them as distinct. For `f != +0.0`, this drops `-0.0` even though the row filter would keep it. I reproduced this with `JAVA_FLOAT_V1` and `evaluate_exact_leaf_predicate`.

`NOT IN` and strict comparisons have the same issue (`f < +0.0` loses `-0.0`). Could we keep these rows, or return `Remain`, and add a test against the row filter?

crates/paimon/src/file_index/range_bitmap.rs
```diff
@@ -0,0 +1,1497 @@
...
+            PredicateOperator::In => self.literals_bitmap(literals, false)?,
+            PredicateOperator::NotIn => self.not(&self.literals_bitmap(literals, true)?),
+            PredicateOperator::Lt if literals.len() == 1 => {
+                self.lt(&self.codec.value(&literals[0])?)
+            }
+            PredicateOperator::LtEq if literals.len() == 1 => {
+                self.lte(&self.codec.value(&literals[0])?)
+            }
+            PredicateOperator::Gt if literals.len() == 1 => {
+                self.gt(&self.codec.value(&literals[0])?)
```
QuakeWang: NaN ordering still causes false negatives. With `JAVA_FLOAT_V1`, `f > f32::from_bits(0xffc00000)` matches all four non-null rows in `evaluate_exact_leaf_predicate`, but the index returns an empty selection because the literal is normalized to positive NaN.

Negative NaN values in the data can also be lost for `f < 0.0`. Could we keep pruning conservative for these cases and add regression coverage?

QuakeWang (approved):
LGTM

PR #853, opened 2026-09-16
fix(parquet): reject unknown file.compression instead of ignoring it

jackylee-ch:
`parse_compression` fell back to `Compression::UNCOMPRESSED` for any unrecognized `file.compression`. Nothing on the plain-parquet path re-checks the codec, so `gzip2` — or a stray space in `'zstd '` — wrote raw pages and recorded codec id 0 in the footer, leaving the finished file indistinguishable from `file.compression=none`.

Java rejects the same value where the per-file writer is built: `RowDataParquetBuilder#createWriter` -> `CompressionCodecName#fromConf`.

Accepted: `none, uncompressed, snappy, gzip, gz, lz4, lz4_raw, zstd`. `lz4_raw` is added because Java accepts it, so a Java-created table can carry it — that also changes its output from uncompressed to LZ4_RAW. `brotli` and `lzo` are rejected: no Java codec class ships for either, and `LZO` has no arrow-rs encoder.

A Rust-only table with a typo'd codec now fails on the first write; Java would already have rejected it.


JingsongLi (commented):

Rejecting an unrecognized persisted file.compression value before opening the writer avoids producing a file whose footer silently records UNCOMPRESSED, and the change preserves the supported codec mappings and Zstd-level fallback. I found no regression in the changed path. Review scope: writer construction and the new codec/error-path coverage.

JingsongLi (approved):
+1

PR #860, opened 2026-09-17
fix(write): avoid double separators in bucket paths

shyjsarah:
## Summary

- route append, primary-key, and postpone writers through the shared `bucket_path_under` helper
- avoid generating `partition//bucket-*` when partition paths already end in `/`
- add a provider-backed regression test that rejects paths OpenDAL would normalize

## Why

`PartitionComputer::generate_partition_path` returns partition directories with a trailing slash. The three writers added another separator while constructing their bucket directories. OpenDAL used to normalize those paths silently, but URI-aware `FileIOProvider` validation correctly rejects them because normalization can change an object key.

## Tests

- `cargo test -p paimon partitioned_writer_does_not_create_double_slash_bucket_paths`
- `cargo test -p paimon --lib`: 2848 passed, 2 ignored; one pre-existing `snapshot_manager` HTTP mock test failed intermittently with connection reset and passed when rerun individually


jerry-024 (approved):
+1

PR #851, opened 2026-09-16
perf(mosaic): prefetch row groups with a Mosaic-owned budget

jianguotian:
## Purpose

The Rust Mosaic reader currently opens row groups serially. Opening a group performs dependent range reads and decompression, so object-store latency remains on the critical path even after manifest and row-selection pruning.

## Design

This implementation follows the ownership boundary of apache/paimon#9740:

- `MosaicFormatReader` owns Mosaic row-group prefetching, its executor, and its decoded-byte/concurrency budget.
- `ParquetReadBudget` remains Parquet-only; Mosaic does not depend on or extend it.
- Row groups are planned after row-selection and statistics pruning, then only surviving groups are prefetched.
- Completed groups are emitted in file order, and all scheduled work is joined on cancellation, errors, or panics.

## Changes

- Open up to 8 Mosaic row groups concurrently while preserving output order.
- Bound queued/decoded row groups with a Mosaic-owned 64 MiB estimated-byte budget.
- Estimate decoded bytes from projected field types and row counts, matching Java Mosaic's sizing model.
- Always admit the head group so an undersized byte budget cannot stall progress.
- Keep residual predicate evaluation and batch emission ordered on the consumer thread.
- Keep the default values aligned with Java:
  - `mosaic.read.prefetch-row-groups = 8`
  - `mosaic.read.prefetch-max-bytes = 64 MiB`

Rust's format-reader factory does not yet expose format-scoped table options, so this PR establishes the bounded behavior with those defaults; option plumbing remains a separate API change.

## Read-only scope

- This PR only changes the Mosaic file read path (`MosaicFormatReader` / `FormatFileReader::read_batch_stream`).
- It does not add or change Mosaic writer, commit, compaction, or other write-path behavior.
- `MosaicWriter` is used only in tests to create read fixtures.

## Verification

- `cargo test -p paimon --lib arrow::format::mosaic::tests -- --nocapture`: 37 passed
- `cargo test -p paimon --lib`: 2,821 passed, 2 ignored
- `cargo clippy -p paimon --lib --tests -- -D warnings -A clippy::unnecessary-cast`
- `cargo fmt --all --check`
- `git diff --check`
- Boundary check: no Mosaic symbol in `parquet_read_budget.rs`, and no `ParquetReadBudget` reference in `mosaic.rs`

## Related Java work

- apache/paimon#9740
- apache/paimon#9742
- apache/paimon#9744

PR #864, opened 2026-09-17
feat(mosaic): read row-group prefetch depth and byte budget from table options

sundapeng:
### Purpose

apache/paimon-rust#851 added Mosaic row-group prefetch with the depth and decoded-byte budget hardcoded to Java's defaults (8 row groups, 64 MiB), because the format-reader factory did not receive format-scoped table options. This change reads the same two table options Java defines in `MosaicFileFormat`, so the setting a table carries applies to both engines:

- `mosaic.read.prefetch-row-groups` (default 8; `0` reads row groups on demand)
- `mosaic.read.prefetch-max-bytes` (default 64 MiB; estimated decoded bytes allowed in flight beyond the head row group)

Both are parsed in `CoreOptions` next to the Parquet row-group budget options and reach `MosaicFormatReader` through `DataFileReader`, so every read path that opens Mosaic files (plain append, key-value merge, data evolution, format tables, audit log) honours them.

Following review, Parquet and Mosaic now share one budget type. `ParquetReadBudget` became `ReadBudget` in `arrow/`: it still bounds in-flight row groups and their estimated bytes, and it now offers a non-blocking `try_acquire` and a `acquire_blocking` next to the async `acquire`, which is what the Mosaic reader needs because it decodes on its own threads. Mosaic's private `MosaicPrefetchBudget` is gone. The two option pairs stay separate on purpose: `mosaic.read.prefetch-*` are Java's keys with Java's defaults and bound the look-ahead inside one file, while `read.parquet.row-group.*` are paimon-rust's own and bound one scan across files. Folding the Mosaic keys into the Parquet ones would make a Java-tuned table read differently here, which is the gap this PR closes.

### Brief change log

- `CoreOptions::mosaic_read_prefetch_row_groups()` and `mosaic_read_prefetch_max_bytes()` with the same validation style as `read.parquet.row-group.*` (`0` is allowed for the depth, the byte budget must be positive).
- `arrow/parquet_read_budget.rs` renamed to `arrow/read_budget.rs`; `ParquetReadBudget`/`ParquetReadPermit` renamed to `ReadBudget`/`ReadPermit`; added `try_acquire`, `acquire_blocking`, a byte granularity (Parquet keeps MiB permits, Mosaic charges exact bytes) and the option name used in the oversized-row-group warning.
- `MosaicFormatReader` takes `MosaicPrefetchOptions` and builds a per-file `ReadBudget` from them; `MosaicPrefetchBudget` and its permit are deleted.
- `DataFileReader::with_mosaic_prefetch`; `configured_mosaic_prefetch(table)` plumbs the options from the table schema in `TableRead`, `KeyValueReadConfig`, `DataEvolutionReader`, `FormatTableRead` and the audit-log read.
- No behaviour change without the options: the defaults equal the values #851 hardcoded.

### Tests

- `spec::core_options::tests::test_mosaic_read_prefetch_options`: defaults, custom values, invalid values.
- `arrow::read_budget::tests::try_acquire_is_cumulative_and_an_oversized_head_takes_the_whole_budget` and `acquire_blocking_waits_off_the_runtime_until_a_permit_is_released` cover the two paths Mosaic uses; the existing async tests still cover the Parquet path.
- `arrow::format::mosaic::tests::test_format_reader_prefetch_options_control_look_ahead`: depth `0` reads on demand (one storage read in flight), the default overlaps reads, both return the same rows.
- `table::table_read::tests::test_mosaic_prefetch_options_come_from_table_options`: the values come from the table options and invalid values fail the read.
- Ran `cargo fmt --all -- --check`, `cargo clippy --locked -p paimon --all-targets --features fulltext,vortex -- -D warnings`, and `cargo test --locked -p paimon --lib`.

Measured on a StarRocks-derived engine reading an 11,374-column Mosaic table through the C bindings: with #851 the Rust scan of the same 54 splits went from 7.1 s to 3.4 s (the Java reader with the same prefetch reads them in 1.2 s), so exposing the knobs lets a table tune depth and budget per workload.

### API and Format

Two new table options, `mosaic.read.prefetch-row-groups` and `mosaic.read.prefetch-max-bytes`, matching the keys and defaults of the Java implementation. `paimon::arrow::ParquetReadBudget` is renamed to `paimon::arrow::ReadBudget`; no other public signature changes. No format change.

### Documentation

None; the option keys mirror `MosaicFileFormat` in Apache Paimon.


JingsongLi (commented):
Could we modify the Parquet budget into a generic budget that both Parquet and Mosaic can use?

sundapeng:
Done - the budget is now one type, `ReadBudget` in `arrow/read_budget.rs`, and both formats take it.
I kept the two option pairs separate, because they are not the same resource:

|                | `read.parquet.row-group.*`                                    | `mosaic.read.prefetch-*`                                   |
| -------------- | ------------------------------------------------------------- | ---------------------------------------------------------- |
| defined by     | paimon-rust only (#633)                                         | Java `MosaicFileFormat` (apache/paimon#9740), 8 / 64 MiB    |
| scope          | one budget per `TableRead`, shared by every file of the scan    | one budget per open file                                    |
| bytes          | footer *uncompressed* size of the projected chunks              | *estimated decoded* Arrow size of the row group              |
| value `0`      | rejected (`1` disables concurrency)                             | valid, "read row groups on demand" - Java's semantics       |
| active         | only on the predicate-free, selection-free path                 | always                                                      |
| acquired from  | async tasks                                                     | the blocking Mosaic decode threads                          |

The Mosaic keys are Java's keys with Java's defaults, so a table tuned on the Java side has to read the same
way here - that is what this PR fixes. If Rust honoured only a generic key, a Java-tuned table would
silently read differently.

So the sharing is in the mechanism: `ParquetReadBudget` is now `ReadBudget`, it grew `try_acquire` and
`acquire_blocking` next to the async `acquire` (the Mosaic reader decodes on its own threads, so it cannot
await), plus a byte granularity - Parquet keeps MiB permits for a scan-wide budget, Mosaic charges the exact
estimated bytes of a row group. Mosaic's private `MosaicPrefetchBudget` and its permit are deleted; the
reader builds a per-file `ReadBudget` from the two Java options. `paimon::arrow::ParquetReadBudget` is
renamed accordingly, no other public signature changes, and no behaviour change on either format.

The branch is also rebased on main (it had conflicted with the BLOB read parallelism change).

JingsongLi (approved):
+1

PR #866, opened 2026-09-17
fix(table): refuse row-level DML on a format table

### LINKED ISSUE #865: [Bug] UPDATE, DELETE and MERGE INTO on a format table leave Paimon files behind and inflate the row count
### Search before asking

### Please describe the bug 🐞

`UPDATE`, `DELETE` and `MERGE INTO` on a Format Table write Paimon data files into the table, and the table then reads more rows than it holds.

The three statements go through the copy-on-write path. `CopyOnWriteMergeWriter` rewrites the rows it keeps into `<partition>/bucket-0/` under the table directory, then the commit tries to publish them. A Format Table has no snapshot, so the commit fails, but the rewritten files stay where they are. A Format Table scan lists a partition directory recursively and `bucket-0` is not a hidden name, so those files are read as table data.

What we measured on a 3-row Format Table (one row matched by the statement):

| statement | result | rows after |
|---|---|---|
| `UPDATE t SET ... WHERE ...` | error `CoW UPDATE failed after 5 retries due to concurrent compaction ... Delete conflict: file ... does not exist in the current snapshot` | 33 |
| `MERGE INTO t ... WHEN MATCHED THEN UPDATE SET ...` | same error for MERGE INTO | 33 |
| `DELETE FROM t WHERE ...` | same error for DELETE | 8 |

It happens with and without `metastore.partitioned-table`, and the files are visible to every engine that reads the table, not only to this client.

`INSERT` into a Format Table is already refused by `FormatWriteBuilder::new_write`, but the copy-on-write path never asks the write builder for a writer, so it goes around that check.

### Solution

Refuse the rewrite for a Format Table in `CopyOnWriteMergeWriter::new`, next to the existing checks for primary keys and data evolution. It is the single place the three statements pass through, and it runs before any file is written, so nothing is left behind. Java refuses the same statements: `PaimonFormatTable` does not implement Spark's row-level operations.

sundapeng:
### Purpose

Linked issue: close #865

`UPDATE`, `DELETE` and `MERGE INTO` on a Format Table wrote Paimon data files into the table and left them there, so the table read more rows than it held. On a 3-row table, `UPDATE` and `MERGE INTO` returned 33 rows afterwards and `DELETE` returned 8, with and without `metastore.partitioned-table`. The rewritten files are visible to every engine that reads the table.

The three statements take the copy-on-write path. `CopyOnWriteMergeWriter` rewrites the rows it keeps into `<partition>/bucket-0/`, then the commit tries to publish them. A Format Table has no snapshot, so the commit fails with `Delete conflict: file ... does not exist in the current snapshot`, but the files stay, and a Format Table scan lists a partition recursively where `bucket-0` is not a hidden name. `INSERT` is already refused by `FormatWriteBuilder::new_write`; the copy-on-write path never asks the write builder for a writer, so it went around that check.

This refuses the rewrite in `CopyOnWriteMergeWriter::new`, which is the one place the three statements pass through and which runs before any file is written. Java refuses the same statements: `PaimonFormatTable` does not implement Spark's row-level operations.

### Brief change log

- `CopyOnWriteMergeWriter::new` returns `Unsupported` for a Format Table, next to its existing checks for primary keys and data evolution. The message names the table and the three statements.
- `docs/src/sql.md` says under Format Table Partitions that `INSERT`, `UPDATE`, `DELETE` and `MERGE INTO` are all refused.

### Tests

`crates/integrations/datafusion/tests/rest_format_table_dml.rs`, both cases running all three statements:

- a Format Table whose partitions a REST catalog manages;
- a Format Table that discovers its partitions from directories, in a filesystem catalog.

Each case asserts the error, that the set of files under the table is byte-for-byte the one it started with, and that a following `SELECT` still returns the original rows. With the new check removed, both fail: the statements report success and the files appear.

```
cargo fmt --all -- --check
cargo clippy --locked -p paimon -p paimon-datafusion --all-targets --features paimon/fulltext,paimon/vortex -- -D warnings
cargo test -p paimon-datafusion --test rest_format_table_dml   # 2 passed
cargo test -p paimon-datafusion --test delete_tests            # 25 passed
cargo test -p paimon-datafusion --test update_tests            # 20 passed
cargo test -p paimon-datafusion --test merge_into_tests        # 29 passed
cargo test -p paimon-datafusion --test append_merge_into       # 34 passed
cargo test -p paimon --lib cow_writer                          # 9 passed
```

### API and Format

`CopyOnWriteMergeWriter::new` now returns an error for a Format Table, where it used to build a writer. No storage format change.

### Documentation

`docs/src/sql.md`, one paragraph under "Format Table Partitions".

Related: #863 fixes TRUNCATE TABLE on a Format Table, which had the same shape (the statement reported success while doing nothing). The two do not overlap in code.


QuakeWang (approved):
+1

PR #868, opened 2026-09-18
perf(btree): optimize sparse IN queries

JingsongLi:
### Purpose

Linked issue: N/A

Improve sparse BTree IN queries so they do not sequentially scan unrelated key ranges between distant targets.

### Brief change log

- Sort, deduplicate, and file-bound target keys before lookup.
- Resolve target keys through the in-memory index block and group them by data block.
- Read each target data block once and binary-seek each requested key within that block.
- Remove the previous merge-style sequential scan across intermediate blocks and entries.

### Tests

- Added a sparse multi-block test that covers unsorted, duplicate, missing, and out-of-bounds keys while asserting only target blocks are read.
- Added a large single-block test that verifies sparse keys use logarithmic seeks instead of a linear entry scan.
- cargo test -p paimon --lib btree::
- cargo clippy -p paimon --lib -- -D warnings
- cargo fmt --all -- --check

### API and Format

No API or storage-format changes.

### Documentation


leaves12138 (approved):
Reviewed commit a986b6e67c70c0d309b983f640a63bee3ebd7ce8. LGTM; no blocking issues found.

The index-block seeks and coalescing of target block handles remove the sparse-IN read amplification without changing the public API or persistent format. I checked the Java writer's one-entry-per-key invariant, so seeking one entry still retrieves all row IDs for that key.

Validation:
- `cargo test -p paimon --lib --locked`: 2,867 passed, 2 ignored.
- `cargo fmt --all -- --check` and `cargo clippy -p paimon --lib --locked -- -D warnings`: passed.
- Additional local checks passed for 600 IN/NOT IN query sets across four compression modes and five block sizes, including negative and missing keys, duplicates, file boundaries, dense and sparse targets, large posting lists, and row IDs above 2^32.
- Additional Java-generated integer/string fixture comparisons and all-null/empty-target checks passed.

PR #869, opened 2026-09-18
feat(btree): add optional bloom filters

JingsongLi:
## Summary

- add the optional `btree-index.bloom-filter.enabled` setting, disabled by default
- write and lazily read Java/Python-compatible Bloom filters in BTree index files
- use exact point lookups for equality and IN predicates so Bloom-negative keys skip data-block reads
- document the option and cover compatibility, I/O, concurrency, and end-to-end build behavior

This ports the behavior introduced by apache/paimon#9946.

## Testing

- `cargo fmt --all`
- `git diff --check`


leaves12138 (approved):
LGTM. Reviewed commit 55213e9baccbe54fec3f4c0872fff98c6451d9b5 against the Java format and implementation in apache/paimon#9946. No blocking issues found within the agreed review scope.

Validation:
- `cargo test -p paimon --lib --locked`: 2,873 passed, 2 ignored.
- `cargo fmt --all --check` and `cargo clippy -p paimon --lib --tests --locked -- -D warnings`: passed.
- Additional differential checks: 15,360 equality lookups and 480 IN query sets across Bloom on/off, three block sizes, and none/zstd/lz4/lzo compression, including duplicate postings, missing keys, out-of-range keys, null rows, and row IDs above 32 bits.
- Java/Python/Rust file interoperability checks passed. Java and Python each read all 80 exported Rust files successfully; the Rust-written Bloom bytes also matched Java's builder exactly.
- Checked default-disabled behavior, preservation of existing constructors, IN data-block coalescing, and shared lazy Bloom loading for concurrent lookups.

PR #870, opened 2026-09-18
[python] Expose streaming native table reader

JingsongLi:
## What changed

- expose TableRead.read_arrow for lazy PyArrow RecordBatch iteration
- keep the existing eager read API for compatibility while sharing validation and stream construction
- release the GIL while advancing the Rust stream and add Python type stubs

## Why

PyPaimon can already ask Rust to plan splits, but native data reads need a lazy batch interface so to_arrow_batch_reader does not materialize the entire result in Rust first.

## Tests

- cargo test -p pypaimon_rust --lib
- bindings/python/.venv/bin/python -m pytest -q bindings/python/tests/test_read.py


leaves12138 (approved):
Reviewed at `4788600e5e580ea269ddf83d608d8ea9672da5d1`. No blocking issues found.

The new iterator preserves lazy data reads while sharing configuration validation and stream construction with the existing eager API. The stream mutex is acquired with the GIL released, and Arrow-to-Python conversion happens after the stream lock is released.

Validation:
- Built and imported the Python extension from this commit.
- 278 Python tests passed: all 186 existing binding tests, 80 read tests rerun through `list(read_arrow(...))` for eager/streaming parity, and 12 additional streaming checks.
- Additional checks cover deferred missing/corrupt-file errors, repeated EOF, mixed `next()`/`read_next_batch()` calls, generator/invalid split inputs, independently advancing readers, batch lifetime after dropping the reader, and concurrent consumers without lost or duplicated rows.
- A local S3-compatible HTTP server running in Python served actual Parquet reads, confirming that stream advancement releases the GIL. Constructing `read_arrow()` issued no data requests before consumption.
- `cargo fmt --all -- --check` passed.

Local validation limitation: `cargo test -p pypaimon_rust --lib` could not run because the installed Python lacks a shared libpython required by PyO3's `auto-initialize` feature. The Python tests above exercised the successfully built extension; the full suite used an isolated warehouse fixture and its optional media dependencies.

PR #871, opened 2026-09-18
feat(python): configure native BLOB read parallelism

JingsongLi:
Follow-up to #870.

This exposes `with_blob_parallelism` through the Rust and Python read builders, and applies the configured limit to both direct `.blob` reads and Data Evolution descriptor resolution. The existing byte budget remains in place.

The setting is propagated through Paimon/format table reads, Data Evolution fallback reads, and diff/incremental paths.

Tests:
- focused direct/descriptor BLOB concurrency tests
- `cargo test -p paimon --lib blob --no-fail-fast` (128 passed; the 2 sandbox-bound REST tests passed separately outside the sandbox)
- `pytest bindings/python/tests/test_read.py -q` (81 passed)
- Apache Paimon native/Data Evolution integration tests against the locally built binding


leaves12138 (approved):
LGTM for the BLOB parallelism changes at 837ced49576dd05eaf76fa0d2b6d1b7facb92179. I did not find a blocking issue introduced by this PR.

Checked the Rust/Python builder plumbing, default behavior, zero-value validation, direct BLOB readers, and Data Evolution descriptor/fallback paths. The descriptor byte budget remains unchanged.

Validation on the PR head:
- `cargo fmt --all -- --check` passed.
- `cargo test -p paimon --lib blob --no-fail-fast`: 130 passed.
- Built and installed the native Python extension; the complete Python binding suite passed (187 tests).
- Added 16 review-only end-to-end checks against a local S3-compatible HTTP server. For both direct `.blob` reads and external descriptors, observed request concurrency matched 1, 2, and 16, and remained 8 by default. These checks covered two BLOB columns, multiple descriptor URIs, both `read()` and `read_arrow()`, result order/content, and changing the builder after `new_read()` without affecting the existing reader.

CI note: the current `integration (python)` failure is also present on the base commit b567b6e6d916078e49f17b55b50c73a78013a307. Both runs fail the same two `test_ltz_schema_alias_write_roundtrip` cases because they expect `LocalZonedTimestamp` while `field_type()` now returns `TIMESTAMP(...) WITH LOCAL TIME ZONE`. That is an existing schema-display/test expectation mismatch, not a regression from this BLOB change. Those expectations still need to be fixed and CI rerun before merging; this approval does not mean all checks are green.

PR #873, opened 2026-09-18
fix(datafusion): preserve active runtime in blocking callbacks

### LINKED ISSUE #872: Blocking DataFusion callbacks should preserve the active Tokio runtime
### Problem

`block_on_with_runtime` detects when it is called from an entered Tokio runtime, but currently runs the future on the process fallback runtime in a new OS thread and synchronously joins that thread.

Moving the future to a different runtime can stall work that depends on the active runtime. It also differs from the behavior of `runtime()` and `await_with_runtime`, which preserve an entered runtime.

### Expected behavior

When called from a Tokio multi-thread runtime, the blocking helper should execute the future on that runtime while marking the blocking section so Tokio can continue scheduling work. Calls made without an entered runtime should continue to use the process fallback runtime. The current-thread runtime needs to retain a compatible fallback because `block_in_place` is not supported there.

### Reproduction

A unit test can enter a multi-thread runtime, call `block_on_with_runtime` with a future that returns `Handle::current().id()`, and compare that value with the caller's runtime ID. The current implementation returns a different ID.

shyjsarah:
### Purpose

Linked issue: close #872

Synchronous DataFusion callbacks may invoke `block_on_with_runtime` while already running on a Tokio multi-thread runtime. The helper currently moves the future to the process fallback runtime and waits for a new OS thread, which can stall work that depends on the active runtime.

### Brief change log

- Use `block_in_place` with the active runtime handle for multi-thread runtimes.
- Keep the existing fallback path for current-thread runtimes, where `block_in_place` is unsupported.
- Add a regression test that verifies the blocking helper preserves the active runtime.

### Tests

- `cargo test -p paimon-datafusion --lib runtime::tests -- --nocapture`
- `cargo fmt --all --check`
- `cargo clippy -p paimon-datafusion --lib --tests -- -D warnings`

The full local lib suite completed 353 tests successfully. Seven table scan tests require shared test tables that were unavailable locally and failed with `TableNotExist`.

### API and Format

### Documentation


JingsongLi:
Could we preserve `LocalSet` compatibility here?

`RuntimeFlavor::MultiThread` does not guarantee that `block_in_place` is allowed. A `LocalSet` running on a multi-thread runtime still disallows in-place blocking.

I reproduced this by calling `PaimonCatalogProvider::schema_names()` inside `LocalSet::run_until(...)`. The same test passes with the previous implementation, but this change causes it to panic at `block_in_place`:

```text
can call blocking only when running on the multi-threaded runtime
```

The single-worker starvation fix works as intended, but this introduces a regression for callers using `LocalSet`. Could we add a regression test and retain a compatible execution path for contexts that disallow in-place blocking, rather than relying solely on `runtime_flavor()`?

crates/integrations/datafusion/src/runtime.rs
```diff
@@ -103,14 +103,18 @@ where
     F: Future + Send + 'static,
     F::Output: Send + 'static,
 {
-    if Handle::try_current().is_ok() {
-        let handle = global_runtime().handle().clone();
-        std::thread::spawn(move || handle.block_on(future))
+    if let Ok(handle) = Handle::try_current() {
+        if handle.runtime_flavor() == RuntimeFlavor::MultiThread {
```
jerry-024, line 107: **[Major] Checking only `RuntimeFlavor::MultiThread` can still panic here.**

`LocalSet::run_until` may run on a multi-thread runtime while explicitly disabling `block_in_place`. `Handle::try_current()` still returns that multi-thread handle, so this branch panics instead of taking the previous fallback path.

Please preserve a non-panicking fallback when in-place blocking is disallowed and add a `multi_thread` + `LocalSet::run_until` regression test.

shyjsarah: Thanks, this is correct. `runtime_flavor()` describes the runtime implementation, but not whether the current execution context permits `block_in_place`. A `LocalSet` changes the latter, so #873 introduces the regression you described and should not be merged as-is.

I reworked the fix in #881. Instead of trying to detect when in-place blocking is safe, it removes the sync-to-async bridge from the affected catalog discovery callbacks: `schema_names`, `schema`, `table_names`, and `table_exist`. Metadata is loaded or refreshed at async boundaries, while those synchronous callbacks only read an in-memory snapshot.

This avoids both the worker-starvation issue from #872 and the `LocalSet` panic. #881 also tests the stronger invariant that synchronous discovery callbacks do not access the remote catalog, and preserves the last good snapshot when refresh fails.

I propose superseding and closing #873 in favor of #881. Thanks for catching this.

PR #874, opened 2026-09-18
fix(arrow): map BLOB to LargeBinary

JingsongLi:
BLOB values can exceed Arrow Binary 32-bit offsets and PyPaimon already exposes BLOB as LargeBinary. This change maps scalar and nested BLOB values to LargeBinary across format readers/writers, descriptor resolution, data evolution, residuals, variant shredding, and DataFusion BLOB functions. BINARY and VARBINARY remain Binary.

This is required by apache/paimon#9954 so native reads return the requested Python schema directly.

Tests:
- cargo test -p paimon --lib blob (129 passed; two sandbox-only REST token tests excluded)
- cargo test -p paimon --lib table::data_evolution_reader::tests:: (75 passed)
- cargo test -p paimon-datafusion --test blob_tests (25 passed)
- cargo test -p paimon-datafusion --test blob_descriptor_functions (4 passed)
- Python binding tests/test_read.py (81 passed)


leaves12138 (approved):
LGTM. Reviewed the LargeBinary mapping across scalar and nested BLOB reads/writes, descriptor resolution, Data Evolution, and DataFusion functions. BINARY and VARBINARY remain Binary.

Validated locally: 132 BLOB core tests, 75 Data Evolution tests, 29 DataFusion BLOB tests, and 81 Python read tests passed. No blocking issues found.

PR #875, opened 2026-09-19
[rust] Add parquet.filter.columnindex.enabled option to control Page Index pruning

XiaoHongbo-Hope:
## Summary

- honor `parquet.filter.columnindex.enabled` with a Java-compatible default of `true`
- skip ColumnIndex/OffsetIndex loading and predicate page pruning when disabled while preserving row selection and row-group pruning semantics
- propagate the option through raw, primary-key, data-evolution, format-table, audit, and search read paths
- document the request-count versus bytes-read tradeoff

## Tradeoff

The default `true` favors selective scans: predicates and row selections can use Page Indexes to avoid reading unrelated data pages, reducing bytes read. This can issue additional fine-grained range requests on object storage.

Set `parquet.filter.columnindex.enabled=false` when object-store GET QPS or per-request overhead is the bottleneck. Rust then avoids loading ColumnIndex/OffsetIndex and falls back to coarser column-chunk or row-group reads. This reduces fine-grained requests but increases bytes transferred and decoded. Predicate results and explicit row-selection results remain unchanged. Row-group pruning and row-level filtering continue to apply.

This is a read-side optimization switch only; it does not change whether writers generate Page Indexes.

## Tests

- `cargo test -p paimon arrow::format::parquet::tests` (55 passed)
- `cargo test -p paimon spec::core_options::tests` (77 passed)
- `cargo test -p paimon a_dynamic_copy_can_disable_parquet_page_indexes`

The local `TrackingFileRead` tests verify that disabling Page Index keeps predicate and explicit row-selection results unchanged while falling back to coarser reads.


QuakeWang (approved):
+1

JingsongLi:
+1

PR #878, opened 2026-09-20
fix(vortex): reject columns the writer cannot convert instead of panicking

jackylee-ch:
`VortexFormatWriter::new` handed the Arrow schema straight to `DType::from_arrow`, which ends in `unimplemented!()` for any Arrow type it has no arm for. `Map` is one of those, and `paimon_type_to_arrow` builds a `Map` for both MAP and MULTISET — so writing a vortex table with either column aborted the process at writer construction instead of returning an error. Reproduced through `create_table` -> `new_write()` -> `write_arrow_batch`.

Screen the schema before the conversion and return `Error::Unsupported` naming the column, following `validate_mosaic_schema`.

The check is on the Arrow schema, not `write_fields`: `kv_file_writer` and `postpone_file_writer` both pass `write_fields: None`, so a `DataField`-based check would silently skip the primary-key write path.


JingsongLi:
Reviewed head 94a6ac61. Requirement fit: SUPPORTED; implementation: CLEAN. I traced Paimon MAP/MULTISET conversion to Arrow Map and the Vortex 0.75 `DType::from_arrow` fallback to `unimplemented!()`. The new schema check reaches nested Map fields before writer construction and returns an actionable `Unsupported` error; the positive nested-ROW path remains accepted. I ran `cargo test -p paimon --lib --features vortex arrow::format::vortex::tests::test_vortex_writer` (6 passed). No actionable issue found in this head.

JingsongLi (approved):

PR #880, opened 2026-09-20
[rust] Coalesce inline BLOB reads with bounded amplification

JingsongLi:
### Purpose

Reduce inline BLOB request overhead for scalar `BLOB`, `ARRAY<BLOB>`, and `MAP<X, BLOB>` without turning sparse row selection into excessive byte reads or retained buffers.

The reader previously issued one physical range read per selected scalar BLOB entry. Nested inline BLOBs were more expensive: ARRAY decoding performed layout preflight reads plus a full entry read for every row, while MAP decoding performed header, index, key-data, and full-entry reads for every row. On local files this created substantial syscall overhead, and on object storage it would multiply request latency.

### Changes

- Coalesce nearby inline BLOB entry ranges using a 1 MiB maximum gap and an 8 MiB maximum span.
- Bound each merged span to at most 2x the unique selected entry bytes, so sparse selections remain separate reads.
- Copy selected entries out of any merged span containing gaps, allowing the unselected backing bytes to be released as soon as that read resolves; contiguous spans remain zero-copy.
- Fetch independent merged spans concurrently with the configured `blob_parallelism`.
- Restore requested row order after unordered concurrent reads.
- Preserve per-entry magic, embedded-length, and CRC32 validation.
- Decode coalesced inline ARRAY and MAP entries through an in-memory `FileRead`, reusing the existing nested layout validation.
- Keep descriptor mode on metadata-only partial reads so it does not fetch BLOB value bodies.
- Strengthen range-read tests, including data-evolution fallback ranges, nested coalescing, ordering, descriptor skipping, bounded concurrency, sparse amplification, and backing-buffer release.

### Sparse-selection safety

The reported layout of 128 selected 4 KiB entries separated by 1 MiB unselected entries is covered by an exact planner regression test. It now produces 128 ranges totaling 514 KiB instead of 16 merged ranges totaling 112.5 MiB. A separate valid-file test verifies the 2x byte bound, and another test verifies that values from a profitable but gapped merge do not retain the merged response buffer.

### Performance

Release Python binding, local filesystem, 8 splits, 5 measured iterations after one warm-up. All results matched Python schemas, row counts, null counts, and value/byte checksums.

Scalar BLOB, 16,000 rows with 4 KiB payloads:

- native parallelism=1, blob_parallelism=1: 152.1 ms -> 17.5-21.2 ms
- native parallelism=4, blob_parallelism=1: 136.8 ms -> 5.8-6.7 ms
- native parallelism=4, blob_parallelism=4: 134.6 ms -> 6.0-6.4 ms
- the parallel native result was 6.3x to 20.6x faster than the Python reader in the measured configurations

Nested BLOB, 8,000 rows with two 2 KiB values per non-null row:

- ARRAY, parallelism=1: 180.7 ms -> 10.1 ms; Python 72.5 ms
- ARRAY, parallelism=4/blob_parallelism=4: 180.5 ms -> 4.47 ms; Python 133.8 ms
- MAP, parallelism=1: 398.7 ms -> 11.9 ms; Python 85.2 ms
- MAP, parallelism=4/blob_parallelism=4: 351.9 ms -> 5.11 ms; Python 251.8 ms

For 512 scalar rows with 128 KiB payloads in one split, increasing `blob_parallelism` from 1 to 4 reduced the native median from 16.5 ms to 13.8 ms, confirming that separate 8 MiB spans still read concurrently.

### Verification

  - 3013 unit tests passed, 2 ignored
  - all integration and example targets passed
- complete BLOB module: 41 tests passed


leaves12138 (approved):
Re-reviewed c810cb0. The 2x unique-selected-byte limit and copying entries out of gapped spans address the sparse-read amplification and retained-buffer issues, while contiguous spans retain the zero-copy path.

Re-ran the original sparse-selection reproducer: 128 selected 4 KiB entries separated by 1 MiB gaps now read and retain 514 KiB with blob_parallelism=1, matching the previous per-entry path. All 137 targeted BLOB-related tests, including the independent reproducer, passed with storage-fs and storage-memory enabled.

LGTM.

PR #881, opened 2026-09-20
fix(datafusion): snapshot metadata for synchronous catalog callbacks

### LINKED ISSUE #872: Blocking DataFusion callbacks should preserve the active Tokio runtime

shyjsarah:
### Purpose

Refs #872. Supersedes #873 for the metadata-discovery path by fixing the catalog boundary instead of changing Tokio blocking behavior.

DataFusion's metadata discovery APIs are synchronous, while Paimon catalog and external table-engine resolution are asynchronous. Calling those async paths from `schema_names`, `schema`, `table_names`, or `table_exist` can block a DataFusion worker that the query itself needs, which is exposed by filtered `information_schema` queries with repartitioning.

This PR intentionally does not close #872. Synchronous mutation callbacks (`register_schema`, `deregister_schema`, and `deregister_table`) still use the existing blocking bridge; removing that bridge requires a separate lifecycle/API change.

### Brief change log

- Build database, table, view, and system-table capability snapshots asynchronously when catalog/schema providers are initialized or explicitly refreshed.
- Make synchronous catalog discovery callbacks read only the in-memory snapshot; concrete table loading remains asynchronous.
- Bound automatic `information_schema` refreshes with a freshness TTL, failure backoff, per-target single-flight, global concurrency limit, and timeout. A failed or timed-out catalog remains best-effort and cannot block healthy catalogs indefinitely.
- Merge concurrent refreshes per database. If a DDL delta rejects a full-refresh publication, re-read that database with bounded retry instead of silently discarding either side.
- Serialize the remote mutation and local snapshot delta for same-database DDL, including database/table DDL interactions, so response ordering cannot invert backend commit order.
- Apply rename deltas from the authoritative mutation outcome. `ALTER TABLE IF EXISTS` uses a strict backend rename; a missing source succeeds without creating a phantom target and removes a stale-positive source from the snapshot.
- Remove the misleading synchronous provider `new(...)` shims. `try_new(...)` is the initialized constructor; `new_uninitialized(...)` is the explicit deferred path used by Python construction.
- Separate base-object existence from Paimon system-table capability. Native object tables and routed tables no longer advertise `$snapshots`; classification is bounded and cached so unchanged refreshes do not reload table metadata.
- Add `Catalog::list_table_types` with a Paimon-only compatibility default and exact bounded implementations for filesystem and REST catalogs.
- Preserve resolver `Ok(None)` as not found in `table()`, `table_exist()`, and `table_type()`.
- Bound retained database-generation tombstones while protecting every tombstone still needed by an active older refresh.
- Keep metadata initialization deferred in Python `PaimonCatalog`; DataFusion registration initializes the provider and `refresh_metadata()` remains an explicit strict refresh.

### Tests

- `cargo clippy --locked -p paimon -p paimon-datafusion --all-targets -- -D warnings`
- `cargo test -p paimon-datafusion --test sql_context_tests` (94 passed)
- `cargo test -p paimon-datafusion --test table_type_routing` (41 passed)
- `cargo test -p paimon-datafusion --lib 'sql_context::tests'` (176 passed)
- `cargo test -p paimon-datafusion --doc` (2 passed, 3 ignored)
- `cargo test -p paimon --test rest_catalog_test test_catalog_lists_declared_table_types` (passed)
- `PYO3_PYTHON=/opt/homebrew/bin/python3.11 cargo check --locked -p pypaimon_rust` (passed)
- Python suite after rebuilding the extension: 185 passed; 4 shared-warehouse fixture tests unavailable locally

### API and Format

`PaimonCatalogProvider::new` and `PaimonSchemaProvider::new` are removed because a synchronous constructor cannot preserve the previous ready discovery behavior without restoring the blocking bridge. Migrate initialized callers to `try_new(...).await?`; use `new_uninitialized(...)` only when initialization is deliberately deferred and complete `initialize_metadata().await?` before exposing discovery callbacks.

`Catalog::list_table_types` is a new defaulted method. Existing Paimon-only catalog implementations remain source compatible; catalogs containing object or external table types should override it.

No storage format changes.

### Documentation

Updated provider API documentation to describe the explicit initialized/deferred lifecycle and compile-check the removed synchronous constructor path.


crates/integrations/datafusion/src/catalog.rs
```diff
@@ -423,15 +646,22 @@ impl CatalogProvider for PaimonCatalogProvider {
         let catalog_name = self.catalog_name.clone();
         let session_state = self.session_state.clone();
         let schema_force_view_types = self.schema_force_view_types;
+        let table_engines = self.table_engines();
+        let metadata = Arc::clone(&self.metadata);
         let name = name.to_string();
         block_on_with_runtime(
```
JingsongLi, line 1353: Could we avoid keeping the blocking bridge in the remaining mutation callbacks before closing #872? `register_schema`, `deregister_schema`, and `deregister_table` still move the future to the process runtime and synchronously join the OS thread. If that future waits for work scheduled on a single-worker caller runtime, the caller worker remains blocked and the same deadlock class as #872 persists. Moving these catalog mutations to explicit async paths, or narrowing the issue scope, would make the claimed supersession precise.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -49,6 +53,118 @@ pub(crate) type SessionStateProvider = Arc<dyn Fn() -> Option<SessionState> + Se
...
+}
+
+impl CatalogMetadataState {
+    fn begin_refresh(&self) -> u64 {
+        self.next_generation.fetch_add(1, Ordering::AcqRel) + 1
+    }
+
+    fn publish(&self, generation: u64, next: Arc<CatalogMetadataSnapshot>) {
+        let mut current = self.snapshot.write().unwrap_or_else(|e| e.into_inner());
+        if generation >= self.published_generation.load(Ordering::Acquire) {
```
JingsongLi: Could we make generation tracking per database, or merge non-conflicting results? A database-scoped refresh with generation N+1 causes an already-running full refresh at N to discard its entire snapshot, including newly loaded metadata for unrelated databases, while still returning `Ok(())`. That makes a successful full refresh silently ineffective for unrelated entries.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -783,12 +1068,19 @@ impl SchemaProvider for PaimonSchemaProvider {
                             declared,
                         ))
                         .await?;
-                    Ok(resolved.map(|inner| {
-                        Arc::new(ReadOnlyTableProvider {
+                    Ok(Some(match resolved {
```
JingsongLi: `TableEngineResolver` documents `Ok(None)` as not found, but this converts it into `Some(UnavailableEngineTableProvider)`. When an engine reports a deleted, absent, or unowned object, DataFusion now treats the table as existing, planning can succeed, and execution fails later. Please preserve `Ok(None)` for a registered resolver; the metadata-only fallback can remain for the no-resolver case.

crates/integrations/datafusion/src/catalog.rs
```diff
@@ -924,101 +1208,25 @@ impl SchemaProvider for PaimonSchemaProvider {
...
-            if !system_tables::is_registered(system_name) {
-                return false;
-            }
+        if object
+            .system_table()
+            .is_some_and(|system_name| !system_tables::is_registered(system_name))
+        {
+            return false;
+        }
+        if object.system_table().is_some() {
```
JingsongLi: Returning false for every recognized system table makes `table_exist("t$snapshots")` disagree with `table("t$snapshots")` and with successful Paimon system-table queries. This regresses the previous behavior and violates the `SchemaProvider::table_exist` existence contract for direct callers. Please derive the answer from the snapshotted base table, or retain enough metadata to distinguish supported and unsupported derived tables.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -473,7 +489,18 @@ impl SQLContext {
             ));
         }
 
-        match &statements[0] {
+        let refresh_metadata_after = self.metadata_change_targets(&statements[0], false)?;
+        let metadata_mutation_targets = self.metadata_change_targets(&statements[0], true)?;
+        {
+            let _refresh_guard = self.metadata_refresh_gate.lock().await;
```
JingsongLi: Could the refresh targets be computed before taking a per-target singleflight lock? This context-wide gate is acquired by every SQL statement, even `SELECT 1`, and is held across remote catalog I/O. One stalled missing-table refresh therefore blocks unrelated existing-table queries and queries against other catalogs.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -751,7 +807,204 @@ impl SQLContext {
                 self.ctx.sql(&expanded.to_string()).await
             }
             _ => self.ctx.sql(sql).await,
+        };
+
+        if result.is_ok() && !refresh_metadata_after.is_empty() {
+            if let Err(error) = self.refresh_metadata_targets(refresh_metadata_after).await {
```
JingsongLi: This refresh runs after the catalog mutation has committed, but the successful DDL result is still withheld while an unbounded metadata listing is awaited. If the listing stalls, the client is left uncertain and may retry an already-applied DDL. Please update the known snapshot directly, or detach or bound this best-effort refresh.

crates/integrations/datafusion/src/sql_context.rs
```diff
@@ -751,7 +807,204 @@ impl SQLContext {
...
+            })?;
+            let provider = provider
+                .downcast_ref::<crate::catalog::PaimonCatalogProvider>()
+                .ok_or_else(|| {
+                    DataFusionError::Plan(format!(
+                        "Catalog '{catalog_name}' is not a Paimon catalog"
+                    ))
+                })?;
+            if schema.eq_ignore_ascii_case("information_schema") {
+                targets.insert(MetadataRefreshTarget::Catalog(catalog_name.to_string()));
```
JingsongLi: DataFusion information schema enumerates every catalog in the session, but this refreshes only the catalog that qualifies the `information_schema` relation. In a multi-catalog session, filtering for rows from another catalog can therefore return stale or missing metadata. Please refresh all participating Paimon catalogs, or safely derive a literal `table_catalog` restriction and refresh that target.

crates/integrations/datafusion/src/catalog.rs
shyjsarah: Agreed. I narrowed the PR scope instead of claiming the remaining bridges are fixed: the description now uses Refs #872, explicitly documents register_schema, deregister_schema, and deregister_table as follow-up work, and leaves #872 open.

crates/integrations/datafusion/src/sql_context.rs
shyjsarah: Fixed in 4f68a94. Refresh targets are computed before locking, locks are keyed by catalog/database and weakly retained, and independent targets refresh concurrently. Metadata-free queries and other databases no longer wait behind a stalled refresh; same-target misses still share the TTL/single-flight behavior.

crates/integrations/datafusion/src/catalog.rs
shyjsarah: Fixed in 4f68a94. A registered resolver returning Ok(None) is now preserved as None. The UnavailableEngineTableProvider fallback is only used when no resolver is registered. Added a regression test for the registered-resolver not-found case.

crates/integrations/datafusion/src/catalog.rs
shyjsarah: Fixed in 4f68a94. table_exist now recognizes supported system-table suffixes and derives existence from the snapshotted base table; missing bases and unknown suffixes remain false. Added direct table_exist coverage.

crates/integrations/datafusion/src/catalog.rs
shyjsarah: Fixed in 4f68a94. Generation tracking is now per database, and a full refresh merges each database independently. Newer scoped updates and tombstones win only for their own database, while non-conflicting results from the older full refresh are still published. Added concurrent merge and stale-resurrection regression tests.

crates/integrations/datafusion/src/sql_context.rs
shyjsarah: Fixed in 4f68a94. Successful DDL now applies the known create/drop/rename delta directly to the in-memory snapshot and does not await a post-commit listing. Added a blocked-listing regression test to verify the DDL result returns immediately.

crates/integrations/datafusion/src/sql_context.rs
shyjsarah: Fixed in 4f68a94. An information_schema reference now refreshes every registered Paimon catalog concurrently. Those catalog-wide refreshes are best-effort so one unavailable catalog preserves its last good snapshot without hiding strict failures for ordinary missing-object targets. Added multi-catalog, failure, and concurrency tests.

shyjsarah:
Follow-up review findings are addressed in 1b705e0:

- automatic `information_schema` refresh now has freshness caching, failure backoff, per-target single-flight, a global concurrency bound, and a timeout;
- rejected same-database full-refresh publications are retried instead of silently dropping unrelated discoveries;
- `ALTER TABLE IF EXISTS` no-op renames no longer create phantom tables;
- same-database remote DDL and local deltas share one gate, including database/table DDL interactions;
- the original synchronous provider constructor signatures are restored, with `try_new` as the initialized async path;
- routed/not-found existence semantics and database-generation tombstone retention are tightened.

Local verification: 92 SQL-context tests, 40 routing tests, Clippy/format checks, and 185 Python tests passed. The PR description now reflects the final API and test matrix.

shyjsarah:
Closing this PR because #891 has landed and addresses the original Python/process-runtime information_schema deadlock.

The snapshot approach here adds a much broader metadata caching, freshness, DDL-coherence, and lifecycle subsystem that is no longer justified for the original incident. #891 does not claim to cover every caller-owned Tokio runtime topology; if that remaining boundary can be reproduced, it should be tracked separately with a focused regression test and a narrower fix.

PR #882, opened 2026-09-20
fix(variant): convert shredded timestamps to the column's time unit

jackylee-ch:
`ShreddedValue::Timestamp` is always micros, but the shredded leaf's Arrow unit follows the declared precision (MILLIS <4, MICROS 4..=6, NANOS >6). Both sides passed the integer through unchanged, so a TIMESTAMP leaf outside 4..=6 was 1000x off. A NANOS leaf is spec-conforming, so foreign files were mis-read too.

Write now scales micros to the leaf unit, flooring into MILLIS, and errors on micros beyond i64 nanoseconds (~1677..2262) instead of wrapping. Read keys off the array's TimeUnit, so a foreign file decodes from its own schema.

Disclosure: files paimon-rust already wrote outside 4..=6 hold micros under a MILLIS/NANOS annotation. They read back correctly only because the reader repeated the mistake; after this change they decode 1000x wrong, and nothing in the file tells them apart. Only an explicit `variant.shreddingSchema` produces such a leaf — inference picks precision 6.


JingsongLi:
Reviewed head b75b8b91. Requirement fit: SUPPORTED. The seven focused shredded-timestamp tests pass, but two follow-ups remain. **Integration/rebase follow-up:** main commit d05cb03a (#884) added `TimeUnit::Second` for precision-zero timestamps after this branch diverged. This PR still has the earlier `0..=3` millisecond arm and no Seconds reader arm. GitHub marks the PR CONFLICTING, so this is not a demonstrated regression in a merged result; please reconcile the two changes on rebase, preserve Seconds↔micros conversion for `TIMESTAMP(0)` and `TIMESTAMP_LTZ(0)`, and add precision-zero read/write tests. **Persisted-file compatibility:** as the PR body discloses, older Rust writes with explicit precision 1–3 or 7–9 shredding schemas stored micros under millis/nanos annotations. The new reader would silently reinterpret those typed values, and the files carry no marker to distinguish them from conforming files. Please make the migration/compatibility policy explicit (for example, a deliberate legacy mode or required rewrite) and include an old-writer fixture so the impact is testable.

JingsongLi (commented):
Reviewed head `a9eff7c6`. Fixing the Arrow unit mismatch has real end-to-end value, especially for spec-conforming files written by another engine. The precision-zero reconciliation and old-writer fixture address parts of my earlier comment. I found two production blockers:

**[P1] New writes silently change a Variant timestamp that the declared leaf cannot represent.** `try_typed_shred` (`crates/paimon/src/variant.rs:2998-3003`) puts any microsecond timestamp into `typed_value`. The new `timestamp_array` conversion (`crates/paimon/src/arrow/shredding/variant.rs:1058-1068`) floors it to seconds/millis, and reassembly reconstructs from that rounded typed value because the original `value` is absent. I ran a temporary complete Variant shred→assemble roundtrip: a `TIMESTAMP_NTZ` value of `1_700_000_000_123_456` microseconds with a `TIMESTAMP(3)` shredding schema came back as `1_700_000_000_123_000`. Please keep the original Variant in the `value` fallback when it is not exactly representable in the typed leaf, and add a persisted-file roundtrip test. The same representability check should send a valid microsecond timestamp outside the nanosecond leaf's range to `value` instead of failing the entire write batch.

**[P1] The on-disk unit change needs an executable migration and rollout path.** For explicit shredding schemas at precision 0–3 or 7–9, a new reader reinterprets old Rust files, while an old reader misinterprets newly written conforming files. The old-writer unit test now documents one wrong-result case, but it does not prevent mixed-version reads/writes or identify which existing files need rewriting; precision-zero historical files need coverage too. Before a production merge, please document how to inventory affected tables/files, stop or coordinate old writers/readers, rewrite affected data using a version that still interprets it correctly, verify values, and roll back safely. Otherwise this can silently change persisted timestamps during a rolling upgrade.

Validation: all 20 `arrow::shredding::variant::tests` passed and head CI is green. My temporary full roundtrip regression failed with the exact values above, then I removed it and restored a clean checkout. The [Parquet Variant shredding specification](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) permits retaining an unrepresentable typed value in the Variant `value` field.

jackylee-ch:
Rebased onto main. Precision-0 SECOND arm + tests, plus a rewrite policy and migration fixture.

PR #883, opened 2026-09-20
[rust] Bound sparse BlobDescriptor range reads

JingsongLi:
### Purpose

Prevent sparse BlobDescriptor selections from causing disproportionate I/O and retained buffers after the inline BLOB coalescing work.

### Changes

- cap merged descriptor ranges at 2x selected-byte amplification in addition to the existing gap and span limits
- copy selected values out of merged reads that contain gaps, so small byte slices do not retain the full backing response
- preserve zero-copy slices for fully selected contiguous ranges
- add regression coverage for sparse selections and backing-buffer release

For 128 selected 4 KiB values separated by 64 KiB gaps, the previous policy merged them into two reads totaling 8.375 MiB for 512 KiB selected data. The new amplification bound keeps this layout as 128 exact reads totaling 512 KiB.

### Tests

- cargo test --locked -p paimon table::blob_resolver::tests::
- cargo clippy --locked --all-targets --workspace --features fulltext,vortex -- -D warnings
- cargo test --locked -p paimon --all-targets --features fulltext,vortex


leaves12138 (approved):
LGTM. The merge budget counts unique selected bytes rather than overlapping/duplicate requests, and gapped reads copy out the selected values before their backing buffers can be retained by collection. Contiguous fully selected ranges retain the zero-copy path, and row placement and the existing shared request/byte limits are preserved.

I ran the BLOB-related paimon library tests at this commit with `--no-default-features --features storage-fs,storage-memory`: 138 passed, including the sparse-range, backing-buffer, ordering, and concurrency coverage.

PR #884, opened 2026-09-20
[rust] Map precision-zero timestamps to Arrow seconds

JingsongLi:
### Purpose

Align TIMESTAMP(0) and TIMESTAMP_LTZ(0) with their Arrow representation. Arrow precision zero uses seconds, while paimon-rust previously emitted milliseconds and forced PyPaimon native reads to fall back.

### Changes

- map precision 0 to Arrow Timestamp(Second), preserving milliseconds for precision 1 through 3
- convert physical millisecond values at row, Avro, partition, and residual-filter boundaries
- preserve Paimon's millisecond Parquet encoding by converting second-based Arrow schemas and batches at the writer boundary
- evaluate timestamp predicates losslessly across second, millisecond, microsecond, and nanosecond Arrow arrays, including fractional literals and existing millisecond Parquet files
- support second arrays in BinaryRow, Variant shredding, and min/max aggregation paths
- add schema, row-format, partition, aggregation, Parquet footer/filter, and DataFileReader round-trip coverage, including pre-epoch values

### Tests

- `cargo test --locked -p paimon timestamp_zero -- --nocapture`


crates/paimon/src/arrow/mod.rs
```diff
@@ -149,7 +149,8 @@ pub fn paimon_type_to_arrow(dt: &PaimonDataType) -> crate::Result<ArrowDataType>
 
 fn timestamp_time_unit(precision: u32) -> crate::Result<TimeUnit> {
     match precision {
-        0..=3 => Ok(TimeUnit::Millisecond),
+        0 => Ok(TimeUnit::Second),
```
leaves12138, line 152: [P1] Keep the Parquet physical timestamp representation in milliseconds

This mapping also reaches the write path (`TableWrite::new` -> `build_target_arrow_schema` -> `ParquetFormatWriter`), not just the public read schema. Parquet has no seconds timestamp logical type. With the current Arrow writer, a `Timestamp(Second)` column is written as an unannotated INT64 containing seconds, whereas Paimon's precision <= 3 Parquet statistics/pruning code still interprets the physical INT64 values as milliseconds.

I reproduced this using `ParquetFormatWriter` with a `TIMESTAMP(0)` field and values `[-1s, 2s, 3s]`, followed by `ParquetFormatReader` with `PredicateBuilder::equal("ts", Datum::Timestamp { millis: 2000, nanos: 0 })`. An unfiltered read returns all three rows, but the filtered read silently returns **zero rows instead of one**. The new file has no Parquet timestamp annotation and physical min/max `[-1, 3]`, so statistics pruning treats 2000 ms as outside the file. On the pre-PR base the same test writes `TIMESTAMP_MILLIS` with min/max `[-1000, 3000]` and returns the matching row.

Could we keep the Arrow seconds representation at the API boundary but normalize the Parquet writer's schema and arrays to millisecond timestamps? Please add a TIMESTAMP(0)/TIMESTAMP_LTZ(0) Parquet write/read/filter round-trip test, including the physical footer representation, so this does not change Paimon's on-disk timestamp encoding.

crates/paimon/src/arrow/residual.rs
```diff
@@ -1114,7 +1114,15 @@ fn timestamp_scalar(
     timezone: Option<&'static str>,
 ) -> crate::Result<Option<ArrayRef>> {
     let array: ArrayRef = match precision {
-        0..=3 => {
+        0 => {
+            let value = millis.div_euclid(1_000);
+            let array = TimestampSecondArray::new_scalar(value).into_inner();
```
leaves12138, line 1284: [P1] Match the decoded Parquet column's timestamp unit before comparison

Existing TIMESTAMP(0) Parquet files are decoded as `Timestamp(Millisecond)`. The single-leaf fast path in `build_parquet_arrow_predicate` passes that physical array directly to `evaluate_exact_leaf_predicate`, but this change now constructs its comparison literal as `Timestamp(Second)`. Unlike the general residual path, this decoder fast path does not normalize the array first.

I reproduced an equality filter on a millisecond-encoded Parquet file containing `[-1000, 2000, 3000]`, with logical field type `TIMESTAMP(0)` and literal `Datum::Timestamp { millis: 2000, nanos: 0 }`. The read now fails with:

```text
Invalid comparison operation: Timestamp(ms) == Timestamp(s)
```

The identical test passes on the pre-PR base. Please reconcile the physical column and literal units in the decoder predicate path (without losing literal precision), and cover filtering existing millisecond-encoded TIMESTAMP(0)/TIMESTAMP_LTZ(0) Parquet files. Changing the returned Arrow schema should not make existing files unreadable under ordinary predicates.

crates/paimon/src/arrow/residual.rs
```diff
@@ -1114,7 +1114,15 @@ fn timestamp_scalar(
     timezone: Option<&'static str>,
 ) -> crate::Result<Option<ArrayRef>> {
     let array: ArrayRef = match precision {
-        0..=3 => {
+        0 => {
+            let value = millis.div_euclid(1_000);
```
leaves12138, line 1283: [P2] Preserve subsecond precision in predicate literals

The column's precision does not mean that a comparison literal can be rounded down to the same precision. `PredicateBuilder` accepts a timestamp literal with `millis: 1500` for a TIMESTAMP(0) field. This division changes that literal from 1.5 seconds to 1 second before an exact comparison.

For a valid TIMESTAMP(0) column containing `[1s, 2s]`, I reproduced both incorrect masks through `evaluate_predicates_mask`:

- `ts = 1.5s`: actual `[true, false]`, expected `[false, false]`.
- `ts >= 1.5s`: actual `[true, true]`, expected `[false, true]`.

Both comparisons are correct on the pre-PR millisecond Arrow representation. This is independent of the Parquet physical-unit issue: it also affects residual evaluation on correctly constructed second-based Arrow arrays. Please compare the column and literal in a common lossless unit, or use exact timestamp-value comparison, rather than truncating the literal. A regression test should exercise a fractional-second literal against whole-second stored values.

leaves12138 (commented):
The Arrow schema alignment makes sense, but I found three correctness regressions that need fixing before approval: the new Parquet write representation can cause false-negative statistics pruning; predicates over existing millisecond-encoded Parquet files fail with an Arrow type mismatch; and fractional-second predicate literals are rounded down, producing incorrect results.

Validation at `6b17b8e`:

- The existing paimon library suite passed: 2912 tests passed, 2 ignored, using `--no-default-features --features storage-fs,storage-memory`.
- I added three focused regression tests locally. All three pass on the pre-PR merge base `29a6510`, and all three fail on this PR head. The Parquet tests use the actual Paimon format writer/reader path; the literal test uses `PredicateBuilder` and the residual evaluator.

The inline comments include the concrete inputs, observed failures, and suggested fixes.

crates/paimon/src/arrow/residual.rs
JingsongLi: Fixed in 531c5a06. Exact timestamp predicates now compare the decoded Arrow value and Datum literal in lossless i128 epoch nanoseconds, independently of whether the physical array is seconds, milliseconds, microseconds, or nanoseconds. Added coverage for existing millisecond-encoded TIMESTAMP(0) and TIMESTAMP_LTZ(0) Parquet files and for the complete DataFileReader path.

crates/paimon/src/arrow/mod.rs
JingsongLi: Fixed in 531c5a06. The Parquet writer now keeps the public input schema as Timestamp(Second) but recursively normalizes second timestamps to Timestamp(Millisecond) for the physical writer and casts each batch at that boundary. The regression test asserts INT64 + TIMESTAMP_MILLIS (including LTZ adjusted-to-UTC), millisecond min/max statistics, a successful filtered read, and the final DataFileReader output returning Timestamp(Second).

crates/paimon/src/arrow/residual.rs
JingsongLi: Fixed in 531c5a06. Timestamp literals are no longer divided to the column precision. The exact evaluator preserves millis plus nanos-of-millisecond and compares in i128 epoch nanoseconds. The focused regression covers equality and greater-than-or-equal for a 1.5-second literal against [1s, 2s] for both TIMESTAMP(0) and TIMESTAMP_LTZ(0).

leaves12138 (approved):
LGTM at `8a48fbc`.

The earlier Parquet encoding, physical-unit comparison, and fractional-literal regressions are fixed. The follow-up Mosaic fix is now on this PR as well: it keeps the format reader's timestamp schema in milliseconds (including nested containers) and leaves the existing DataFileReader output conversion to expose Arrow seconds. This preserves the on-disk representation and avoids the Mosaic seconds-validation failure.

I verified that the remote commit tree exactly matches the locally tested fix and reran the default-feature paimon library suite: **2925 passed, 2 ignored**. The new Mosaic end-to-end regression covers TIMESTAMP(0), TIMESTAMP_LTZ(0), pre-epoch values, NULLs, ARRAY/MAP timestamp values, fractional predicates, array membership, disjoint row selections, and enabled/disabled prefetch. `cargo clippy --locked -p paimon --lib --tests -- -D warnings` and `cargo fmt --all -- --check` also pass.

No remaining blocking issue found in this review. CI still needs to complete before merging.

PR #886, opened 2026-09-20
fix(scan): keep the predicate when a partition set cannot express it

jackylee-ch:
**Defect.** `from_predicate` recognised only `And` plus `Eq`/`In`/`IsNull` leaves; everything else fell through `_ => {}`, discarded. With every field covered it built a `PartitionSet`, and `matches_entry` is then pure `HashSet` membership — the predicate is never read again. A partition-only filter is `Exact`, so DataFusion drops its residual: `WHERE dt='2024-01-01' AND dt>='2024-01-02'` returned 1 row, not 0.

**Fix.** `collect_eq_candidates` reports whether it consumed the whole tree; `from_predicate` keeps the predicate when it did not. A second conjunct on an already-pinned field also bails: the old code overwrote it, keeping whichever came last — possibly the wider one.

**Cost.** The fallback pays `from_serialized_bytes` + `eval_row` per entry instead of a hash lookup, and an `In` leaf loses its min/max bound (`stats_filter.rs:73`). Only on shapes answered wrongly before.


JingsongLi:
Reviewed head f63548c9. Requirement fit: SUPPORTED; implementation: CLEAN. The pre-fix PartitionSet could drop a conjunct while DataFusion treated a partition-only filter as exact; returning the original predicate for any unrepresented node or repeated field restores the entry-level check, and manifest pruning remains conservative. I ran `cargo test -p paimon table::partition_filter::tests --lib` (11 passed) and `cargo test -p paimon-datafusion --test partition_conjunct_pushdown` (3 passed). No actionable issue found in this head. FYI #858 also changes `partition_filter.rs`; if it lands first, please preserve these contradictory-conjunct SQL regressions when reconciling the implementations.

JingsongLi (approved):

PR #889, opened 2026-09-20
[python] Deserialize cross-language splits for native reads

JingsongLi:
### Purpose

Let PyPaimon pass splits produced by its Python planner to the Rust reader through the stable Java-compatible `SplitSerializer` v1 wire format. This keeps native physical reads available when planning cannot run in Rust, including Data Evolution row-range reads.

Paired PyPaimon change: apache/paimon#10017.

### Brief change log

- add `pypaimon_rust.datafusion.Split.deserialize(bytes)` for `SplitSerializer` v1 frames
- keep the stable wire API separate from the opaque JSON state used by pickle
- support DataSplit v8/v9 and score-free IndexedSplit row ranges through the existing Rust decoder
- expose the new entry point in the Python type stub
- verify ordinary and Data Evolution reads after a serialize/deserialize round trip
- reject malformed wire frames with Python `ValueError`

### Tests

- `cargo test -p pypaimon_rust` (76 passed)
- `cargo clippy -p paimon -p pypaimon_rust --all-targets -- -D warnings`
- built and installed the wheel, then ran `bindings/python/tests/test_read.py` (86 passed)
- paired PyPaimon native integration suite (35 passed)

### API and Format

Adds a Python binding API. It does not change a storage format; the input is the existing `SplitSerializer` v1 format.

### Documentation

The type stub documents the distinction between stable split bytes and pickle state. No user guide changes are required.


leaves12138 (approved):
LGTM. The new entry point delegates to the existing checked SplitSerializer v1 decoder and keeps the stable wire API separate from opaque pickle state. Verified Java DataSplit v8/v9 frames, streaming flags, score-free IndexedSplit row ranges, pickle compatibility, and malformed/unsupported-input error mapping. Validation passed: 7 wire codec tests, 76 Rust binding tests, 86 Python read tests, 9 additional fixture/error-boundary checks, and 55 paired PyPaimon integration/serialization tests, including actual Python-planned splits read through Rust. No blocking findings.

